PR #6, opened 2016-01-22
Add acceptance tests for basic Mender functionality.

kacf:
- Update with a bogus image.
- Update a proper image.
- Fail to update an image because of insufficient space.
- Commit image

### DIFF
--- a/tests/acceptance/README.md
+++ b/tests/acceptance/README.md
@@ -1,0 +1,38 @@
+Acceptance tests for Mender
+===========================
+
+
+Prerequisites
+-------------
+
+* PyTest
+* [Fabric](http://www.fabfile.org)
+
+
+Running
+-------
+
+To run these tests you need to be in the tests/acceptance directory, and the
+meta-mender-qemu layer must be next to the meta-mender layer (this layer). In
+addition you must provide a symlink, "image.dat", pointing to image you want to
+test. Normally this is the "*.ext3" image inside of your build directory, in
+"tmp/deploy/images/vexpress-qemu". Then run the tests with:
+
+```
+$ py.test -s
+```
+
+
+Known issues
+------------
+
+The tests currently leave behind a modified image. They won't currently pass if
+you run them again a second time without rebuilding the image first.
+
+
+Troubleshooting
+---------------
+
+A common problem is that the SSH server inside the image stops working. This
+often happens if you run the tests several times after one another. It is
+unknown why, but rebuilding the Yocto image seems to make the problem go away.
--- a/tests/acceptance/common.py
+++ b/tests/acceptance/common.py
@@ -1,0 +1,151 @@
+#!/usr/bin/python
+# Copyright 2016 Mender Software AS
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+from fabric.api import *
+import fabric.network
+
+import pytest
+import os
+import signal
+import subprocess
+import time
+
+import conftest
+
+# Return Popen object
+def start_qemu():
+    # Relies on the meta-mender layer being next to meta-mender-qemu.
+    proc = subprocess.Popen("../../../meta-mender-qemu/scripts/mender-qemu")
+    execute(qemu_prep_after_boot, hosts = conftest.current_hosts())
+    return proc
+
+
+def kill_qemu():
+    os.system("pkill qemu-system-arm")
+
+
+def reboot(wait = 60):
+    with settings(warn_only = True):
+        sudo("reboot")
+
+    # Make sure reboot has had time to take effect.
+    time.sleep(5)
+
+    fabric.network.disconnect_all()
+
+    run_after_connect("true", wait = wait)
+
+    qemu_prep_after_boot()
+
+
+def run_after_connect(cmd, wait = 60):
+    output = ""
+    start_time = time.time()
+    # Use shorter timeout to get a faster cycle.
+    with settings(timeout = 5, abort_exception = Exception):
+        while True:
+            attempt_time = time.time()
+            try:
+                output = run(cmd)
+                break
+            except Exception as e:
+                print("Could not connect to host %s: %s" % (env.host_string, e))
+                if attempt_time >= start_time + wait:
+                    raise Exception("Could not reconnect to QEMU")
+                now = time.time()
+                if now - attempt_time < 5:
+                    time.sleep(5 - (now - attempt_time))
+                continue
+    return output
+
+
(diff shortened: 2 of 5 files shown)


kacf:
@pasinskim: Maybe you want to take a look, since you're familiar with PyTest and Fabric.

tests/acceptance/README.md
```diff
@@ -0,0 +1,38 @@
+Acceptance tests for Mender
+===========================
+
+
+Prerequisites
+-------------
+
+* PyTest
+* Fabric (fabfile.org)
+
```
pasinskim: [Fabric](fabfile.org)

tests/acceptance/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+Running
+-------
+
+To run these tests you need to be in this directory, and the meta-mender-qemu
```
pasinskim: you need to be in ```tests/acceptance``` directory

tests/acceptance/README.md
```diff
@@ -0,0 +1,38 @@
...
+layer must be next to the meta-mender layer (this layer). In addition, there
+must be a symlink "image.dat" pointing to the image you want to test. Normally
```
pasinskim: Maybe: In addition you must provide ```image.dat``` symlink pointing to image you want to test.

tests/acceptance/README.md
```diff
@@ -0,0 +1,38 @@
...
+this is the "*.ext3" image inside of your build, in
```
pasinskim: of your build directory

tests/acceptance/common.py
```diff
@@ -0,0 +1,151 @@
...
+import os
+import signal
+import subprocess
+import time
+
+import conftest
+
+# Return Popen object
+def start_qemu():
+    # Relies on being run from a meta-layer next to meta-mender-qemu.
```
pasinskim: Shouldn't be: form meta-mender?

tests/acceptance/common.py
```diff
@@ -0,0 +1,151 @@
...
+    proc = subprocess.Popen("../../../meta-mender-qemu/scripts/mender-qemu")
+    execute(qemu_prep_after_boot, hosts = conftest.current_hosts())
+    return proc
+
+
+def kill_qemu():
+    os.system("pkill qemu-system-arm")
```
pasinskim: Maybe I'm looking too much into future now, but what if we will start supporting something else than arm? Maybe we can not hardcode those commands?

tests/acceptance/common.py
```diff
@@ -0,0 +1,151 @@
...
+    elif len(host_parts) == 1:
+        host = host_parts[0]
+        port = ""
+    else:
+        raise Exception("Malformed host string")
+
+    return (cmd, host, port)
+
+
+# Yocto build SSH is lacking SFTP, let's override and use regular SCP instead.
```
pasinskim: Nice workaround!

tests/acceptance/common.py
```diff
@@ -0,0 +1,151 @@
...
+    (scp, host, port) = scp_prep_args()
+
+    local("%s %s %s@%s:%s/%s %s" %
+          (scp, port, env.user, host, remote_path, file, local_path))
+
+
+def qemu_prep_after_boot():
+    # TODO: It is unknown why this is needed. "/u-boot" with "auto"
+    # attribute is in fstab already... This should be removed.
+    run_after_connect("mount /u-boot || true")
```
pasinskim: This shouldn't bee needed any more. I have tested latest build and it works just great without manual mounting.

tests/acceptance/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+
+import common
+
+def pytest_addoption(parser):
+    parser.addoption("--host", action = "store",
+                     help = ("IP to connect to, with optional port. Defaults " +
+                             "to localhost:8822, which is what the QEMU " +
+                             "script sets up."))
+    parser.addoption("--user", action = "store",
+                     help = "user to log into remote hosts with (default is root)")
```
pasinskim: Will we ever need password? If we allow changing user maybe we should add option to modify password as well?

tests/acceptance/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+        host = config.getoption("host")
+    env.hosts = [host]
+
+    env.user = "root"
+    if config.getoption("user"):
+        env.user = config.getoption("user")
+
+    env.password = ""
+
+    # Disable known_hosts file, to avoid "host identification changed" errors.
```
pasinskim: Nice to know this one!

tests/acceptance/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+    env.disable_known_hosts = True
+
+    env.abort_on_prompts = True
+
+
+def pytest_unconfigure(config):
+    common.kill_qemu()
```
pasinskim: Is it enough? Do we need to somehow clean u-boot environment so that we will be able to redo all the tests?

tests/acceptance/test_update.py
```diff
@@ -0,0 +1,121 @@
...
+
+        kill_qemu()
+        start_qemu()
+
+        output = run_after_connect("mount")
+
+        # The update should have reverted to /dev/mmcblk0p2, since the image was
+        # bogus.
+        assert(output.find("/dev/mmcblk0p2") >= 0)
+        assert(output.find("/dev/mmcblk0p3") < 0)
```
pasinskim: Maybe we can make this configurable? If we will play with image creating Yocto recipes this might be different in future.

pasinskim:
Great stuff! Looking forward to have this integrated with Jenkins :+1:

tests/acceptance/common.py
kacf: I know, I don't like this one either. I tried to use `subprocess.terminate()`, but I couldn't get this to work, despite trying a lot. It seemed surprisingly difficult to do it the right way, so in the end I settled for this. I agree we should change this in the more longterm, but I think it's not worth spending more time on it right now.

tests/acceptance/common.py
kacf: I tried this quite recently, and it didn't work for me, but I will retry.

tests/acceptance/conftest.py
kacf: The `core-image-full-cmdline` always uses passwordless login AFAIK. And in either case it should be key based, not password based.

tests/acceptance/conftest.py
kacf: No, it's not enough. But the steps required to clean up *properly* are a bit more involved: Basically you need to convert the image to a qcow2 image first, which you can use as a copy-on-write snapshot for QEMU. This would allow you to revert to the old image on every test. But I consider this a separate task, since it's not a quick fix.

tests/acceptance/test_update.py
kacf: I'm torn on what is the best way here... I could make it a constant in the common file and refer to it from there? But still it won't get it from Yocto. I'm open to suggestions here..

tests/acceptance/common.py
kacf: Nope, this doesn't work in latest master.

kacf:
All other issues not explicitly commented fixed!

tests/acceptance/test_update.py
kacf: Will leave it for now, and we can decide how to solve this when we add more boards with different memory card properties.

PR #10, opened 2016-02-23
Fix documentation and move Qemu launching script from meta-mender-qemu.

pasinskim:


pasinskim:
This is documentation fixing and moving scripts form other repository so looks safe to merge.

PR #14, opened 2016-02-24
Make sure that correct version of u-boot will be used.

pasinskim:
Each Yocto version is updating u-boot. We need to make sure that we can
apply custom patches to support both beaglebone and vexpress-quemu.
For this we are using u-boot v2015.10 and this is defined as
PREFFERED_VERSION in mender-install class.


pasinskim:
@kacf can you take a brief look at this one to make sure I am not doing anything stupid here?

kacf:
Most of this is copied directly from poky, I assume?

pasinskim:
@kacf u-boot is a poky copy, indeed.

kacf:
Ok, it looks good to me.

PR #15, opened 2016-02-26
Have Mender to be systemd daemon

pasinskim:


pasinskim:
@kacf knowing your prior experience with systemd I would like you to take a look at this.

recipes-mender/mender/files/mender.timer
```diff
@@ -0,0 +1,11 @@
...
+Description=Timer for running Mender periodically 
+After=syslog.target
+
+[Timer]
+OnBootSec=1min
+OnUnitActiveSec=5min
+Unit=mender.service
+
+[Install]
+WantedBy=multi.user.target
```
kacf: As discussed I think we definitely don't want this for Mender. Mender should either be a daemon or launched from some sort of on demand service daemon, such as xinetd (I think systemd also has some hooks for this). But most likely a standalone daemon.

recipes-mender/mender/files/mender.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Mender OTA update service
+
+[Service]
+Type=simple
+User=root
+Group=root
+ExecStart=/usr/bin/mender
```
kacf: I envision we might call mender with something like `--daemon` here. But at the moment we don't have it.

recipes-mender/mender/mender_0.1.bb
```diff
@@ -7,12 +7,25 @@ S = "${WORKDIR}/git"
...
+
 SRCREV = "${AUTOREV}"
 
 LICENSE = "Apache-2.0"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/Apache-2.0;md5=89aea4e17d99a7cacdbeed46a0096b10"
 
+inherit systemd
+
+#FILESEXTRAPATHS_prepend := " ${THISDIR}/mender:"
+#SRC_URI += "file://mender.service"
```
kacf: Remove?

kacf:
Looks good otherwise!

recipes-mender/mender/files/mender.timer
pasinskim: I agree. Fixed!

recipes-mender/mender/files/mender.service
pasinskim: Yup. Since we don't have it right now systemd will just fail starting Mender but this is enough for now to continue with Mender Go implementation.

recipes-mender/mender/mender_0.1.bb
pasinskim: Fixed!

pasinskim:
@kacf can you take a second look. I've removed timer that was supposed to run Mender every 5 minutes. I've also added network support so that we will be able to proceed with user-space go app development.

recipes-mender/mender/mender_0.1.bb
kacf: This is not fixed, it's still there!

recipes-core/systemd/systemd_%.bbappend
```diff
@@ -0,0 +1,17 @@
...
+
+FILES_${PN} += "${sysconfdir}/systemd/network/eth.network"
+
+
+do_install_append() {
+  if ${@bb.utils.contains('PACKAGECONFIG','networkd','true','false',d)}; then
+        install -d ${D}${sysconfdir}/systemd/network
+        install -m 0755 ${WORKDIR}/eth.network ${D}${sysconfdir}/systemd/network
+  fi
+}
```
kacf: I find it incredibly odd that there isn't already a standard way to achieve this. But no use in wondering, this is a good fix for now!

kacf:
If you fix the not-yet-quite-fixed-comment, you can go ahead and merge this yourself. It looks good otherwise!

pasinskim:
Done now. Merging!

PR #16, opened 2016-03-02
Fixes

pasinskim:
Add various fixes for building image for vexpress-qemu and beagleboard.


pasinskim:
@kacf those are various fixes that were needed to run Mender on vexpress and beaglebone. I've tested that this stuff is working on both so should be OK to merge this.

recipes-kernel/linux/linux-yocto/vexpress-qemu-standard.scc
```diff
@@ -0,0 +1,5 @@
+define KMACHINE vexpress-qemu 
+
+define KARCH arm
+define KTYPE standard
+branch master
```
kacf: Aha! So that's what was missing! I could never figure that one out.

recipes-kernel/linux/linux-yocto_4.1.bbappend
```diff
@@ -1,9 +1,13 @@
 # Use custom defconfig in order to enable use of the vexpress model.
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
-SRC_URI_append_vexpress-qemu += "file://defconfig"
+SRC_URI_append_vexpress-qemu = " file://defconfig \
+                                file://vexpress-qemu-standard.scc \
+                               " 
+
+COMPATIBLE_MACHINE_vexpress-qemu = "vexpress-qemu"
```
kacf: Hehe, this looks a bit like a circular statement. But if it works... :-)

PR #19, opened 2016-03-03
Make sdimg depend on native dosfstools.

kacf:
We need it because we use mkfs.vfat.


kacf:
Untested, but Greg needs it immediately.

PR #26, opened 2016-03-31
Make Ctrl-c not terminate emulator.

kacf:
It's quite annoying since it's common to press Ctrl-C inside the
emulator. The -nographic argument does some extra magic to avoid that.

Use "Ctrl-a x" instead to terminate emulator.


kacf:
@GregorioDiStefano: For your pleasure!

GregorioDiStefano:
Woohoo!

PR #28, opened 2016-03-31
Fix systemd hanging when detection device files.

kacf:
systemd apparently depends on the CONFIG_FHANDLE feature in the kernel
to do this.


kacf:
Thanks to @jimis for digging this out somewhere deep in the web!

PR #29, opened 2016-04-01
Remove host preparation that is not needed anymore.

kacf:
Systemd should have taken care of the mounting problem, and the mender
binary is always present in the image now.


kacf:
@pasinskim: Unless you know anywhere where auto-mounting `/u-boot` and `/data` does not work, I think we can merge this.

pasinskim:
This works great now. I think we can merge this one.

PR #30, opened 2016-04-01
Follow change in logic regarding bootcount when committing update.

kacf:
We don't update the bootcount anymore when Mender commits an update,
because it's not necessary (it's not used as long as
"upgrade_available" is 0), so don't check for that. But add some
additional checks right after we have updated, but before rebooting,
to make sure the expected environment is there.


kacf:
Should be merged after https://github.com/mendersoftware/mender/pull/39 is merged.

PR #31, opened 2016-04-01
Install ca-certificates so that Go TLS finds them when fetching an image update via HTTPS.

jimis:


jimis:
I tested this manually, it's really all that's needed for HTTPS to work.

kacf:
Looks good!

PR #32, opened 2016-04-01
Auto-reboot if boot command fails

jimis:
This way we fall back to the original partition after a failed image update.


jimis:
@eysteins With this patch, doing a failed image update (for example I flashed a deb package as rootfs) tries to boot, fails, and then automatically reboots to the original partition.

jimis:
@kacf @pasinskim I'm merging for the demo's sake, you can still review/comment later.

kacf:
Nice, and +1 for fixing the test as well!

PR #33, opened 2016-04-04
Add patch files from original Yocto u-boot meta recipe.

pasinskim:
Lack of one of those files is preventing form booting image
using Beaglebone Black. Next step should be to figure out which patches
are not important and remove those.


eysteins:
Verified to work on BBB.

eysteins:
Nice job!

PR #37, opened 2016-05-09
Prevent multiple ${IMAGE_BOOT_FILES} entries from breaking the build.

kacf:
Since we prepend the path we need to do it one by one.

### DIFF
--- a/classes/sdimg.bbclass
+++ b/classes/sdimg.bbclass
@@ -82,7 +82,10 @@ IMAGE_CMD_sdimg() {
     fi
 
     # Copy boot files to boot partition
-    mcopy -i "${WORKDIR}/fat.dat" -s ${DEPLOY_DIR_IMAGE}/${IMAGE_BOOT_FILES} ::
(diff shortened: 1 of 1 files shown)


kacf:
@pasinskim

classes/sdimg.bbclass
```diff
@@ -82,7 +82,10 @@ IMAGE_CMD_sdimg() {
     fi
 
     # Copy boot files to boot partition
-    mcopy -i "${WORKDIR}/fat.dat" -s ${DEPLOY_DIR_IMAGE}/${IMAGE_BOOT_FILES} ::
```
bboozzoo: Since I'm the author of IMAGE_BOOT_FILES I feel entitled to comment here :) IMO, instead of doing all of this manually  you should use sdimage-bootpart and just add 2 partitions with source set to rootfs. IMAGE_BOOT_FILES will be automagically handled behind the scenes.

kacf: You mean use `--source rootfs` in the wks command below? Yes, that makes sense.

PR #38, opened 2016-05-09
MEN-307: BBB: Don't fall back on internal flash if boot fails.

kacf:
Instead reset and retry from the memory card. Otherwise we do not fall
back to the previous partition, which is both important for testing,
and it is what Mender is supposed to do.

This removes the safety net of falling back to the internal flash
altogether, but I think this is a separate concern that we can tackle
later, perhaps when we have a bit more clarity on what Mender should
do in such a situation.

### DIFF
--- a/recipes-bsp/u-boot/files/0001-beaglebone-mender-specific-configuration.patch
+++ b/recipes-bsp/u-boot/files/0001-beaglebone-mender-specific-configuration.patch
@@ -63,7 +63,7 @@ index 3cf768e..540a82a 100644
  	"mmcloados=run args_mmc; " \
  		"if test ${boot_fdt} = yes || test ${boot_fdt} = try; then " \
  			"if run loadfdt; then " \
-@@ -120,56 +137,43 @@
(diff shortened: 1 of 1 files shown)


GregorioDiStefano:
Was this change tested with the qemu/fabric acceptance tests? 

BBB acceptance tests are working fine, but the qemu based ones are failing.

kacf:
This change has no effect on QEMU, this patch is not used there. I thought you said that the BBB acceptance test that writes a bad update doesn't work. Wasn't that why we wanted this fix?

GregorioDiStefano:
Thats correct. Therefore we are currently skipping that test on BBB, but I will reenable it. All other tests are passing on the BBB.

I'm just saying that the image update test is now failing for the qemu/libvirt environment and curious if this can somehow be related.

kacf:
Not to this, that's for sure. Feel free to support a ticket for QEMU failing.

GregorioDiStefano:
Okay, great. I'll look into it a bit and see if I can find out what's going on before opening a ticket.

PR #44, opened 2016-05-23
 Make Mender daemon service log to file system with debug verbosity

GregorioDiStefano:


recipes-mender/mender/files/mender.service
```diff
@@ -5,7 +5,7 @@ Description=Mender OTA update service
 Type=simple
 User=root
 Group=root
-ExecStart=/usr/bin/mender -daemon
+ExecStart=/usr/bin/mender -daemon -log-level=debug -log-file=/var/log/mender.log
```
bboozzoo: I'm concerned that we're adding another log file instead of just consolidating on journal/syslog. Is there anything special about mender logs that they need to be separate from the rest of the system?

GregorioDiStefano: Nope, I will change this soon, and remove the log file

kacf: I think we should not enable this generally. What do you need this for? Testing? You can just enable it in your local build.

Otherwise I agree with using central log files.

GregorioDiStefano: Why shouldn't debug be enabled for the time being?

I mean, the client is under heavily development, and it might be interesting to have it enabled by default.

kacf: "TODO: Remove this" type of tasks have a tendency to stick around longer than anticipated. It can easily get forgotten, and it's not given that everyone needs this (I would use a different approach in my work, for example). If you are debugging issues anyway, then why not enable it in your local build?

PR #46, opened 2016-05-25
Port meta-mender to Yocto daisy branch

kacf:


kacf:
@pasinskim: Can you take a look?

classes/mender-image-buildinfo.bbclass
```diff
@@ -1,15 +1,13 @@
-inherit image-buildinfo
-
 DEVICE_TYPE ?= "${MACHINE}"
 IMAGE_ID = "${PN}-${DATETIME}"
-IMAGE_BUILDINFO_VARS = "DISTRO DATETIME PN IMAGE_ID DEVICE_TYPE"
 
 buildinfo_mender () {
 cat > ${IMAGE_ROOTFS}${sysconfdir}/build_mender << END
```
pasinskim: If we are changing this maybe we  can change the horrible name of this file. But I guess we can do it in separate PR.

classes/sdimg.bbclass
```diff
@@ -39,21 +42,24 @@ IMAGE_BOOT_ENV_FILE ?= "uboot.env"
 # u-boot environment file
 IMAGE_UENV_TXT_FILE ?= "uEnv.txt"
 
+# These are usually defined in the machine section, but for daisy this
+# concept doesn't exist yet.
+IMAGE_BOOT_FILES ?= "u-boot.${UBOOT_SUFFIX}"
+IMAGE_BOOT_FILES_append_beaglebone ?= " MLO"
```
pasinskim: Shouldn't we have `IMAGE_BOOT_FILES_append_beaglebone = " MLO"`?

classes/sdimg.bbclass
```diff
@@ -100,20 +103,88 @@ IMAGE_CMD_sdimg() {
...
+        # COMMIT changes to image file
+        echo p
+        echo w
+
+    ) | fdisk -c=nondos -u=sectors $SDIMG
+
+    dd if=${WORKDIR}/fat.dat of=$SDIMG seek=$PART1_START_SECTORS conv=notrunc
+    dd if=${WORKDIR}/part1.tmp of=$SDIMG seek=$PART2_START_SECTORS conv=notrunc
+    dd if=${WORKDIR}/part2.tmp of=$SDIMG seek=$PART3_START_SECTORS conv=notrunc
+    dd if=${WORKDIR}/data.ext3 of=$SDIMG seek=$PART4_START_SECTORS conv=notrunc
```
pasinskim: You did a nice cleanup with the names on master branch. Maybe something similar here? `BOOT_PART_START`, `INACTIVE_START`, and so on?

kacf: You're right, I should do that here as well.

classes/sdimg.bbclass
kacf: Hmm, my impression is that it's better to leave things so they can be overridden by users of the layer. I'm unsure what's the best practice here. @bboozzoo: Do you have any advice?

classes/mender-image-buildinfo.bbclass
kacf: Yes, that's preexisting, I agree it has a horrible name, but I won't handle that in the PR.

classes/sdimg.bbclass
bboozzoo: IIRC `SOMEVAR ?= "foo"` makes sense for assignments only.

Make sure to keep the append as `SOMEVAR_append_beaglebone` (`beaglebone` coming in after `append`) though.

kacf: Ok, I will switch to `=` then.

kacf:
Issues fixed!

PR #48, opened 2016-05-26
Tests modified since mender is running as daemon

GregorioDiStefano:
2 changes:

1) make sure mender is running before each test
2) remove checks after rebooting, since the daemon should already have commited the changes.

I also added a small bit of code to check the logs for errors, but no point in using right now, since there are spurious errors right now.


kacf:
Removed the "applied/reverted" pair and merged manually!

PR #50, opened 2016-05-31
PV changes

bboozzoo:
Include git revision info in PV. Make sure that agent is rebuilt each time mender repository revision is changed.

@pasinskim @kacf


pasinskim:
Nice!

PR #51, opened 2016-05-31
mender: create /data/mender during startup

bboozzoo:
Automatically create /data/mender path before starting mender service.

@kacf @pasinskim


kacf:
Good idea!

PR #52, opened 2016-06-01
mender: use make for building

bboozzoo:
The changes should be merged only after https://github.com/mendersoftware/mender/pull/63 to mender repository has been merged.

### DIFF
--- a/recipes-mender/mender/mender_0.1.bb
+++ b/recipes-mender/mender/mender_0.1.bb
@@ -36,21 +36,32 @@ do_compile() {
   PATH="${B}/bin:$PATH"
   export PATH
 
-  cd ${S}
-  godep go build -o ${B}/mender
+  # mender is using vendored dependencies, any 3rd party libraries go to into
+  # /vendor directory inside mender source tree. In order for `go build` to pick
+  # up vendored deps from our source tree, the mender source tree itself must be
+  # located inside $GOPATH/src, for instance $GOPATH/src/mender
+  #
+  # recreate temporary $GOPATH/src/mender structure and link our source tree
+  mkdir -p ${B}/src
+  test -e ${B}/src/mender || ln -s ${S} ${B}/src/mender
+  cd ${B}/src/mender
+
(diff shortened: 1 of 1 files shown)


recipes-mender/mender/mender_0.1.bb
```diff
@@ -37,7 +37,7 @@ do_compile() {
   export PATH
 
   cd ${S}
-  godep go build -o ${B}/mender
+  oe_runmake GO='godep go'
```
kacf: We might need to use pure go here after https://github.com/mendersoftware/mender/pull/62, but I'm not sure yet, so we can leave it for now. Otherwise a good change.

recipes-mender/mender/mender_0.1.bb
```diff
@@ -36,8 +36,18 @@ do_compile() {
...
+  # up vendored deps from our source tree, the mender source tree itself must be
+  # located inside $GOPATH/src, for instance $GOPATH/src/mender
+  #
+  # recreate temporary $GOPATH/src/mender structure and link our source tree
+  mkdir -p ${B}/src
+  test -e ${B}/src/mender || ln -s ${S} ${B}/src/mender
+  cd ${B}/src/mender
+
+  # run verbose build, we should see which dependencies are pulled in
+  oe_runmake V=1
```
kacf: I think this doesn't preserve the `-o` functionality. This is important, because if using the Yocto `externalsrc` functionality, the ${S} will be in my home directory, and I don't want it to write there.

kacf: (And in fact I would expect the install step to fail, since it looks in `${B}`)

bboozzoo: Right, this will need another change in mender then.

bboozzoo: Waiting for https://github.com/mendersoftware/mender/pull/64

kacf:
Great!

PR #54, opened 2016-06-07
MEN-440: Configurable partitions for daisy

kacf:


kacf:
I still need to test this before it can be merged.

The master version of this branch is at #55.

kacf:
Tested and works!

classes/mender-install.bbclass
```diff
@@ -1,23 +1,24 @@
...
-IMAGE_INSTALL_append = " u-boot-fw-utils mender ca-certificates"
+# The storage device that holds the device partitions.
+MENDER_STORAGE_DEVICE ?= "/dev/mmcblk0"
 
+# The base name of the devices that hold individual partitions.
+# This is often MENDER_STORAGE_DEVICE + "p".
+MENDER_STORAGE_DEVICE_BASE ?= "/dev/mmcblk0p"
 
+# The numbers of the two rootfs partitions in the A/B partition layout.
+MENDER_PARTITION_A_NUMBER ?= "2"
```
pasinskim: I know that this might be not needed at the moment (or at all), but since we are changing this anyway maybe we should just add `MENDER_ROOTFS_PART_A = "dev/mmcblk0p2"` and `MENDER_ROOTFS_PART_B = "dev/mmcblk0p3"`. The reason is that what will happen if we will decide that active partition will be internal flash and inactive will be SD card. The option of using full partition string will work in all cases.

classes/mender-boot.bbclass
```diff
@@ -0,0 +1,5 @@
+EXTRA_IMAGEDEPENDS += "u-boot u-boot-fw-utils"
+PREFERRED_VERSION_u-boot = "v2015.10%"
+PREFERRED_VERSION_u-boot-fw-utils = "v2015.10%"
+
+IMAGE_INSTALL_append = " u-boot-fw-utils"
```
pasinskim: Will it be possible to just inherit one class and use `mender-boot` only? The reason I am asking is that it is no longer obvious which one you should include in your configuration to have Mender up and running.

recipes-core/base-files/base-files_3.%.bbappend
```diff
@@ -1,8 +1,20 @@
-FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
+FILESEXTRAPATHS_prepend_menderimage := "${THISDIR}/${PN}:"
 
 #TODO: ${@bb.utils.contains("MACHINE_FEATURES", "sd", "packagegroup-mender-sd", "", d)}
 #./meta/recipes-core/packagegroups/packagegroup-base.bb
-do_install_append () {
+
+do_patch_device() {
+    # Noop if we're not building a mender-image.
+    true
```
pasinskim: Maybe @bboozzoo can comment here, but do we need this? I think I haven't seen anything similar so far...

recipes-mender/mender/files/mender.conf
```diff
@@ -1,12 +1,13 @@
...
-  "ServerCertificate": "@MENDER_CERT_LOCATION@",
   "ClientProtocol": "https",
+  "DeviceID": "DEVICE_ID",
   "HttpsClient": {
     "Certificate": "",
     "Key": ""
-  }
+  },
+  "PartitionANumber": "@MENDER_PARTITION_A_NUMBER@",
+  "PartitionBNumber": "@MENDER_PARTITION_B_NUMBER@",
```
pasinskim: Similar as above, maybe instead of numbers we should use whole strings defining partition name?

classes/mender-install.bbclass
kacf: As discussed in the office we cannot do this because U-Boot doesn't understand Unix notation.

recipes-core/base-files/base-files_3.%.bbappend
kacf: Yes, I would be interested to know, I also find this a bit kludgy.

classes/mender-boot.bbclass
kacf: It is `mender-full` you need to include to get everything. But maybe it would be better if this class was called `mender-uboot`, not `mender-boot`.

classes/mender-install.bbclass
kacf: Hmm, crap, I'm running into a different problem though: We need to know in U-Boot as well which device we should load from, and we cannot use `mmcblk` notation. This is tricky...

kacf: Ok, as discussed in the office, I will change this to accept `/dev/mmcblk` notation, and simply reject patterns that don't match that one. Then we can gradually accept more patterns as we add support for storage devices.

kacf:
@pasinskim, @bboozzoo: Alright, I believe I have addressed all the issues, including the ones from #55. I will provide the master version too, but I'll stick to daisy until it is reviewed and merged.

bboozzoo:
I'm still not happy with how systemd and DISTRO_FETURES are played with, but we'll have to live with it. The rest of the patches look good to me.

kacf:
Yeah, to tell you the truth I'm not that happy either. It's messy. But it's taken so much time already, we should get this merged and we can improve as we go. Thanks for reviewing!

kacf:
Unfortunately, I can't merge this yet because:

1. This depends on https://github.com/mendersoftware/mender/pull/67.
2. https://github.com/mendersoftware/mender/pull/67 depends on https://github.com/mendersoftware/meta-mender/pull/55.
3. https://github.com/mendersoftware/meta-mender/pull/55 depends on me adapting the patches to master.
4. Adapting the patches to master requires waiting for the build and fixing. So it's WIP.

But otherwise this is ready.

PR #55, opened 2016-06-07
MEN-440: Configurable partitions

kacf:


kacf:
I still need to test this before it can be merged.

The daisy version of this branch is at #54.

kacf:
Tested and works!

pasinskim:
See my comments here: https://github.com/mendersoftware/meta-mender/pull/54. Most are relevant here as well.

classes/sdimg.bbclass
```diff
@@ -42,6 +42,11 @@ IMAGE_BOOT_ENV_FILE ?= "uboot.env"
 # u-boot environment file
 IMAGE_UENV_TXT_FILE ?= "uEnv.txt"
 
+# This will be embedded into the boot sector, where depends on the offset
+# variable.
+IMAGE_BOOTLOADER_FILE ?= "u-boot.${UBOOT_SUFFIX}"
+IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET ?= "2"
```
bboozzoo: Document that offset is in bs=512

classes/sdimg.bbclass
```diff
@@ -120,6 +125,16 @@ EOF
     # Call WIC
     IMAGE_CMD_wic
 
+    # Embed boot loader in boot sector.
```
bboozzoo: Might want to edit the comments and commit message, technically only the first 512 bytes is the boot sector. instead, we're actually writing u-boot (with FSL's header) at a specific offset so that iMX.6 ROM will find it.

classes/mender-install.bbclass
```diff
@@ -13,11 +6,4 @@ VIRTUAL-RUNTIME_init_manager = "systemd"
...
 
-
-#Make sure we are creating sdimg with all needed partitioning.
-IMAGE_CLASSES += "sdimg"
-IMAGE_FSTYPES += "sdimg"
-
-IMAGE_INSTALL_append = " u-boot-fw-utils mender ca-certificates"
-
-
+IMAGE_INSTALL_append = " mender ca-certificates"
```
bboozzoo: This has bothered me for some time already. I'm not convinced that classes should be used to install packages into the image. In my opinion mender should be installed by `*.bbappend` to one of the base images.

recipes-core/base-files/base-files_3.%.bbappend
```diff
@@ -1,8 +1,20 @@
-FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
+FILESEXTRAPATHS_prepend_menderimage := "${THISDIR}/${PN}:"
 
 #TODO: ${@bb.utils.contains("MACHINE_FEATURES", "sd", "packagegroup-mender-sd", "", d)}
 #./meta/recipes-core/packagegroups/packagegroup-base.bb
-do_install_append () {
+
+do_patch_device() {
```
bboozzoo: Why have a task that does nothing?

recipes-core/base-files/base-files_3.%.bbappend
```diff
@@ -1,8 +1,20 @@
...
+    # Noop if we're not building a mender-image.
+    true
+}
+do_patch_device_menderimage() {
+    if [ -n "${MENDER_STORAGE_DEVICE_BASE}" ]; then
+        sed -i -e 's,/dev/mmcblk0p,${MENDER_STORAGE_DEVICE_BASE},g' ${S}/fstab
```
bboozzoo: Can you explain what is the purpose here? What if there's no `/dev/mmcblk0p*` in fstab?

classes/mender-install.bbclass
```diff
@@ -1,6 +1,10 @@
 PREFERRED_VERSION_go_cross = "1.6%"
 
-#Add support for systemd
+# Add support for systemd
+#
+# Note that a very common error with systemd is that the system hangs during the
+# boot process while looking for devices. This is almost always because the
+# kernel feature CONFIG_FHANDLE is not enabled.
 DISTRO_FEATURES_append = " systemd"
```
bboozzoo: Another thing that has bothered me for some time, Clas is not a suitable place to break my distro features. If you really need to play with `DISTRO_FEATURES` make your own reference distro.

classes/sdimg.bbclass
kacf: I will!

classes/sdimg.bbclass
kacf: Right, that might be more technically correct.

classes/mender-install.bbclass
kacf: But how do you then deal with mender being included on a variety of images? Are they all based on a common image base, say core-image-minimal?

recipes-core/base-files/base-files_3.%.bbappend
kacf: The intention is to make the device configurable. But it should perhaps be made clearer here by replacing `@MENDER_STORAGE_DEVICE_BASE@` instead of `mmcblk0p`. I can change that.

recipes-core/base-files/base-files_3.%.bbappend
kacf: This is what we were discussing in the other ticket. The only way I could find to do conditionals in the bitbake recipe is via `OVERRIDES`. But since the task will be added to the run queue regardless of whether `menderimage` triggers or not, I needed an empty one to make it happy. But this is probably suboptimal, if you know a better way I'm all for changing it.

classes/mender-install.bbclass
kacf: Well, it's not like we depend on a whole distro, we just depend on certain features of the existing distros. What would be the appropriate way, produce error if it's not included?

recipes-core/base-files/base-files_3.%.bbappend
kacf: Oh, right, and to be clear, the fstab file comes from our repository, so we know the string is clear. But yes, this isn't obvious. I'll change it.

classes/mender-install.bbclass
bboozzoo: Firstly, our integration is limited to shipping `mender.service`, so that's not a big deal. We may equally well ship a simple sysvinit script and work with poky.

Try to follow the principle of least surprise. Integration with systemd is a rather big deal, and people end up running into different problems, even delaying/dropping the work. Now, if one ends up evaluating mender with an existing non-systemd distribution (or even poky), things may blow up. Tracking the that back to `mender-install.bbclass` will be surprising to say the least.

classes/mender-install.bbclass
bboozzoo: Try bbappending `packagegroup-core-boot`. Mender kind of interacts with the boot process, so it might fit wit the other packages in `packagegroup-core-boot`.

It's also good if we document explicitly that `mender` package will be added to `core-boot` group, so that people are not surprised.

BTW. I think that we don't really have a dependency on `ca-certificates`. First we're shipping our own server certificate. Secondly, as far as I know, Go stdlib does not look for certificates in `${sysconfdir}/ca-certificates` or `${datadir}/ca-certificates`

classes/mender-install.bbclass
kacf: Yes, agree, switching to systemd shouldn't be done behind your back. We decided that we didn't wanna deal with sysv scripts (this was backed by Eystein's field surveys), so I propose we just produce an error then. Do you know of a mechanism for this in Yocto, or should we simply test the `DISTRO_FEATURES` variable in a task snippet?

classes/mender-install.bbclass
kacf: I suspect the ca-certificates was in order for S3 links to work.

`packagegroup-core-boot` sounds like a good idea, but this is a pretty big task, not really related to what I'm doing in this PR, so I won't fix it here. But it's something to consider for later.

classes/mender-install.bbclass
bboozzoo: Just document that systemd is needed then, but neven enable it in some random place.

Also update mender recipe to ship mender.service only if systemd is enabled in DISTRO_FEATURES. You can check like this:
```
if ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'true', 'false', d)}; then
    install -t ${systemd_unitdir}system/..
fi
```

classes/mender-install.bbclass
bboozzoo: > packagegroup-core-boot sounds like a good idea, but this is a pretty big task, not really related to what I'm doing in this PR, so I won't fix it here. But it's something to consider for later.

Can you add that to JIRA? I can the this task once I'm done with the agent, it looks like a quick fix.

kacf: [Done](https://tracker.mender.io/browse/MEN-448)! Feel free to add more clarification to it.

bboozzoo: Thanks.

classes/mender-install.bbclass
kacf: Right now I'm going for the approach where, if you inherit `mender-full` it produces error if you haven't enabled systemd, but you can omit that by selectively inheriting only the mender classes you need. For QEMU it will not produce error, but will always enable systemd, for convenience. I think this is a good usability vs surprise compromise. Do you agree?

PR #56, opened 2016-06-09
Change the name of boot partition to /uboot.

pasinskim:
As there is some limitation in older versions of u-boot that
environment utilities (fw_printenv and fe_setenv) can read environment
location consisting of 16 characters at most.
Changing partition name from /u-boot to /uboot will cause that
'/uboot/uboot.env' string will be 16 characters long.


pasinskim:
@kacf this will cause some changes in HIO stuff you are working with as well.

kacf:
Looks good. This will probably need a patch for daisy as well.

pasinskim:
@kacf I've tested it and works fine with QEMU.

PR #58, opened 2016-06-14
Simplify testing with server.

pasinskim:


kacf:
This will need to be changed, but alright for now.

PR #60, opened 2016-06-15
Make it possible to use 'file://' URIs even in externalsrc builds.

kacf:
The problem is that externalsrc causes *all* source fetching to be
skipped, even files that are local. These files are not mirrored in
the mender repository, so it makes it quite awkward to work it. Adding
manual copy rules for those files is too error prone, since we're
likely to add more files later.

So instead, do custom fetching of all files in SRC_URI that
specifically start with 'file://'.

### DIFF
--- a/recipes-mender/mender/mender_0.1.bb
+++ b/recipes-mender/mender/mender_0.1.bb
@@ -30,6 +30,35 @@ FILES_${PN} += "${systemd_unitdir}/system/mender.service \
                 ${sysconfdir}/mender.conf \
                "
 
+do_configure() {
+  # Move/copy all file:// URIs into WORKDIR. We need to manually move/copy them
+  # because externalsrc builds do not automatically fetch local files from the
+  # bb recipe, and we need to put them in WORKDIR because we should not copy
+  # them into S if it is external.
+
+  # Bitbake quoting rules mean we need to make an intermediate variable here.
+  SRC_URI='${SRC_URI}'
+  for file in $SRC_URI; do
+    case $file in
+      file://)
+        if [ -n "${EXTERNALSRC}" ]; then
+          for path in ${FILESPATH}; do
+            if [ -e $path/$file ]; then
+              cp -r $path/$file ${WORKDIR}
+              break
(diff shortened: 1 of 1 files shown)


kacf:
@GregorioDiStefano: This should solve the Jenkins build problem where it doesn't find `mender.conf`.

@bboozzoo: Unless you know of a better way...

bboozzoo:
I'm sorry but it makes no sense to me. Nowhere in the recipe there's `inherit externalsrc`. If you do fancy stuff that breaks the recipe in your CI setup, then your CI setup is the right place to fix it.

The way it's usually done is that you have a separate layer just for CI purposes. There you can override the recipe with `*.bbappend `as needed, but still in isolation.

kacf:
No, the recipe itself is not expected to have `inherit externalsrc`, this is something that almost always goes into `local.conf`, as an alternative way of building. It is used in the `local.conf` of the Jenkins configuration, as well as on my own computer in daily Mender development. This would break for everyone using `externalsrc`, not just the CI.

We won't get away from supporting `externalsrc` in the recipe, the question is more about how to do it.

bboozzoo:
Nobody except from you and CI is expected to build mender from external source. Everyone else is supposed to fetch sources from github. Am I missing something?

kacf:
And they will continue to do so. Is it wrong to support an alternative build method in the recipe? I fail to see why that is a problem. `externalsrc` is a well documented feature of poky...

bboozzoo:
First things first, you're trying to fix your local issue (yours and CI's) in a recipe that is used by everyone, by adding code that fetches the files in `do_configure` task (when fetching is to be done in `do_fetch`), in a way that nobody else does.

Secondly, `externalrsc` will fetch your local files, there's nothing you have to do, see this: https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/classes/externalsrc.bbclass#n30

Lastly, the build does not work without externalsrc. so the root of the problem is elsewhere.

kacf:
Superseded by #61.

PR #64, opened 2016-06-20
Move build_mender to mender directory.

kacf:


kacf:
@pasinskim 

Should we also find another name for the file itself?

PR #65, opened 2016-06-21
Give boot partition a more reasonable size.

kacf:


classes/mender-sdimg.bbclass
```diff
@@ -26,7 +26,7 @@ SDIMG_DATA_PART_DIR ?= ""
 SDIMG_DATA_PART_SIZE_MB ?= "128"
 
 # Size of the first (FAT) partition, that contains the bootloader
-SDIMG_BOOT_PART_SIZE_MB ?= "128"
+SDIMG_BOOT_PART_SIZE_MB ?= "8"
```
bboozzoo: 16 maybe? No strong preference here.

kacf: Maybe. But currently we're only storing the U-Boot environment and the u-boot binary there, which are both much less than 8M.

PR #66, opened 2016-06-21
mender-image: make sure that required FSTYPES are present

bboozzoo:
@kacf @pasinskim

### DIFF
--- a/classes/mender-image.bbclass
+++ b/classes/mender-image.bbclass
@@ -9,4 +9,4 @@ inherit mender-image-buildinfo
 
 #Make sure we are creating sdimg with all needed partitioning.
 IMAGE_CLASSES += "mender-sdimg"
-IMAGE_FSTYPES ?= "tar.gz ext3 sdimg"
+IMAGE_FSTYPES_append = " sdimg"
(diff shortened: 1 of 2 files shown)


classes/mender-image.bbclass
```diff
@@ -9,4 +9,4 @@ inherit mender-image-buildinfo
 
 #Make sure we are creating sdimg with all needed partitioning.
 IMAGE_CLASSES += "mender-sdimg"
-IMAGE_FSTYPES ?= "tar.gz ext3 sdimg"
+IMAGE_FSTYPES_append = " sdimg"
```
kacf: Right, there we have it!

kacf: Hmm, one thing I didn't consider is that this will make it impossible to *not* build the image. This may in fact be quite common, since the subsequent updates only need the rootfs, not the sdimg. Any thoughts on this?

bboozzoo: IMO this shouldn't be a large problem. The user can always not inherit mender-image, since the class is only useful when building the base/initial image.

PR #68, opened 2016-06-21
Move vexpress specific systemd config to vexpress-qemu.conf file

GregorioDiStefano:
@kacf

### DIFF
--- a/classes/mender-systemd.bbclass
+++ b/classes/mender-systemd.bbclass
@@ -1,15 +1,11 @@
 # Class to help deal with systemd and Mender.
 
-# Suggested method for adding support for systemd, enabled by default for QEMU.
+# Suggested method for adding support for systemd, enabled by default for QEMU (vexpress-qemu.conf)
 #
 #DISTRO_FEATURES_append = " systemd"
 #VIRTUAL-RUNTIME_init_manager = "systemd"
 #DISTRO_FEATURES_BACKFILL_CONSIDERED = "sysvinit"
 #VIRTUAL-RUNTIME_initscripts = ""
-DISTRO_FEATURES_append_vexpress-qemu = " systemd"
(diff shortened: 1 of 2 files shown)


bboozzoo:
Machine config defines hardware policy and settings. systemd is not hardware. Mangling distro features in `mender-systemd` class is bad, but moving that to machine config is even worse. 

I've already complained to @kacf about that, but if you reallly need to enable systemd, although mender has very little dependency on systemd, define your own distro and base it on poky. In practice people will already have defined their own distros and/or machine configs, so you'll likely to break their setups. 

Not sure how I can phrase it to make the message more clear, but it's like hiding pieces of configuration that have wide ranging effects at some random locations against all good practices.

kacf:
I know, we don't enable it anywhere else, but vexpress-qemu is a demo machine, not used for production, so we just want the smoothest possible path.

kacf:
@bboozzoo: Btw, you mentioned earlier that it's possible to bbappend the core image types. We could do it there instead if it makes you more comfortable, but ultimately it would achieve the same thing.

bboozzoo:
bbappending is equally bad, we're trying to set a system wide policy here, bc that's what enabling systemd is.

Why not define our own distro, call it `mender` and use it in our testing? You can start with something as simple as https://github.com/open-rnd/meta-openrnd/blob/master/conf/distro/openrnd-poky-systemd.conf

As for Poky, systemd is not enabled there, so we're technically not compatible with Poky.

kacf:
Alright, I get the point of not messing with this outside of `mender-systemd.bbclass`. I still think that a distro is too heavyweight for we're trying to do here. Basically all we want to express is:

```
DEPENDS_mender += "systemd"
```

But because of systemd's pervasive nature it doesn't work like that.

So I've gone for the simple fix now, check https://github.com/mendersoftware/meta-mender/pull/70.

PR #70, opened 2016-06-24
Fix incorrect error for systemd when compiling vexpress-qemu.

kacf:
It is actually enabled for vexpress-qemu by default, but because of
the way bitbake variables are resolved, it comes too late for the test
to pick it up, so skip the check in that case.


kacf:
Tested on both vexpress and beaglebone.

PR #71, opened 2016-06-24
Various fixes from hio port

kacf:


kacf:
@bboozzoo

bboozzoo:
Seems ok, only make sure BBB is not broken by accident.

GregorioDiStefano:
This commit is actually suppose to fix the broken BBB :)

PR #74, opened 2016-06-27
Dockerize meta-mender Yocto image

GregorioDiStefano:
This allows us to run the mender qemu environment inside of Docker.

As soon as I get a hold of the docker hub credentials, I will integrate this with the build machine, and push the latest container to docker hub.
 
I've added this since it would facilitate running the latest mender qemu image, and we will need this for integration testing. 

I also changed mender-qemu with some default parameters, and use qemu-system-arm from PATH.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM ubuntu:trusty
+
+RUN apt-get -y update
+RUN apt-get -y upgrade
+
+#Use: --build-arg VEXPRESS_IMAGE=core-image-full-cmdline-vexpress-qemu.sdimg --build-arg UBOOT_ELF=u-boot.elf when building
+ARG VEXPRESS_IMAGE
+ARG UBOOT_ELF
+
+# Install QEMU/KVM
+RUN apt-get -y install qemu-kvm qemu-system-arm
+
+ADD $UBOOT_ELF ./
+ADD $VEXPRESS_IMAGE ./
+
+ADD scripts/mender-qemu ./
+ADD scripts/docker/entrypoint.sh ./
+
+RUN chmod +x entrypoint.sh mender-qemu
+EXPOSE 8822
+ENTRYPOINT ["./entrypoint.sh"]
--- a/scripts/mender-qemu
+++ b/scripts/mender-qemu
@@ -6,7 +6,7 @@
 #
 # if image name is not provided, the script will attempt to run
 # core-image-full-cmdline
-
+#
 
 # NOTE: current settings forward
 #           ssh: port 8822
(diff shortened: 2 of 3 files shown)


scripts/mender-qemu
```diff
@@ -31,18 +31,21 @@ else
...
 else
     BUILDTMP=${BUILDDIR}/tmp
 fi
 
+UBOOT_ELF=${UBOOT_ELF:="${BUILDTMP}/deploy/images/vexpress-qemu/u-boot.elf"}
+VEXPRESS_SDIMG=${VEXPRESS_SDIMG:="${BUILDTMP}/deploy/images/vexpress-qemu/${IMAGE_NAME}-vexpress-qemu.sdimg"}
+
 QEMU_AUDIO_DRV=none \
-    ${BUILDTMP}/sysroots/x86_64-linux/usr/bin/qemu-system-arm \
+    qemu-system-arm \
```
bboozzoo: Don't we want to use qemu binary that was built from our layer by OE?

GregorioDiStefano: I changed this in a later commit:

https://github.com/GregorioDiStefano/meta-mender/blob/509761d1eab721fd3ca6f3ad9aefdacac2a12362/scripts/mender-qemu

Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ubuntu:trusty
+
+RUN apt-get -y update
+RUN apt-get -y upgrade
+
+#Use: --build-arg VEXPRESS_IMAGE=scripts/core-image-full-cmdline-vexpress-qemu.sdimg --build-arg UBOOT_ELF=u-boot.elf when building
```
bboozzoo: How come these files end up in scripts? Maybe provide a script for building docker image so that we don't need to set these variables explicitly in the command line.

GregorioDiStefano: The comment was actually an error, the file isn't in the scripts directory..

GregorioDiStefano:
updated

scripts/mender-qemu
```diff
@@ -37,12 +37,16 @@ else
     BUILDTMP=${BUILDDIR}/tmp-glibc
 fi
 
+UBOOT_ELF=${UBOOT_ELF:="${BUILDTMP}/deploy/images/vexpress-qemu/u-boot.elf"}
+VEXPRESS_SDIMG=${VEXPRESS_SDIMG:="${BUILDTMP}/deploy/images/vexpress-qemu/${IMAGE_NAME}-vexpress-qemu.sdimg"}
+QEMU_SYSTEM_ARM=${QEMU_SYSTEM_ARM:="${BUILDTMP}/sysroots/x86_64-linux/usr/bin/qemu-system-arm"}
```
kacf: Use `:-`, not `:=`. The latter has some surprising semantics, and although it doesn't matter in this particular case, it's better style to use the former.

kacf: Basically if you had written:

```
MY_VAR=${UBOOT_ELF:="${BUILDTMP}/deploy/images/vexpress-qemu/u-boot.elf"}
```

the string would be assigned to both `MY_VAR` *and* `UBOOT_ELF`, which isn't what you would normally expect.

kacf:
Looks good otherwise!

PR #84, opened 2016-07-04
Fix several issues with oe-meta-go layer integration.

kacf:
Get rid of a lot of custom patches.

However, we need one patch, where ${GOROOT_FINAL} has been mistakenly
used without ${D}. Without this, the recipe will install directly into
the sysroot, bypassing the caching and tracking system of bitbake.
This is bad from a consistency perspective, but also has additional
side effects: Since we not only write into, but also read from, this
location, other packages can influence what happens, and in a parallel
build, this depends on timing. It would sometimes lead to errors where
go-cross's files were conflicting with those of gcc, just because at a
certain point in time, the go-cross recipe would run a copy that then
included files from other packages.


GregorioDiStefano:
trigger build

GregorioDiStefano:
Merging this since it works, and I need it to move forward ;)

PR #89, opened 2016-07-06
Dockerfile: reduce docker image size

bboozzoo:
The current images are large ~700MB, this should reduce the size of
intermediate snapshots thus trimming the image to something like ~220MB.

@GregorioDiStefano can you try that in your environment?


kacf:
Lol, apt takes that much? That's hilarious! Nice one!

PR #91, opened 2016-07-07
Changing test condition after error output changed.

kacf:


kacf:
@GregorioDiStefano: I'm merging this immediately, which will temporarily break tests, but it's in anticipation of merging https://github.com/mendersoftware/mender/pull/83, which then restores them.

PR #93, opened 2016-07-12
sysvinit support & systemd no auto enable

bboozzoo:
The series starts witha small cleanup of how we handle B. Right now we just use the default `B=${S}`, but that's a bit inconvenient due to fake $GOPATH we need to create for the build, thus polluting `${S}`.

The second patch will prevent mender service being automatically enabled in systemd setups.

The last patch adds a sysvinit startup script and required recipe support. Similarly to systemd variant, the service is disabled by default.

@pasinskim @kacf @eysteins @GregorioDiStefano

### DIFF
--- a/recipes-mender/mender/mender_0.1.bb
+++ b/recipes-mender/mender/mender_0.1.bb
@@ -15,6 +15,7 @@ SRC_URI = "git://github.com/mendersoftware/mender;protocol=https \
            file://mender.service \
            file://mender.conf \
            file://server.crt \
+           file://sysvinit-mender \
           "
 
 SRCREV = "${AUTOREV}"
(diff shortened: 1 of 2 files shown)


recipes-mender/mender/mender_0.1.bb
```diff
@@ -32,6 +32,7 @@ SYSTEMD_SERVICE_${PN} = "mender.service"
 FILES_${PN} += "${systemd_unitdir}/system/mender.service \
                 ${sysconfdir}/mender.conf \
                "
+SYSTEMD_AUTO_ENABLE = "disable"
```
kacf: I think we should not make patches that knowingly break production functionality, even if it is temporary. Either the tests should build a separate image, or modify their own copy of the original (which I believe they do ATM).

recipes-mender/mender/mender_0.1.bb
```diff
@@ -7,13 +7,15 @@ MENDER_CERT_LOCATION ?= ""
 #From oe-meta-go (https://github.com/mem/oe-meta-go)
 DEPENDS = "go-cross godep"
 S = "${WORKDIR}/git"
+B = "${WORKDIR}/build"
```
kacf: +1

recipes-mender/mender/mender_0.1.bb
bboozzoo: Yeah, ideally they should just disable it in bbappend, but I'm completely lost at this as to what we actually want. There are conflicting opinions about what's good for integration testing and customers doing evaluation testing. I have no preference here.

kacf:
> mender: sysvinit support

Nice! With this we can remove the horrible systemd hackery. Although it is an extra configuration to test... @GregorioDiStefano: Should we consider multiple configurations?

@bboozzoo: You should put [MEN-525](https://tracker.mender.io/browse/MEN-525) in the commit message for that one, and close that task when you're done.

recipes-mender/mender/mender_0.1.bb
kacf: I just talked to @GregorioDiStefano, he's working around it in the tests, so it's alright to leave this out. The published image definitely needs to have Mender enabled, so I think that is what our guide should be.

kacf:
Decided to drop this for the time being after discussing internally, but we may return to it later.

PR #94, opened 2016-07-12
Get rid of specialized U-Boot recipes.

kacf:
It is not needed anymore.


kacf:
Don't merge yet, this is untested and will probably break on older branches.

kacf:
trigger build

kacf:
trigger build

PR #97, opened 2016-07-15
mender: set build dir

bboozzoo:
Poky bitbake setup defaults to B=S what is inconvenient in case of
mender as we need to create a mock GOPATH structure for the sake of
build. Make sure to set B to something remotely reasonable.


kacf:
trigger build

PR #98, opened 2016-07-18
Fix all hardcoded test paths to use proper workspace.

kacf:


kacf:
trigger build

PR #99, opened 2016-07-18
Pick up secret credentials from common location instead of each works…

kacf:
…pace.


kacf:
trigger build

kacf:
trigger build

PR #101, opened 2016-07-21
Use DISTRO_FEATURES for Mender features instead of classes.

kacf:
The big advantages are that they are easier to test for in snippets,
and one can selectively both enable and disable features. Classes
OTOH, can only be added; if you want to remove one, then you have to
remove "mender-full" and list every feature *but* the one you want.

This also changes a bit how the layer works. You no longer need to
inherit anything for the layer to work, instead you can selectively
disable features by using DISTRO_FEATURES_BACKFILL_CONSIDERED.


kacf:
Don't merge this, it's not tested enough. I'm just putting it up so people can see.

kacf:
Closing for now, this is unlikely to happen soon. Ref https://tracker.mender.io/browse/MEN-576.

PR #102, opened 2016-07-21
Don't reference WORKSPACE which doesn't exist in local builds.

kacf:
We assume that the tests are always executed from the test directory
anyway.


kacf:
trigger build

PR #103, opened 2016-07-22
MEN-546: Fix hardcoded filesystem type in sdimg creation.

kacf:
Use first ext2/3/4 filesystem found in IMAGE_FSTYPES instead.


kacf:
trigger build

kacf:
Ported to daisy as well.

PR #104, opened 2016-07-27
Removing deprecated DeviceID key

GregorioDiStefano:
@bboozzoo I see DeviceID is no longer referenced in the client, I think this was lingering artifact


bboozzoo:
:+1:

PR #87, opened 2016-07-06
Remove superfluous addition of fw_env.config.

kacf:
This is already handled inside u-boot-mender.inc.

Signed-off-by: Kristian Amlie <kristian.amlie@mender.io>

PR #106, opened 2016-08-08
build with poky master

bboozzoo:
Fixes build with poky master, builds on top of #87. `fw_env.config` will be generated and shipped by `u-boot-fw-utils` only.

@kacf @GregorioDiStefano @pasinskim


kacf:
trigger build

PR #107, opened 2016-08-08
mender: tenant token handling

bboozzoo:
Automatic setup of tenant token when starting mender client. The user
should override MENDER_TENANT_TOKEN and set it to value of the token
assigned by the system.

### DIFF
--- a/recipes-mender/mender/files/mender.service
+++ b/recipes-mender/mender/files/mender.service
@@ -6,6 +6,7 @@ Type=simple
 User=root
 Group=root
 ExecStartPre=/bin/mkdir -p -m 0700 /data/mender
+ExecStartPre=/bin/ln -sf /etc/mender/tenant.conf /var/lib/mender/authtentoken
 ExecStart=/usr/bin/mender -daemon
 Restart=on-abort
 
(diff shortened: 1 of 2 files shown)


recipes-mender/mender/files/mender.service
```diff
@@ -6,6 +6,7 @@ Type=simple
 User=root
 Group=root
 ExecStartPre=/bin/mkdir -p -m 0700 /data/mender
+ExecStartPre=/bin/ln -s /etc/mender/tenant.conf /var/lib/mender/authtentoken
```
kacf: Won't `ln` fail if the link already exists? Quote from systemd docs:

> If any of those commands (not prefixed with "-") fail, the rest are not executed and the unit is considered failed.

So the unit may only start on the first boot, but not thereafter.

kacf: And the fix appears to be adding `-` in front, like in a Makefile.

bboozzoo: Where do you see `-` ??

bboozzoo: Replaced with `ln -sf ..`

kacf: The quote I gave mentions the prefix "-", which seems to be used to silence such failures, at least the way I read it. But using `-sf` also works.

kacf:
trigger build

PR #109, opened 2016-08-10
Remove need to set two storage device variables when one will do.

kacf:


kacf:
trigger build

kacf:
trigger build

PR #115, opened 2016-08-12
Fix bug which causes subtle differences between stock and forked u-bo…

kacf:
…ots.

Basically we were asking people to include u-boot-mender.inc, but this
would leave out some key parts that were hidden in the bbappend files
that we use in meta-mender for stock u-boot and u-boot-fw-utils. Since
the sections for the two recipes are different, we have no choice but
to split the include file in two and ask people to include the right
one from either recipe.

In addition added a section which tries to detect inclusion of the
wrong file, since this used to be the old way.


kacf:
This is mostly moving stuff. It becomes much more readable with `git show -C -C -M`.

kacf:
trigger build

kacf:
trigger build

kacf:
Note to self: Also split generic patch.

PR #118, opened 2016-08-18
Make sure Mender runs after network-online target is met

GregorioDiStefano:
a HTTP call right after starting mender will fail, because the network stack is not completely up and functional.

note: if we would move to sysv, we'd have to adjust that script as well, maybe it's smarter to fix this in the client?

### DIFF
--- a/recipes-mender/mender/files/mender.service
+++ b/recipes-mender/mender/files/mender.service
@@ -1,5 +1,7 @@
 [Unit]
 Description=Mender OTA update service
+After=network-online.target network.target
+Requires=network-online.target
 
 [Service]
 Type=simple


kacf:
One concern though: If the network doesn't come back at all, Mender won't start, and rollback won't work.

One solution is then to put it in the client instead. Another is to assume that it's the watchdog's responsibility to do this. But I'm unsure if watchdogs care about network connectivity. I'm not sure what's the best approach here.

@pasinskim, @eysteins: Opinions?

pasinskim:
@kacf that is a good point. Let's think how to solve it.

kacf:
Closing this for now. I think we agree that we need to solve this in a different way.

PR #123, opened 2016-08-24
Fix U-Boot bootargs variable not being expanded properly.

kacf:
Since bootargs is only used implicitly by various boot commands, and
not explicitly like most "run" commands, it is not expanded when it is
used. This is a problem if for example ${mender_kernel_root} is
hardcoded directly in bootargs, without ever being explicitly set (and
thereby evaluated) by the script.

See also the U-Boot documentation: "Note: You cannot use this method
directly to define for example the "bootargs" environment variable, as
the implicit usage of this variable by the "bootm" command will not
trigger variable expansion - this happens only when using the "setenv"
command." (originally fetched from
http://www.denx.de/wiki/view/DULG/LinuxBootArgs)

We work around this by simply setting bootargs to itself, thereby
forcing reevaluation.


kacf:
Already tested by myself and a user on the mailing list.

PR #124, opened 2016-08-25
Set StrictHostKeyChecking=no when creating BBB ssh tunnel

GregorioDiStefano:


kacf:
It might be a good idea to add `-o BatchMode=yes` as well. Guarantees that it will never be stuck at a password prompt.

PR #128, opened 2016-09-06
Change PollIntervalSeconds to 5 seconds

GregorioDiStefano:


GregorioDiStefano:
@kacf

kacf:
It's for every build. It's not a great setting, but we figured that probably everyone under the sun will modify this anyway. I expect that even 60 seconds is way too frequent for many people.

GregorioDiStefano:
can this be merged?

PR #129, opened 2016-09-09
Move identity script to /usr/share/mender/identity.

kacf:
It's not needed in PATH, and it's better to keep Mender's private
data/scripts isolated. This also brings it in line with the upcoming
inventory scripts.


bboozzoo:
Since https://github.com/mendersoftware/mender/pull/108 is merged we need to merge this PR too

PR #131, opened 2016-09-19
Fix broken Yocto build.

pasinskim:
The checksum for license file seems to be incorrect.


pasinskim:
@kacf I am merging this for now. Feel free to revert if this should not go in.

kacf:
I checked the license, should be ok, but thanks for notifying!

PR #135, opened 2016-09-20
Add example inventory script.

pasinskim:
Signed-off-by: Marcin Pasinski <marcin.pasinski@cfengine.com>


pasinskim:
This looks to be already there.

PR #138, opened 2016-09-21
sdimg build fixes

bboozzoo:
sdimg can be properly built for me using recent poky master (49a7839)

### DIFF
--- a/classes/mender-sdimg.bbclass
+++ b/classes/mender-sdimg.bbclass
@@ -169,7 +169,7 @@ EOF
 
     # Embed boot loader in image, offset relative to boot sector.
     if [ -n "${IMAGE_BOOTLOADER_FILE}" ]; then
-        if [ $(expr ${SDIMG_PARTITION_ALIGNMENT_MB} \* 1048576 - ${IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET} \* 512) -lt $(stat -c %s ${IMAGE_BOOTLOADER_FILE}) ]; then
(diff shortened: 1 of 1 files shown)


classes/mender-sdimg.bbclass
```diff
@@ -167,15 +167,18 @@ EOF
     # Call WIC
     IMAGE_CMD_wic
 
+    wicimgname="${IMGDEPLOYDIR}/${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.wic"
```
kacf: Is `IMGDEPLOYDIR` documented anywhere? I can't find it in the Yocto documentation. Is it one of the "secret" variables that wic uses?

kacf: Oh, right I see the poky patch mentions it.

But I expect this will be incompatible with krogoth then? Which means we'll have to diverge the branches, unfortunately.

bboozzoo: Yeah, I'm building with master only. We'll need to cherry pick 9793035a68434a0da1b3f4633a0de3a8d9be3a1a for krogoth manually.

bboozzoo: BTW. a similar patch was recently submitted for meta-raspberrypi https://lists.yoctoproject.org/pipermail/yocto/2016-September/032086.html

kacf: I'm using merge strategy for these branches, since they are 99% identical (actually 100% until now), so I'll merge as usual and revert the krogoth-incompatible patch specifically.

kacf:
Merged into, and then reverted 640e5dc1882bdf34afb6da79c37e11c7c9615380 from, krogoth.

PR #140, opened 2016-09-27
make qemu vm only allocate 64mb of ram

GregorioDiStefano:
This will make using `docker-compose scale` more useful.

Currently, after booting up qemu with the vexpress image, only 12mb of RAM is used.


GregorioDiStefano:
@pasinskim @kacf

kacf:
Geez, 512 even!

PR #164, opened 2016-10-20
Test improvements

kacf:
```
commit e00aa6fab6d93465b8eaf74aabeff912273e5ebc
Author: Kristian Amlie <kristian.amlie@mender.io>
Date:   Wed Oct 19 12:59:34 2016

    Add test for IMAGE_BOOTLOADER_FILE bitbake option.

    At the time of writing, this test will in fact be skipped on all
    platforms we currently test. But it's a nice test to keep around for
    when a platform that uses IMAGE_BOOTLOADER_FILE comes around.

commit 741df67e274afc74dbd20a4ae1922cc00d7ae267
Author: Kristian Amlie <kristian.amlie@mender.io>
Date:   Wed Oct 19 12:58:26 2016

    Add autodetection of correct image to use for tests.

    It simply uses the newest image of each type: ext4, sdimg and mender.

```

### DIFF
--- a/tests/acceptance/common.py
+++ b/tests/acceptance/common.py
@@ -248,11 +248,56 @@ def qemu_running(request):
 
 @pytest.fixture(scope="function")
 def no_image_file(qemu_running):
+    """Make sure 'image.dat' is not present on the device."""
     execute(no_image_file_impl, hosts=conftest.current_hosts())
 
 
 def no_image_file_impl():
     run("rm -f image.dat")
+
+def latest_build_artifact(extension):
+    if os.environ.get('BUILDDIR', False):
+        raise Exception("BUILDDIR needs to be defined")
+
+    output = subprocess.check_output(["sh", "-c", "ls -t $BUILDDIR/tmp*/deploy/images/*/*%s | head -n 1" % extension])
+    output = output.rstrip('\r\n')
+    print("Found latest image of type '%s' to be: %s" % (extension, output))
+    return output
+
+@pytest.fixture(scope="session")
+def latest_rootfs():
+    # Find latest built rootfs.
+    return latest_build_artifact(".ext[234]")
+
+@pytest.fixture(scope="session")
+def latest_sdimg():
+    # Find latest built rootfs.
+    return latest_build_artifact(".sdimg")
+
+@pytest.fixture(scope="session")
+def latest_mender_image():
+    # Find latest built rootfs.
+    return latest_build_artifact(".mender")
+
+@pytest.fixture(scope="function")
+def image_dat(request, latest_rootfs):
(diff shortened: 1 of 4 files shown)


tests/acceptance/common.py
```diff
@@ -248,9 +248,53 @@ def qemu_finalizer_impl():
...
 def no_image_file(qemu_running):
+    """Make sure 'image.dat' is not present on the device."""
     execute(no_image_file_impl, hosts=conftest.current_hosts())
 
 
 def no_image_file_impl():
     run("rm -f image.dat")
 
+def latest_build_artifact(extension):
+    if len(os.environ['BUILDDIR']) <= 0:
```
GregorioDiStefano: a more pythonic way of doing this would be:

`os.environ.get("BUILDDIR", False)`

tests/acceptance/common.py
```diff
@@ -248,9 +248,53 @@ def qemu_finalizer_impl():
...
+        raise Exception("BUILDDIR needs to be defined")
+
+    output = subprocess.check_output(["sh", "-c", "ls -t $BUILDDIR/tmp*/deploy/images/*/*%s | head -n 1" % extension])
```
GregorioDiStefano: `output = subprocess.check_output(["sh", "-c", "ls -t $BUILDDIR/tmp*/deploy/images/*/*%s | head -n 1" % extension]).strip()`

tests/acceptance/test_sdimg.py
```diff
@@ -0,0 +1,88 @@
...
+    loader = None
+    offset = 0
+
+    def __init__(self, loader, offset):
+        self.loader = loader
+        self.offset = offset
+
+@pytest.fixture(scope="session")
+def embedded_bootloader():
+    assert(len(os.environ['BUILDDIR']) > 0, "BUILDDIR must be set")
```
GregorioDiStefano: same here

tests/acceptance/common.py
kacf: Will fix!

tests/acceptance/test_sdimg.py
kacf: Will fix!

GregorioDiStefano: actually, this will throw a key error, if builddir doesn't exist

tests/acceptance/common.py
kacf: I think I prefer to keep these separate. The stripping is easily forgotten, so I think it should stand out.

kacf:
Alright, issues fixed. Will merge this soon, but yoctobuild-master will also require some minor changes because of this (mainly avoid using image.dat for vexpress-qemu), which I will take care of a bit later.

PR #172, opened 2016-11-02
Updated acceptance tests to use mender artifacts instead of ext4 images

GregorioDiStefano:
@kacf please take a look

### DIFF
--- a/tests/acceptance/test_update.py
+++ b/tests/acceptance/test_update.py
@@ -21,12 +21,16 @@ import subprocess
 # Make sure common is imported after fabric, because we override some functions.
 from common import *
 
+if pytest.config.getoption("--bbb"):
+    image_type = "beaglebone"
+else:
+    image_type = "vexpress-qemu"
 
 class Helpers:
     @staticmethod
     def upload_to_s3():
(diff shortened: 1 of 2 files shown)


GregorioDiStefano:
Please don't merge yet

tests/acceptance/test_update.py
```diff
@@ -67,33 +73,36 @@ def test_too_big_image_update(self):
...
 
-    def test_network_based_image_update(self, image_dat):
+    def test_network_based_image_update(self):
         http_server_location = pytest.config.getoption("--http-server")
         bbb = pytest.config.getoption("--bbb")
 
         if not env.host_string:
             # This means we are not inside execute(). Recurse into it!
-            execute(self.test_network_based_image_update, image_dat)
+            execute(self.test_network_based_image_update)
```
kacf: This will probably break local testing unless you create `image.dat` manually, but that's alright, I can fix that later (I have some related work to do in that area anyway).

kacf: Hmm, actually, it will also break master... Maybe I'll put a small fix on top of yours to autodetect `image.dat` (for vexpress, that is).

kacf:
Superseded by #173.

PR #174, opened 2016-11-04
Fix rootfs always overshooting the alignment by 1MiB.

kacf:
Passing count=1 to dd would write one additional sector after the
sparse sectors created by using seek. Avoid that by setting to zero.

Also add acceptance test for verifying that image size, partition
sizes and partition alignments are correct in the image.

### DIFF
--- a/classes/mender-sdimg.bbclass
+++ b/classes/mender-sdimg.bbclass
@@ -149,7 +149,7 @@ IMAGE_CMD_sdimg() {
     ROOTFS_SIZE=$(expr $REMAINING_SIZE / 2)
 
     # create rootfs
-    dd if=/dev/zero of=${WORKDIR}/rootfs.$FSTYPE count=1 seek=$ROOTFS_SIZE bs=1M
+    dd if=/dev/zero of=${WORKDIR}/rootfs.$FSTYPE count=0 seek=$ROOTFS_SIZE bs=1M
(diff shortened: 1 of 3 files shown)


classes/mender-sdimg.bbclass
```diff
@@ -149,7 +149,7 @@ IMAGE_CMD_sdimg() {
     ROOTFS_SIZE=$(expr $REMAINING_SIZE / 2)
 
     # create rootfs
-    dd if=/dev/zero of=${WORKDIR}/rootfs.$FSTYPE count=1 seek=$ROOTFS_SIZE bs=1M
+    dd if=/dev/zero of=${WORKDIR}/rootfs.$FSTYPE count=0 seek=$ROOTFS_SIZE bs=1M
```
kacf: @bboozzoo

bboozzoo: IIRC mkfs.ext* would fail to create a FS if the whole image is sparse file. Have you verified that the rootfs is mountable?

kacf: Yes, the other py.test run the tests using qemu, which mounts the image being made here. I'll double check manually just to be sure...

kacf:
https://ci.cfengine.com/job/yoctobuild-pr/95/

kacf:
https://ci.cfengine.com/job/yoctobuild-pr/98/

PR #176, opened 2016-11-08
Group u-boot patch files into boards.

kacf:
Much easier to see which patches belong where.


kacf:
@bboozzoo

bboozzoo:
I'm leaning towards NACK. I suppose now could be the right time to introduce BSP/device layers. Uboot would fit there perfecly.

kacf:
Would this be a separate layer, `meta-mender-bsp`?

kacf:
I think you're right. It's more work, but it's better long term.

PR #181, opened 2016-11-10
add recipe for mender-artifact

bboozzoo:
The series adds a recipe for mender-artifact package. The package delivers `${sysconfdir}/mender/artifact-info`. The file is populated in rootfs preprocessing.

In order to preserve backwards compatibility with, `${sysconfdir}/mender/build_mender` is a symlink to `artifact-info`.

Setting mender dependencies in `mender-uboot` is not a good idea and bitbake correctly raises a warning when parsing recipes. `RDEPENDS` of mender were moved to mender recipe. `u-boot` was dropped from `[R]DEPENDS`, as it should get built & installed as a result of machine setup. 

`DEPENDS_mender` was removed. AFAIK there's no build dependency on `u-boot` nor on `u-boot-fw-utils`.

@pasinskim @kacf

### DIFF
--- a/classes/mender-uboot.bbclass
+++ b/classes/mender-uboot.bbclass
@@ -7,3 +7,4 @@ IMAGE_BOOT_ENV_FILE ?= "uboot.env"
 IMAGE_BOOT_FILES_append = " ${IMAGE_BOOT_ENV_FILE}"
 
 EXTRA_IMAGEDEPENDS += "u-boot"
+PACKAGECONFIG_append_pn-mender = " u-boot"
--- a/classes/mender-uboot.bbclass
+++ b/classes/mender-uboot.bbclass
@@ -7,6 +7,3 @@ IMAGE_BOOT_ENV_FILE ?= "uboot.env"
 IMAGE_BOOT_FILES_append = " ${IMAGE_BOOT_ENV_FILE}"
 
 EXTRA_IMAGEDEPENDS += "u-boot"
-
-DEPENDS_mender_append = " u-boot u-boot-fw-utils"
-RDEPENDS_mender_append = " u-boot u-boot-fw-utils"
--- a/recipes-mender/mender-artifact/mender-artifact.bb
+++ b/recipes-mender/mender-artifact/mender-artifact.bb
@@ -1,0 +1,22 @@
+DESCRIPTION = "Mender artifact information"
+HOMEPAGE = "https://mender.io"
+LICENSE = "Apache-2.0"
+
+inherit allarch
+
+PV = "0.1"
+
+do_compile() {
+    echo "# populate this file with build info" > ${B}/artifact-info
+}
+
+do_install() {
+    install -d ${D}${sysconfdir}/mender
+    install -t ${D}${sysconfdir}/mender ${B}/artifact-info
+    ln -s artifact-info ${D}${sysconfdir}/mender/build_mender
+}
+
+FILES_${PN} += " \
+    ${sysconfdir}/mender/artifact-info \
+    ${sysconfdir}/mender/build_mender \
+"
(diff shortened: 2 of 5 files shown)


recipes-mender/mender-artifact/mender-artifact.bb
```diff
@@ -0,0 +1,22 @@
...
+PV = "0.1"
+
+do_compile() {
+    echo "# populate this file with build info" > ${B}/artifact-info
+}
+
+do_install() {
+    install -d ${D}${sysconfdir}/mender
+    install -t ${D}${sysconfdir}/mender ${B}/artifact-info
+    ln -s artifact-info ${D}${sysconfdir}/mender/build_mender
```
kacf: I'm thinking we should get rid of this symlink while we can. I don't think anyone external is relying on it. Makes sense?

classes/mender-uboot.bbclass
```diff
@@ -7,6 +7,3 @@ IMAGE_BOOT_ENV_FILE ?= "uboot.env"
 IMAGE_BOOT_FILES_append = " ${IMAGE_BOOT_ENV_FILE}"
 
 EXTRA_IMAGEDEPENDS += "u-boot"
-
-DEPENDS_mender_append = " u-boot u-boot-fw-utils"
-RDEPENDS_mender_append = " u-boot u-boot-fw-utils"
```
kacf: NACK. There needs to be a way to turn off the U-Boot dependency for Mender. I don't think it's working fully as it should with the `mender-uboot` class; it will need more fixing and perhaps switching to [MEN-595](https://tracker.mender.io/browse/MEN-595); but putting it here would bring us even further from that goal.

In the future, we will allow Mender to update boot loader properties by calling user supplied commands, and in this case there is no dependency on U-Boot.

What is the warning that bitbake raises regarding this?

bboozzoo: This chunk is actually removing hard coded dependency of mender on uboot, so what's the issue here?

Besides, the only real dependency we have now is on uboot-fw-utils and that is added to RDEPENDS.

kacf: I think they should come in a pair, depending on `u-boot-fw-utils` without `u-boot` makes no sense. But that's not so important for this discussion anyway.

The point is that you add this dependency to `mender.bb`, so it's not possible to turn it off after this change. Right now you can turn it off by not using the `mender-uboot` class.

recipes-mender/mender-artifact/mender-artifact.bb
kacf: Well, that was my intention, to change the client also. :-)

classes/mender-uboot.bbclass
kacf: In its out-of-the-box shape, yes. But there are several reasons for it:

1. Later we will provide ways for people to call user defined scripts as commands for switching partition, updating boot counter, etc.
2. People are already testing Mender without U-Boot. Definitely an experimental thing to do at this point, but should still be possible.

recipes-mender/mender-artifact/mender-artifact.bb
pasinskim: I like the idea of leaving this as a symlink for now and then remove it once client will be implemented.

kacf: Alright, I guess I'm voted down. :-)

classes/mender-uboot.bbclass
kacf: Alright, fixed with latest commit.

kacf:
Testing in Jenkins.

kacf:
@bboozzoo: Could you rebase your PR? The tests are failing because it expects Marcin's latest changes to be there.

PR #189, opened 2016-11-25
MEN-749: Add acceptance test for redundant U-Boot environment.

kacf:
Because it bricks the device if it fails, it's currently disabled, but
should be enabled as part of MEN-480.

The last commit is the actual test, the rest is just a few cleanups.


kacf:
Superseded by #196.

PR #190, opened 2016-11-28
Add support for Raspberry Pi boards

mirzak:
This pull request adds support to build mender for Raspberry Pi boards. I have only tested this on a Raspberry Pi 2.

This adds a dependency on (when building for a Raspberry Pi board):
```
URI: git://git.yoctoproject.org/meta-raspberrypi
branch: master
revision: HEAD
```
Following needs to be added to local.conf when building for Raspberry Pi`s:
```
MACHINE ?= "raspberrypi2"

KERNEL_IMAGETYPE = "uImage"

INHERIT += "mender-full"

DISTRO_FEATURES_append = " systemd"
VIRTUAL-RUNTIME_init_manager = "systemd"
DISTRO_FEATURES_BACKFILL_CONSIDERED = "sysvinit"
VIRTUAL-RUNTIME_initscripts = ""
IMAGE_FSTYPES = "ext4"

MENDER_ARTIFACT_NAME = "my-mender-image-1.0"
MENDER_PARTITION_ALIGNMENT_MB = "4"
MENDER_BOOT_PART_SIZE_MB = "40"

MACHINE_ESSENTIAL_EXTRA_RDEPENDS_append = " bcm2835-bootfiles"

# raspberrypi files aligned with mender layout requirements
IMAGE_BOOT_FILES_append = " boot.scr u-boot.bin;${SDIMG_KERNELIMAGE}"
IMAGE_INSTALL_append = " kernel-image kernel-devicetree"
```

### DIFF
--- a/meta-mender-core/classes/mender-sdimg.bbclass
+++ b/meta-mender-core/classes/mender-sdimg.bbclass
@@ -97,7 +97,7 @@ python() {
             d.setVar('IMAGE_TYPEDEP_sdimg_append', " " + fs)
 }
 
-IMAGE_DEPENDS_sdimg = "${IMAGE_DEPENDS_wic} dosfstools-native mtools-native"
+IMAGE_DEPENDS_sdimg += "${IMAGE_DEPENDS_wic} dosfstools-native mtools-native"
 
 IMAGE_CMD_sdimg() {
     set -e
--- a/meta-mender-raspberrypi/README.md
+++ b/meta-mender-raspberrypi/README.md
@@ -1,0 +1,40 @@
+# meta-mender-raspberrypi
+
+This Yocto layers contains recipes which enables support of building Mender client for Raspberry Pi boards.
+
+**NOTE!**. To be able to support update of Linux kernel and DTB these are installed to `/boot` on rootfs. This breaks support of user configurations based on `config.txt` because the Raspberry Pi boot firmware requires that the DTB file is in the same partition as the boot firmware files (mmcblk0p1). Boot firmware normally patches the DTB file based on configurations in `config.txt`.
+
+Above should not pose any problems if you do not require any changes in `config.txt` and the default configuration certainly is enough to run Mender client.
+
+## Dependencies
+
+This layer depends on:
+
+    URI: git://git.yoctoproject.org/meta-raspberrypi
+    branch: master
+    revision: HEAD
+
+in addition to `meta-mender` dependencies.
+
+## Build instructions
+
+- Read [the Mender documentation on Building a Mender Yocto image](https://docs.mender.io/Artifacts/Building-Mender-Yocto-image) for Mender specific configuration.
+- Set MACHINE to one of the following
+    - raspberrypi
+    - raspberrypi0
+    - raspberrypi2
+    - raspberrypi3
+- Add following to your local.conf (including configuration required by meta-mender-core)
+
+        KERNEL_IMAGETYPE = "uImage"
+
+        MENDER_PARTITION_ALIGNMENT_MB = "4"
+        MENDER_BOOT_PART_SIZE_MB = "40"
+
+        IMAGE_DEPENDS_sdimg += " bcm2835-bootfiles"
+
+        # raspberrypi files aligned with mender layout requirements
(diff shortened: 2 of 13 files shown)


kacf:
Awesome! I'll take a look shortly!

meta-mender-core/classes/mender-sdimg.bbclass
```diff
@@ -97,7 +101,7 @@ python() {
             d.setVar('IMAGE_TYPEDEP_sdimg_append', " " + fs)
 }
 
-IMAGE_DEPENDS_sdimg = "${IMAGE_DEPENDS_wic} dosfstools-native mtools-native"
+IMAGE_DEPENDS_sdimg = "${MENDER_IMAGE_EXTRA_DEPENDS} ${IMAGE_DEPENDS_wic} dosfstools-native mtools-native"
```
kacf: I believe you can get the same effect without adding this, by adding this to `local.conf`:

```
MACHINE_ESSENTIAL_EXTRA_RDEPENDS_append = " bcm2835-bootfiles"
```

mirzak: Makes sense. Will drop that patch.

Thank you for you review.

mirzak:
Was a bit fast accepting that change :).

Build fails when I use

```
MACHINE_ESSENTIAL_EXTRA_RDEPENDS_append = " bcm2835-bootfiles"
```

Will need to look in to this a bit

bboozzoo:
I would start by adding this to your `local.conf`:
```
EXTRA_IMAGEDEPENDS += " bcm2835-bootfiles"
```

I suppose this should actually be set by machine conf. From what I can see, `meta-raspberrypi` does not set it, so may want to submit a patch upstream at some point.

kacf:
> `EXTRA_IMAGEDEPENDS += "bcm2835-bootfiles"`

At least for me this one didn't work, because it ended up building it too late, after the sdimg stage.

mirzak:
Yes, got the same problem with EXTRA_IMAGEDEPENDS.

Following works and is the equivalent of my initial patch

```
IMAGE_DEPENDS_sdimg += " bcm2835-bootfiles"
```

Acceptable?

kacf:
I'm good with that! Maybe this, and the other `local.conf` modifications you have, are good candidates for a `README.md` file in the `meta-mender-raspberry` folder, what do you think?

mirzak:
> I'm good with that! Maybe this, and the other local.conf modifications you have, are good candidates for a README.md file in the meta-mender-raspberry folder, what do you think?

+1 on that. I will create one and push to this pull request.

mirzak:
> Yes, got the same problem with EXTRA_IMAGEDEPENDS.

> Following works and is the equivalent of my initial patch

> IMAGE_DEPENDS_sdimg += " bcm2835-bootfiles"
> Acceptable?

Did a full re-build. Above does not work, I was fooled by "old" build state.

To start from the beginning. Ultimately I would like to add a dependency to "sdimg". Which my initial patch did and I think that this is the only "correct" way (not maybe how I did it, but I need a way to add this dependency). At the moment I can not see how I could do that.

```
 IMAGE_DEPENDS_sdimg += " bcm2835-bootfiles"
```
is overwritten by
```
IMAGE_DEPENDS_sdimg = "${IMAGE_DEPENDS_wic} dosfstools-native mtools-native"
```
in the mender-sdimg.bbclass.

What bcm2835 provides are some binary files which are in IMAGE_BOOT_FILES, and hence the only dependency is "sdimg/wic". To add this dependency anywhere else seems wrong and produces unwanted behavior.

Something that would fix my problem here is changing:

```
IMAGE_DEPENDS_sdimg = "${IMAGE_DEPENDS_wic} dosfstools-native mtools-native"
```
to

```
IMAGE_DEPENDS_sdimg += "${IMAGE_DEPENDS_wic} dosfstools-native mtools-native"
```

Then it wont overwrite my change in local.conf.

Thoughts? Tips?

kacf:
I think that's alright, changing `=` to `+=`.

mirzak:
Any comments on the updates?

kacf:
And in it goes! Thanks a lot!

PR #195, opened 2016-12-02
Fix build for U-Boot 2016.11 and later.

kacf:
Based on following user feedback:

Tested meta-mender on U-boot 2016.11 and get following:
| Error: You must add new CONFIG options using Kconfig
| The following new ad-hoc CONFIG options were detected:
| CONFIG_MENDER_H
|
| Please add these via Kconfig instead. Find a suitable Kconfig
| file and add a 'config' or 'menuconfig' option.

This is because of [PATCH 1/2] Generic boot code for Mender.

It is probably not the best naming convention to begin a header file
name with CONFIG_ since this is obviously now interpreted as
configuration option and hence the error.


mirzak:
It does fix it.

This also does https://github.com/mirzak/meta-mender/commit/a3c69958e88ee8f24ac223e73d7b289258a6f216. :)

kacf:
Green in Jenkins, so I'll go with this one. :-)

PR #196, opened 2016-12-08
MEN-480, MEN-749: Redundant U-Boot environment

kacf:
This is best reviewed commit by commit.

### DIFF
--- a/meta-mender-core/classes/mender-install.bbclass
+++ b/meta-mender-core/classes/mender-install.bbclass
@@ -50,7 +50,7 @@ MENDER_DATA_PART_DIR ?= ""
 MENDER_DATA_PART_SIZE_MB ?= "128"
 
 # Size of the first (FAT) partition, that contains the bootloader
-MENDER_BOOT_PART_SIZE_MB ?= "32"
+MENDER_BOOT_PART_SIZE_MB ?= "16"
 
 # For performance reasons, we try to align the partitions to the SD
 # card's erase block. It is impossible to know this information with
--- a/meta-mender-core/classes/mender-install.bbclass
+++ b/meta-mender-core/classes/mender-install.bbclass
@@ -35,6 +35,33 @@ MENDER_DEVICE_TYPE ?= "${MACHINE}"
 # Space separated list of device types compatible with the built update.
 MENDER_DEVICE_TYPES_COMPATIBLE ?= "${MACHINE}"
 
+# Total size of the medium that mender sdimg will be written to. The size of
+# rootfs partition will be calculated automatically by subtracting the size of
+# boot and data partitions along with some predefined overhead (see
+# MENDER_PARTITIONING_OVERHEAD_MB).
+MENDER_STORAGE_TOTAL_SIZE_MB ?= "1024"
+
+# Optional location where a directory can be specified with content that should
+# be included on the data partition. Some of Mender's own files will be added to
+# this (e.g. OpenSSL certificates).
+MENDER_DATA_PART_DIR ?= ""
+
+# Size of the data partition, which is preserved across updates.
+MENDER_DATA_PART_SIZE_MB ?= "128"
+
+# Size of the first (FAT) partition, that contains the bootloader
+MENDER_BOOT_PART_SIZE_MB ?= "32"
+
+# For performance reasons, we try to align the partitions to the SD
+# card's erase block. It is impossible to know this information with
+# certainty, but one way to find out is to run the "flashbench" tool on
+# your SD card and study the results. If you do, feel free to override
+# this default.
+#
+# 8MB alignment is a safe setting that might waste some space if the
+# erase block is smaller.
+MENDER_PARTITION_ALIGNMENT_MB ?= "8"
+
 # --------------------------- END OF CONFIGURATION -----------------------------
 
 
--- a/meta-mender-core/classes/mender-sdimg.bbclass
+++ b/meta-mender-core/classes/mender-sdimg.bbclass
@@ -100,7 +100,7 @@ python() {
 IMAGE_DEPENDS_sdimg += "${IMAGE_DEPENDS_wic} dosfstools-native mtools-native"
 
 IMAGE_CMD_sdimg() {
-    set -e
+    set -ex
 
     # For some reason, logging is not working correctly inside IMAGE_CMD bodies,
     # so wrap all logging in these functions that also have an echo. This won't
--- a/meta-mender-core/classes/mender-sdimg.bbclass
+++ b/meta-mender-core/classes/mender-sdimg.bbclass
@@ -158,10 +158,10 @@ IMAGE_CMD_sdimg() {
     MENDER_PARTITION_ALIGNMENT_KB=$(expr ${MENDER_PARTITION_ALIGNMENT_MB} \* 1024)
 
     rm -rf "${WORKDIR}/data" || true
+    mkdir -p "${WORKDIR}/data"
+
     if [ -n "${MENDER_DATA_PART_DIR}" ]; then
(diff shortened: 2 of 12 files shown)


kacf:
@pasinskim, @bboozzoo

meta-mender-core/classes/mender-install.bbclass
```diff
@@ -35,6 +35,33 @@ MENDER_DEVICE_TYPE ?= "${MACHINE}"
 # Space separated list of device types compatible with the built update.
 MENDER_DEVICE_TYPES_COMPATIBLE ?= "${MACHINE}"
 
+# Total size of the medium that mender sdimg will be written to. The size of
```
pasinskim: `The total size of the medium where the Mender sdimg will be written to. The size of the rootfs` I am surprised myself I am noticing this stuff now. It's because of @eysteins complaining about not using articles :)

meta-mender-core/classes/mender-sdimg.bbclass
```diff
@@ -175,12 +152,27 @@ IMAGE_CMD_sdimg() {
     rm -f "$wks"
     if [ -n "${IMAGE_BOOTLOADER_FILE}" ]; then
         bootloader_align_kb=$(expr $(expr ${IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET} \* 512) / 1024)
+        bootloader_size=$(stat -c '%s' "${DEPLOY_DIR_IMAGE}/${IMAGE_BOOTLOADER_FILE}")
+        bootloader_end=$(expr $bootloader_align_kb \* 1024 + $bootloader_size)
+        if [ $bootloader_end -gt ${MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET} ]; then
+            bberror "Size of boot loader specified in IMAGE_BOOTLOADER_FILE" \
```
pasinskim: Whereas this is not standardized and this is not error, I would still prefer `bootloader` over `boot loader`.

meta-mender-core/recipes-bsp/u-boot/patches/0001-Generic-boot-code-for-Mender.patch
```diff
@@ -40,55 +40,48 @@ index 0000000..3c1ecd6
...
 +
-+#define FAT_ENV_INTERFACE       MENDER_UBOOT_STORAGE_INTERFACE
-+#define FAT_ENV_DEVICE_AND_PART __stringify(MENDER_UBOOT_STORAGE_DEVICE) ":" __stringify(MENDER_BOOT_PART_NUMBER)
-+#define FAT_ENV_DEVICE          MENDER_UBOOT_STORAGE_DEVICE
-+#define FAT_ENV_PART            MENDER_BOOT_PART_NUMBER
-+#define FAT_ENV_FILE            "uboot.env"
-+
-+#define CONFIG_FAT_WRITE
++#define CONFIG_SYS_MMC_ENV_DEV   MENDER_UBOOT_STORAGE_DEVICE
++/* PART in this context is not regular partition, but mmc partition, a
```
pasinskim: Maybe change to `CONFIG_SYS_MMC_ENV_PART is in this ...`. It was hard to catch up what do you mean by part. Also, maybe just saying this is different thing you can add some comment what this is.

meta-mender-core/recipes-bsp/u-boot/u-boot-fw-utils-mender.inc
```diff
@@ -1,14 +1,36 @@
...
+    install -d -m 755 ${D}${sysconfdir}
+    ln -sf /data/u-boot/fw_env.config ${D}${sysconfdir}/fw_env.config
+}
+
+do_deploy() {
+    install -d ${WORKDIR}/data/u-boot
+    install -m 0644 ${WORKDIR}/fw_env.config ${WORKDIR}/data/u-boot/fw_env.config
+    # We need to keep the permissions in a tar file, because they are not
+    # remembered by bitbake's fakeroot mechanism if we put it directly in the
+    # deploy directory.
```
pasinskim: That is a nice workaround!

tests/acceptance/test_update.py
```diff
@@ -139,24 +184,133 @@ def test_network_based_image_update(self, successful_image_update_mender):
...
+        subprocess.call("dd if=/dev/zero of=image.dat bs=1M count=0 seek=8", shell=True)
+        subprocess.call("artifacts write rootfs-image -t %s -n test-update -u image.dat -o image.mender" % image_type, shell=True)
+        put("image.mender", remote_path="/var/tmp/image.mender")
+        run("mender -rootfs /var/tmp/image.mender")
+
+        new_checksums = Helpers.get_env_checksums(bitbake_variables)
+
+        # Exactly one checksum should be different.
+        assert(old_checksums[0] == new_checksums[0] or old_checksums[1] == new_checksums[1])
+        assert(old_checksums[0] != new_checksums[0] or old_checksums[1] != new_checksums[1])
```
pasinskim: This can be possibly removed as we are checking this later on anyway and we are doing `assert False, "Should not happen"` if this is the case.

tests/acceptance/test_update.py
```diff
@@ -139,24 +184,133 @@ def test_network_based_image_update(self, successful_image_update_mender):
...
+        reboot()
+
+        # We should have recovered.
+        run_after_connect("true")
+
+        # And we should be back at the second rootfs partition.
+        (active, passive) = determine_active_passive_part(bitbake_variables)
+        assert(active == bitbake_variables["MENDER_ROOTFS_PART_B"])
+
+        # Cleanup.
```
pasinskim: There is a lot of assertions here, so possibly we will exit without cleaning this. If we don't care and everything will be wiped out after the test this is OK. If this is not the case, maybe we should close this inside `try` and then do `except AssertionError:` here.

meta-mender-core/recipes-bsp/u-boot/patches/0001-Generic-boot-code-for-Mender.patch
kacf: Lol, I guess you're right! Even when I read it now I was confusing myself. I'll make it more clear!

meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
```diff
@@ -47,15 +46,39 @@ do_provide_mender_defines() {
...
+
+MENDER_BOOTENV_TOTAL_ALIGNED_SIZE = "${@mender_get_env_total_aligned_size(${BOOTENV_SIZE}, ${MENDER_PARTITION_ALIGNMENT_MB})}"
+
+def mender_get_env_offset(start_offset, index, total_aligned_size):
+    if index == 1:
+        return "0x%x" % int(start_offset)
+    elif index == 2:
+        return "0x%x" % int(start_offset + total_aligned_size / 2)
+    else:
+        raise Exception("Should not happen")
```
kacf: You're right, that's better, will fix.

tests/acceptance/test_update.py
kacf: Hmm, not really, the if statement below won't catch the case if both are different (it will enter the first branch in this case). Perhaps it's better if I remove the `else` below, because that one should never be hit after these asserts.

tests/acceptance/test_update.py
kacf: Yes, that will be better, I'll change that.

kacf:
Thanks for the review. Indeed this is quite complicated stuff, and we're trying to automate something which isn't easy to automate (automatic partitioning and U-Boot environment).

tests/acceptance/test_update.py
pasinskim: Right, my bad.

kacf:
Updated!

PR #197, opened 2016-12-08
Update U-Boot patches after move to U-Boot v2016.11.

kacf:


kacf:
@GregorioDiStefano: I merged this since it does make it better. However, the BBB is still broken, so we need to keep working on it.

PR #201, opened 2016-12-13
raspberrypi: u-boot: move U-boot environment to MMC

mirzak:
Since https://github.com/mendersoftware/meta-
mender/commit/2a8ea346aa0d12da755d36f426164a0e54f757b3, it is expected
that the U-boot environment is on the MMC instead of a file on VFAT
partition.


kacf:
Cool, thanks!

PR #202, opened 2016-12-15
MEN-874 Add retry poll interval configuration option.

pasinskim:


pasinskim:
Rebased and fixed missing sed. Jenkins is green now.

PR #204, opened 2016-12-15
Enable Mender daemon by default.

kacf:


kacf:
@GregorioDiStefano: This passes the client tests. Will it have any implications for the integration tests, or can I merge?

PR #208, opened 2016-12-16
Rename artifacts to mender-artifact.

pasinskim:


pasinskim:
@kacf thanks for spotting this. Should be fixed now, but unfortunately diff is a bit messy after moving all those files.

meta-mender-core/classes/mender-artifactimg.bbclass
```diff
@@ -22,7 +22,7 @@ IMAGE_CMD_mender () {
         bberror "MENDER_DEVICE_TYPES_COMPATIBLE variable cannot be empty."
     fi
 
-    artifacts write rootfs-image \
+    menedr-artifact write rootfs-image \
```
kacf: `s/menedr/mender/`

PR #218, opened 2016-12-28
sdimg: Start using wic's --fixed-size option.

kacf:
This is much more predictable than --size, and will produce an error
if the size can't be honored.


kacf:
@bboozzoo: Finally we can use it! :-)

kacf:
@bboozzoo: Nice, much better than the old option.

PR #220, opened 2016-12-30
Default to empty images in docker build.

kacf:
This is primarily for development, where instead of creating a new
docker image every time you've built a new Yocto image, you grab them
from /mnt/build mount point instead. The image won't work unless you
specify either the arguments or the mount, but this was the case
before this patch as well.


kacf:
@GregorioDiStefano

PR #224, opened 2017-01-11
meta-mender-selftest: oe-selftest support

bboozzoo:
See README for details. Basically you can implement the same things as are implemented for OE core right here: https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/lib/oeqa/selftest

Perhaps instead of using a separate selftest layer is a bit of an overkill. The tests could be part of meta-mender-core after all. We can revise it later though, if we decide this PR is something that we want do or not.

### DIFF
--- a/meta-mender-selftest/README.md
+++ b/meta-mender-selftest/README.md
@@ -1,0 +1,15 @@
+## meta-mender-selftest
+
+OE selftest layer for mender Yocto support. Before using it, you need to add it
+to `BBLAYERS` like this:
+
+```
+BBLAYERS ?= " \
+  ...
+  <your-path>/meta-mender/meta-mender-selftest \
+  ...
+```
+
+Tests can be listed with `oe-selftest --list-tests-by module mender`. Individual
+tests are run by `oe-selftest -r mender.Mender.test_hello`. See `oe-selftest
+--help` for details.
--- a/meta-mender-selftest/conf/layer.conf
+++ b/meta-mender-selftest/conf/layer.conf
@@ -1,0 +1,13 @@
+# Layer configuration for meta-mender
+# Copyright 2016 Mender Software AS
+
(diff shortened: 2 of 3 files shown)


kacf:
It would be nice to use the OE official way, but I think we should avoid the situation of having two test implementations. If we did this at all I think we would need to convert our existing test cases.

Do you know if there is any particular advantage to oe-selftest over pytest and fabric, other than aligning the workflow with OE?

bboozzoo:
I guess it depends on a test case. For instance, if you want to verify that recipes are correct, image gets built, it has a particular layout, certain files are at specific locations, all this is doable in selftest. Also, there's no special 3rd party libraries needed, everything is already there. You could also run basic qemu based tests, wic tests do that already.

Running backend services not so much, this I'd leave to integration tests.

kacf:
Yes, what I worry about the most is the interaction with the clients, we are using the SSH connection extensively both for QEMU and Beaglebone. I'm sure it's all possible in oe-selftest, but I expect some work will be involved to achieve it.

bboozzoo:
Some work yes, but maybe not that much, for instance this is how wic tests do it:

```
    def test_qemu(self):
        """Test wic-image-minimal under qemu"""
        self.assertEqual(0, bitbake('wic-image-minimal').status)

        with runqemu('wic-image-minimal', ssh=False) as qemu:
            cmd = "mount |grep '^/dev/' | cut -f1,3 -d ' '"
            status, output = qemu.run_serial(cmd)
            self.assertEqual(1, status, 'Failed to run command "%s": %s' % (cmd, output))
            self.assertEqual(output, '/dev/root /\r\n/dev/vda3 /mnt')
```

There are also qemu wrappers and SSH wrappers in oeqa.

I suppose the ultimate goal is to make sure that image gets built correctly and contains everything we expect it to have without running integration tests.

PR #226, opened 2017-01-12
MEN-934: Always respect storage conf when determining rootfs size.

kacf:
This is "payload" of the PR:

```
MEN-934: Always respect storage conf when determining rootfs size.

Previously it was only being respected when building the sdimg. We
want it to be respected also when building mender updates.

For this we auto-set the standard IMAGE_ROOTFS_SIZE Yocto variable,
however it's still possible to override it by the user.
```

The rest of the commits are simply enablers for this to work.

### DIFF
--- a/meta-mender-core/classes/mender-install.bbclass
+++ b/meta-mender-core/classes/mender-install.bbclass
@@ -77,3 +77,27 @@ IMAGE_INSTALL_append = " \
     ca-certificates \
     mender-artifact-info \
     "
+
+# Estimate how much space may be lost due to partitioning alignment. Use a
+# simple heuristic for now - 4 partitions * alignment
+def mender_get_part_overhead(d):
+    align = d.getVar('MENDER_PARTITION_ALIGNMENT_MB', True)
+    if align:
+        return 4 * int(align)
+    return 0
+
+# Overhead lost due to partitioning.
+MENDER_PARTITIONING_OVERHEAD_MB ?= "${@mender_get_part_overhead(d)}"
+
+
+def mender_calculate_rootfs_size_kb(total_mb, boot_mb, data_mb, overhead_mb, reserved_space_size):
+    return int(((total_mb - boot_mb - data_mb - overhead_mb) * 1048576 - reserved_space_size) / 2 / 1024)
+
+# Auto detect image size from other settings.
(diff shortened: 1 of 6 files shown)


meta-mender-core/classes/mender-install.bbclass
```diff
@@ -77,3 +77,27 @@ IMAGE_INSTALL_append = " \
...
+    return int(((total_mb - boot_mb - data_mb - overhead_mb) * 1048576 - uboot_env_size) / 2 / 1024)
+
+# Auto detect image size from other settings.
+MENDER_CALC_ROOTFS_SIZE = "${@mender_calculate_rootfs_size_kb(${MENDER_STORAGE_TOTAL_SIZE_MB}, \
+                                                              ${MENDER_BOOT_PART_SIZE_MB}, \
+                                                              ${MENDER_DATA_PART_SIZE_MB}, \
+                                                              ${MENDER_PARTITIONING_OVERHEAD_MB}, \
+                                                              ${MENDER_STORAGE_RESERVED_RAW_SPACE})}"
+# Gently apply this as the default image size.
+IMAGE_ROOTFS_SIZE ?= "${MENDER_CALC_ROOTFS_SIZE}"
```
bboozzoo: There's also IMAGE_ROOTFS_EXTRA_SPACE and IMAGE_ROOTFS_OVERHEAD_FACTOR. We should probably set the first one to 0 and overhead to 1.0

kacf: Yes, good point.

kacf: There doesn't seem to be a sane way to set either. Both are defined in poky using `?=`, but if ours come later, they won't be respected. We could use `=`, but then it's more difficult for the user to override our default (will be order dependent, most likely).

Not sure what is the best approach. We could leave it out for now, and submit a patch to poky, changing `?=` to `??=`, with the argument that it makes it easier for downstream layers to apply gentle defaults. What do you think?

kacf:
No clear path to set the other variables, so I'm leaving that for now. I added some extra information in the warning you get if the rootfs size is too big, referring to those variables, so the user will at least get a pointer where to look.

PR #227, opened 2017-01-14
demo: netbase: add option to specify address of Mender server

mirzak:
This will add an option so that the user can specify the Mender server
address during build time.

e.g. in local.conf:

	MENDER_DEMO_SERVER_ADDRESS = "192.168.0.100"

This is usually a manual step that the user has to perform each time a
new image is built.

### DIFF
--- a/meta-mender-demo/recipes-core/netbase/netbase_%.bbappend
+++ b/meta-mender-demo/recipes-core/netbase/netbase_%.bbappend
@@ -1,0 +1,7 @@
+MENDER_DEMO_HOST_IP_ADDRESS ?= ""
+
+do_install_append() {
+    if [ -n "${MENDER_DEMO_HOST_IP_ADDRESS}" ]; then
+        echo "${MENDER_DEMO_HOST_IP_ADDRESS} docker.mender.io s3.docker.mender.io" >> ${D}${sysconfdir}/hosts
+    fi
+}


kacf:
Can't you use the existing `MENDER_SERVER_URL` in the mender recipe?

I see now that it's not documented btw, but it should be.

mirzak:
But that only affects the mender.conf? 

What I am trying to "automate" is the following https://docs.mender.io/Getting-started/Deploy-to-physical-devices#insert-the-address-of-mender-server.

mirzak:
Ok, got it now. MENDER_SERVER_URL all ready does this. The docs confused me a bit.

mirzak:
Hmm, changed MENDER_SERVER_URL to point to a machine on the local network and then I get the following error:

```
Jan 06 00:05:20 raspberrypi2 mender[222]: level=error msg="authorize failed: transient error: authorization request failed: failed to execute authorization request: Post https://192.168.1.20:8080/api/devices/0.1/authentication/auth_requests: x509: cannot validate certificate for 192.168.1.20 because it doesn't contain any IP SANs" module=state
```

Maybe I should have sent this to the mailing list instead?

bboozzoo:
s/MENDER_SERVER_DEMO_ADDRESS/MENDER_SERVER_DEMO_IP_ADDRESS/ since only IP address is allowed there.

I think I would rather append to /etc/hosts iff MENDER_SERVER_DEMO_IP_ADDRESS is set, rather than overwrite the file completely.

As for MENDER_SERVER_URL, you're looking at a problem with certificate validation. The demo certificate is for `docker.mender.io` domain, and this is what CN is set to in the certificate. We do not set any SANs (IP SANs in particular). What happens in your case is that when a device is accessing mender backend using IP address, the backend presents a certificate for `docker.mender.io` that does not contain any IP SANs, hence it's not possible to verify the host (identified by IP address) the client is connecting to.

At this point you have 2 options, either generate new certificates and set IP SAN or use the current certificate and alias `docker.mender.io` to your backend IP address. Keeping that in mind, the workaround with adding entries to `/etc/hosts` is the easiest to implement and IMO fits demo purposes quite well (we actually do the same when using tools integration repo on the host running the backend stack)

mirzak:
> s/MENDER_SERVER_DEMO_ADDRESS/MENDER_SERVER_DEMO_IP_ADDRESS/ since only IP address is allowed there.

I actually had _ADDRESS initially, but for some reason I changed that thinking it would look better. Will changes this.

> I think I would rather append to /etc/hosts iff MENDER_SERVER_DEMO_IP_ADDRESS is set, rather than overwrite the file completely.

Good point.

> At this point you have 2 options, either generate new certificates and set IP SAN or use the current certificate and alias docker.mender.io to your backend IP address. Keeping that in mind, the workaround with adding entries to /etc/hosts is the easiest to implement and IMO fits demo purposes quite well (we actually do the same when using tools integration repo on the host running the backend stack)

Not really gonna generate new certificates for demo/test purpose. My thought was that meta-mender-demo should work for my current usage and purpose with the default certificate.

So this pull request makes sense for meta-mender-demo? With the updates that you suggested?

bboozzoo:
Yes, it makes sense to merge this into demo layer.

kacf:
Good points, the existing `MENDER_SERVER_URL` will be impractical. But I think this change will break the Docker setup, because Docker will resolve docker.mender.io to an actual IP, and if this is merged it will always point to `127.0.0.1`. Is there any way you could make this conditional? Maybe default to empty instead, and then append the entry to `/etc/hosts` only if it's non-empty.

mirzak:
> Good points, the existing MENDER_SERVER_URL will be impractical. But I think this change will break the Docker setup, because Docker will resolve docker.mender.io to an actual IP, and if this is merged it will always point to 127.0.0.1. Is there any way you could make this conditional? Maybe default to empty instead, and then append the entry to /etc/hosts only if it's non-empty.

Makes sense to make it optional, will do that.

mirzak:
Updated

kacf:
Great, thanks! Will run this through our CI.

kacf:
Thanks! I also added https://github.com/mendersoftware/mender-docs/pull/92.

PR #232, opened 2017-01-22
add UBIFS support to meta-mender

mirzak:
I will label this as an RFC , even though I have put a lot of work in to this all ready and as I see it there should be only minor tweaking left. Hope you agree :).

The idea with this pull-request is to add UBIFS support to meta-mender. This also adds support for a new board (Toradex Colibri V61) which is used as an reference board for UBIFS + Mender.

First 6 patches are only to prepare for the actually UBIFS support.

An new "installer" is added which is called "ubimg". This will create an UBI image with the following volume layout:
ubi0_0 - rootfsa
ubi0_1 - rootfsb
ubi0_2 - data

This UBI image can be used to provision a new device so that the "partitioning" is according to Mender requirements. Devices that all ready use UBIFS today specify an image type that is called "ubifs", this image type is part of poky. The "ubifs" image is used when creating the Mender artifact.

No work has been done on the Mender client and artifact tool yet. Wanted to get your feedback on this part first.

It is probably easier to review this "one patch a time".


mirzak:
Hmm, the different patches seem out of order in the pull-request. That is not order I have in the git tree. https://github.com/mirzak/meta-mender/commits/ubifs

meta-mender-core/classes/mender-install-ubi.bbclass
```diff
@@ -0,0 +1,25 @@
+ARTIFACTIMG_FSTYPE ?= "ubifs"
+
+MENDER_STORAGE_DEVICE ?= "ubi0"
```
kacf: Here we're running into a tricky issue, because this will be reused by the Mender client to open the device file, hence this has to be a `/dev` reference. Is the plain `ubi0` reference used anywhere outside of U-Boot? Won't the kernel always refer to a device file, like `/dev/mtd`?

The way I suggest is to stick to `/dev` references in this variable, and use the functions in `mender-helpers.bbclass` to obtain the references needed for U-Boot. But I'm still a bit confused about the different naming, so I might be way off the mark.

meta-mender-core/classes/mender-ubimg.bbclass
```diff
@@ -0,0 +1,107 @@
...
+
+    # Workaround for the fact that the image builder requires this directory,
+    # despite not using it. If "rm_work" is enabled, this directory won't always
+    # exist.
+    mkdir -p "${IMAGE_ROOTFS}"
+
+    REMAINING_SIZE=$(expr ${MENDER_STORAGE_TOTAL_SIZE_MB} - \
+         ${MENDER_DATA_PART_SIZE_MB} - 50)
+
+    CALC_ROOTFS_SIZE=$(expr $REMAINING_SIZE / 2)
```
kacf: This has recently been moved to be a global calculation in `mender-install.bbclass`. It probably needs some adjustments there.

meta-mender-core/classes/mender-sdimg.bbclass
```diff
@@ -114,8 +114,6 @@ IMAGE_CMD_sdimg() {
                           $boot_env_size_mb)
     CALC_ROOTFS_SIZE=$(expr $REMAINING_SIZE / 2)
 
-    MENDER_PARTITION_ALIGNMENT_KB=$(expr ${MENDER_PARTITION_ALIGNMENT_MB} \* 1024)
-
```
kacf: Careful! The reference below is a shell reference. You need to add the `{}` to it.

meta-mender-core/classes/mender-install.bbclass
```diff
@@ -61,7 +61,7 @@ MENDER_BOOT_PART_SIZE_MB ?= "16"
 #
 # 8MB alignment is a safe setting that might waste some space if the
 # erase block is smaller.
-MENDER_PARTITION_ALIGNMENT_MB ?= "8"
+MENDER_PARTITION_ALIGNMENT_KB ?= "8192"
```
kacf: This variable is already used by many, can you add a snippet ala:

```
python() {
    if d.getVar('MENDER_PARTITION_ALIGNMENT_MB', True) != "":
        bb.fatal("MENDER_PARTITION_ALIGNMENT_MB is deprecated. Please define MENDER_PARTITION_ALIGNMENT_KB instead.")
}
```

meta-mender-toradex-nxp/recipes-bsp/u-boot/u-boot-toradex-fsl-fw-utils_%.bbappend
```diff
@@ -0,0 +1 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
```
kacf: I would replace this with `require u-boot-mender-toradex-nxp.inc` so everything is in the same place.

meta-mender-core/recipes-core/base-files/base-files_3.%.bbappend
```diff
@@ -8,13 +8,18 @@ do_patch_device() {
...
         exit 1
     fi
 
-    sed -i -e 's,@MENDER_BOOT_PART@,${MENDER_BOOT_PART},g' ${S}/fstab
-    sed -i -e 's,@MENDER_BOOT_FSTYPE@,${MENDER_BOOT_FSTYPE},g' ${S}/fstab
+    if [ -n "${MENDER_BOOT_PART}" ]; then
+        sed -i -e 's,@MENDER_BOOT_PART@,${MENDER_BOOT_PART},g' ${S}/fstab
+        sed -i -e 's,@MENDER_BOOT_FSTYPE@,${MENDER_BOOT_FSTYPE},g' ${S}/fstab
+    else
+        bbwarn "MENDER_BOOT_PART not set. Removing from fstab..."
```
mirzak: Agree that `bbwarn` is a bit overkill.

Maybe drop it completely?  But might be good to have something in the logs, will update with `bbdebug`

meta-mender-core/classes/mender-install.bbclass
mirzak: Got it!

meta-mender-core/classes/mender-helpers.bbclass
```diff
@@ -19,6 +22,9 @@ get_uboot_device_from_device() {
         /dev/mmcblk[0-9])
             echo ${1#/dev/mmcblk}
             ;;
+        ubi[0-9])
+            echo ubi
```
mirzak: Glad you caught this, I wanted to drop this change. Because it is incorrect.

If we look at my kernel command line:
``` 
Kernel command line: ubi.mtd=ubi root=ubi0_0 rootfstype=ubifs ubi.fm_autoconvert=1
```

and in U-boot:
```
ubiargs=ubi.mtd=${mender_uboot_dev} root=${mender_kernel_root} rootfstype=ubifs ubi.fm_autoconvert=
```

As you can see I have used `mender_uboot_dev` and that is why I modified helper class. But what I need to assign to `ubi.mtd=` is the `name` or `number` of the MTD part where UBI is present. 

In my case it is `3` or `ubi` taken from below configuration
```
Colibri VFxx # mtdparts

device nand0 <vf610_nfc>, # parts = 4
 #: name                size            offset          mask_flags
 0: vf-bcb              0x00020000      0x00000000      1
 1: u-boot              0x00160000      0x00020000      1
 2: u-boot-env          0x00100000      0x00180000      0
 3: ubi                 0x1fd80000      0x00280000      0
```

It is probably bad to enforce the `ubi` name which I do with how it works currently in `mender-helper`.

All of this was building up to a question. Is it correct of me to re-use the `mender_uboot_dev` for this use-case, or should I create a new configuration option?

If it is OK to re-use the `mender_uboot_dev` I will probably drop above change and require it to be configured with `MENDER_UBOOT_STORAGE_DEVICE` only.

meta-mender-core/recipes-bsp/u-boot/patches/0001-Generic-boot-code-for-Mender.patch
```diff
@@ -156,6 +156,6 @@ index 0000000..2c43fd3
 +    "fi\0"
 +
 +#endif /* HEADER_ENV_MENDER_H */
--- 
+--
```
mirzak: Got it!

meta-mender-core/classes/mender-install-ubi.bbclass
mirzak: Tricky indeed. I actually started out using `/dev` but gave up on it, and went with current implementation. Will try to explain in detail, and it will probably help me as well to decide what to do.

Ultimately what Mender client will use to update the volumes is `/dev/ubi0_0` and `/dev/ubi0_1`. And actually `/etc/fstab` can handle `/dev` as well, this was not the case last time I tested but I probably made a mistake first time around. So far so good :).

This problem with using `/dev` is then only that the kernel `root=` argument can not handle it. And this was the reason I dropped it initially as well.

To summarize, UBI volumes can be accessed/referenced in three different ways (this is Linux, u-boot only works with `ubi0:rootfsa` example but I all ready created a special variable for this and it is used only when we need to mount the file system to load kernel and dtb):

- /dev/ubi0_0
- ubi0_0
- ubi0:rootfsa

Above all mean the same, but none of these work everywhere :).

- /dev/ubi0_0
   - works in `/etc/fstab` (for data volume)
    - works for Mender client to update
    - does not work as `root=` argument

- ubi0_0
   - works in `/etc/fstab` (for data volume)
    - does not work for Mender client to use for update
    - works as `root=` argument.

- ubi0:rootfsa
   - works in `/etc/fstab` (for data volume)
   - does not work for Mender client to update
   - works as `root=` argument.

Above means that we need to do something special somewhere. Question is where does it make the most sense?

meta-mender-core/classes/mender-ubimg.bbclass
mirzak: Indeed it has. Will update

meta-mender-core/classes/mender-sdimg.bbclass
mirzak: Ops...Good catch.

meta-mender-toradex-nxp/recipes-bsp/u-boot/u-boot-toradex-fsl-fw-utils_%.bbappend
mirzak: Makes sense, will update.

meta-mender-core/classes/mender-install-ubi.bbclass
mirzak: Actually the `/dev` approach does not work to but in `/etc/fstab`.

```
[   ***] (2 of 2) A start job is running for dev-ubi0_2.device (28s / 1min 30s)[   32.210882] random: nonblocking pool is initialized
[ TIME ] Timed out waiting for device dev-ubi0_2.device.
[DEPEND] Dependency failed for /data.
[DEPEND] Dependency failed for Local File Systems.
[ TIME ] Timed out waiting for device dev-ttymxc0.device.
```

Seems like systemd has an issue with it. `mount -a` works otherwise.

meta-mender-core/recipes-core/base-files/base-files_3.%.bbappend
kacf: Sounds good!

meta-mender-core/classes/mender-helpers.bbclass
kacf: Alright, there doesn't seem to be any option here which isn't at least a little confusing, so I'll try to make a suggestion which is as non-confusing and consistent as possible, hopefully also addressing your other lengthy comment:

* I suppose since a specification without `/dev` is so common, it makes sense to have `MENDER_STORAGE_DEVICE` refer to `ubi0` directly.

* Rename `MENDER_UBOOT_STORAGE_DEVICE` to `MENDER_UBOOT_STORAGE_DEVICE_NUMBER`, not just for ubi, but for existing sdimg as well. A number is what it has been all along, so this is more accurate and descriptive.

* Similarly, rename `get_uboot_device_from_device` to `get_uboot_device_number_from_device` and have it return `${1#ubi}`, IOW just a number.

* Taken together, the previous two changes will make `mender_uboot_dev` into a number. Now replace:
  ```
  ubi part ${mender_uboot_dev}
  ```
  with:
  ```
  ubi part ${mender_uboot_if}${mender_uboot_dev}
  ```

* For Mender opening the ubi block device, we got two choices:

  1. Add the reference `/dev/` when rendering the config file in mender_0.1.bb.

  2. Have the Mender client itself accept bare references to `ubi*`, and transparently add the right path in the code. I can help with this, if you need it.

  I think I vote for the second option, since the Mender client appears to be the only component in this puzzle that isn't capable of handling bare ubi references.

* Create the U-Boot variable `mender_uboot_root_name` as an alias to `mender_kernel_root_name`. This has no functional impact, it's just to have unambiguous variables for uboot and kernel references, respectively.

* Replace:
  ```
  ubifsmount ${mender_kernel_root_name}
  ```
  with:
  ```
  ubifsmount ${mender_uboot_root_name}
  ```

What do you think?

kacf: > All of this was building up to a question. Is it correct of me to re-use the mender_uboot_dev for this use-case, or should I create a new configuration option?

I think you should probably create a new option for this, since the existing variables are reserved for partitions inside UBI. There isn't an existing mechanism for telling where the "outer" MTD partitions are.

Maybe bitbake variable `MENDER_MTD_UBI_DEVICE` and U-Boot variable `mender_mtd_ubi_dev`? I guess this has to be configured separately regardless.

meta-mender-core/classes/mender-install-ubi.bbclass
mirzak: Actually there is a work-around for above, https://lists.freedesktop.org/archives/systemd-devel/2017-January/038197.html.

But do not know if we need it.

meta-mender-core/classes/mender-helpers.bbclass
mirzak: > I suppose since a specification without /dev is so common, it makes sense to have MENDER_STORAGE_DEVICE refer to ubi0 directly.

Agreed, and this was reason that I initially suggested that.

> Rename MENDER_UBOOT_STORAGE_DEVICE to MENDER_UBOOT_STORAGE_DEVICE_NUMBER, not just for ubi, but for existing sdimg as well. A number is what it has been all along, so this is more accurate and descriptive.

I can do this to make the it accurate and descriptive, but will not help my case. Explanation below.

> Similarly, rename get_uboot_device_from_device to get_uboot_device_number_from_device and have it return ${1#ubi}, IOW just a number.

This is where I would run in to problems. There is no way to parse this number (MTD) from the "volume" table (ubi0, ubi0_X), which this would represent. That is why I want to create the "`MENDER_MTD_UBI_DEVICE`"

> Taken together, the previous two changes will make mender_uboot_dev into a number. Now replace:

> ubi part ${mender_uboot_dev}
> with:

> ubi part ${mender_uboot_if}${mender_uboot_dev}

Sadly the `ubi part` command only takes a name as argument (ubi) in my case.

> For Mender opening the ubi block device, we got two choices:

> Add the reference /dev/ when rendering the config file in mender_0.1.bb.

> Have the Mender client itself accept bare references to ubi*, and transparently add the right path in the code. I can help with this, if you need it.

> I think I vote for the second option, since the Mender client appears to be the only component in this puzzle that isn't capable of handling bare ubi references.

I vote second option as well, and was my initial idea.

> Create the U-Boot variable mender_uboot_root_name as an alias to mender_kernel_root_name. This has no functional impact, it's just to have unambiguous variables for uboot and kernel references, respectively.

Yes. I like this, was my though as well.

> Replace:

> ubifsmount ${mender_kernel_root_name}
> with:

> ubifsmount ${mender_uboot_root_name}

Yes on this one.

I will try and summarize my TODO list.

- stick to `ubi0` as MENDER_STORAGE_DEVICE

- WONT rename `MENDER_UBOOT_STORAGE_DEVICE` since this does not really help me out, we could do this in a separate pull-request?

- Create the U-Boot variable `mender_uboot_root_name` as an alias to `mender_kernel_root_name`

- Create `MENDER_MTD_UBI_DEVICE_NAME` and `mender_mtd_ubi_dev_name`, since `ubi part` only takes a name.

- Remove UBI support from mender-helpers? I wont be using ${mender_uboot_boot}, override to `dummy` in mender-install-ubi to prevent it to try to parse it?

And I would end up with the following usage in u-boot:

`ubiargs=ubi.mtd=${mender_mtd_ubi_dev_name} root=${mender_kernel_root} rootfstype=ubifs`

`ubi part ${mender_mtd_ubi_dev_name}`

`ubifsmount ${mender_uboot_root_name}`

If you agree with above I will do a re-spin and we can continue the discussions.

meta-mender-core/classes/mender-install.bbclass
```diff
@@ -16,6 +16,11 @@ MENDER_BOOT_PART ??= "${MENDER_STORAGE_DEVICE_BASE}1"
 MENDER_ROOTFS_PART_A ??= "${MENDER_STORAGE_DEVICE_BASE}2"
 MENDER_ROOTFS_PART_B ??= "${MENDER_STORAGE_DEVICE_BASE}3"
 
+# The names of the two rootfs partitions in the A/B partition layout. By default
+# it is the same name as MENDER_ROOTFS_PART_A and MENDER_ROOTFS_B
+MENDER_ROOTFS_PART_A_NAME ??= "${MENDER_ROOTFS_PART_A}"
+MENDER_ROOTFS_PART_B_NAME ??= "${MENDER_ROOTFS_PART_B}"
```
mirzak: Should end up in `ubi0:rootfs`. 

It is actually override here https://github.com/mirzak/meta-mender/blob/ubifs/meta-mender-core/classes/mender-install-ubi.bbclass#L22-L23.

I realize that I can use at least

`MENDER_ROOTFS_PART_A_NAME ??= ${MENDER_STORAGE_DEVICE}:rootfsa`.

meta-mender-core/classes/mender-helpers.bbclass
kacf: Hehe, sorry I didn't meant to repeat your own suggestions back to you. This is all part of wrapping my own head around this as well! 

>> Rename MENDER_UBOOT_STORAGE_DEVICE to MENDER_UBOOT_STORAGE_DEVICE_NUMBER, not just for ubi, but for existing sdimg as well. A number is what it has been all along, so this is more accurate and descriptive.
> 
> I can do this to make the it accurate and descriptive, but will not help my case. Explanation below.
> 
>> Similarly, rename get_uboot_device_from_device to get_uboot_device_number_from_device and have it return ${1#ubi}, IOW just a number.
> 
> This is where I would run in to problems. There is no way to parse this number (MTD) from the "volume" table (ubi0, ubi0_X), which this would represent. That is why I want to create the "`MENDER_MTD_UBI_DEVICE`"

You don't need to. The number we're after here is only the initial zero, which you get directly from the string. But see next reply as well.

> 
>> Taken together, the previous two changes will make mender_uboot_dev into a number. Now replace:
> 
>> ubi part ${mender_uboot_dev}
>> with:
> 
>> ubi part ${mender_uboot_if}${mender_uboot_dev}
> 
> Sadly the `ubi part` command only takes a name as argument (ubi) in my case.

Actually those two together would add up to the correct string. But maybe you're right that it's better to use `mender_uboot_dev` directly, and include the whole name in it. My suggestion might be overly convoluted.

>> Replace:
> 
>> ubifsmount ${mender_kernel_root_name}
>> with:
> 
>> ubifsmount ${mender_uboot_root_name}
> 
> Yes on this one.
> 
> I will try and summarize my TODO list.
> 
> - stick to `ubi0` as MENDER_STORAGE_DEVICE
> 
> - WONT rename `MENDER_UBOOT_STORAGE_DEVICE` since this does not really help me out, we could do this in a separate pull-request?

Don't worry, you can leave it, it's not really important for this PR.

> - Create the U-Boot variable `mender_uboot_root_name` as an alias to `mender_kernel_root_name`
> 
> - Create `MENDER_MTD_UBI_DEVICE_NAME` and `mender_mtd_ubi_dev_name`, since `ubi part` only takes a name.
> 
> - Remove UBI support from mender-helpers? I wont be using ${mender_uboot_boot}, override to `dummy` in mender-install-ubi to prevent it to try to parse it?

Agreed, it doesn't look like you'll have much use for it.

> 
> And I would end up with the following usage in u-boot:
> 
> `ubiargs=ubi.mtd=${mender_mtd_ubi_dev_name} root=${mender_kernel_root} rootfstype=ubifs`
> 
> `ubi part ${mender_mtd_ubi_dev_name}`
> 
> `ubifsmount ${mender_uboot_root_name}`
> 
> If you agree with above I will do a re-spin and we can continue the discussions.

Yep, looks good!

meta-mender-toradex-nxp/recipes-bsp/u-boot/u-boot-toradex-fsl-fw-utils_%.bbappend
mirzak: Almost done with my rebase, but stumbled upon a problem with above change. When I use the `require u-boot-mender-toradex-nxp.inc` I get following error:

```
Log data follows:
| DEBUG: Executing python function sstate_task_prefunc
| DEBUG: Python function sstate_task_prefunc finished
| DEBUG: Executing shell function do_deploy
| /home/mirzak/project/mender.io/ubifs-support/build/tmp/work/colibri_vf-poky-linux-gnueabi/u-boot-toradex-fsl-fw-utils/v2016.11-v2.6.2b1+gitAUTOINC+b5f6ee6ab1-r0/temp/run.do_deploy.3352: 104: /home/mirzak/project/mender.io/ubifs-support/build/tmp/work/colibri_vf-poky-linux-gnueabi/u-boot-toradex-fsl-fw-utils/v2016.11-v2.6.2b1+gitAUTOINC+b5f6ee6ab1-r0/temp/run.do_deploy.3352: do_deploy: not found
| WARNING: exit code 127 from a shell command.
| ERROR: Function failed: do_deploy (log file is located at /home/mirzak/project/mender.io/ubifs-support/build/tmp/work/colibri_vf-poky-linux-gnueabi/u-boot-toradex-fsl-fw-utils/v2016.11-v2.6.2b1+gitAUTOINC+b5f6ee6ab1-r0/temp/log.do_deploy.3352)
ERROR: Task (/home/mirzak/project/mender.io/ubifs-support/layers/meta-toradex-nxp/recipes-bsp/u-boot/u-boot-toradex-fsl-fw-utils_git.bb:do_deploy) failed with exit code '1'
```

A bit strange, it creates the do_deploy task but does not populate anything in to it. Anything you seen before?

kacf: Hmm, no I can't say I've seen this before.

I don't fully trust poky/master these days though, after the huge "recipe specific sysroots" feature was merged upstream there has been a lot of small issues. Might be worth temporarily reverting to poky/master from a couple of weeks ago, and see if it's still a problem.

mirzak:
My text editor is to "smart", still managed to remove white-spaces and a new line from `0001-Generic-boot-code-for-Mender.patch` :(

mirzak:
Did another re-base. Fixed the unwanted changes to `0001-Generic-boot-code-for-Mender.patch`.

Also had to make some additional changes to the last patch (board support Colibri V50/V61), since was some changes upstream that broke my patch-set.

meta-mender-toradex-nxp/recipes-bsp/u-boot/u-boot-toradex-fsl-fw-utils_%.bbappend
```diff
@@ -0,0 +1,4 @@
+# We provide our custom fw_env.config and that is why we include
+# u-boot-mender-common instead of u-boot-fw-utils-mender
+require recipes-bsp/u-boot/u-boot-mender-common.inc
```
kacf: Mostly a question for future improvement of Mender: What's the reason you need to provide a custom one?

mirzak: I believe we covered this on the mailing list extensively [1].

[1]. https://groups.google.com/a/lists.mender.io/d/msg/mender/1vxZYPT0ysc/P9WJc4KZAwAJ

kacf: Doh, of course. Where is my mind today...

mirzak:
> Sorry for the delay, basically all hell broke loose after upstream poky merged the new sysroot feature. This looks good now, I have one question below, but it doesn't impact the mergability of this.

I understand that there has been a lot with the new poky changes.

> It may take a little bit of time before this can be tested and merged, because of the sysroot issue I mentioned. But it's in my queue, I will get to it as soon as the builds are stable again.

Understood.

My next step is to add support for these changes in `mender` and `mender-artifact`. This will progress a bit slower for me since golang is new to me, but that is part of the experience.

I realize also that there should be some mender-docs changes since we have added configuration variables and changed name on some. I can start preparing this as well.

kacf:
Great, sounds good! Let me know if you need help with the Golang stuff.

kacf:
Just FYI, I haven't forgotten about this. After a very hectic 1.0 release and conferences I am finally ready to return to this.

I can see there are some conflicts, but you can leave them as they are for now. I plan to work a bit in this area anyway, so I will just incorporate this PR into that. I'll let you know if I need any help.

mirzak:
Great.

I understand that it has been hectic lately and there has not been any rush since I have not gotten to work on the Mender client. I have started there and found some "issues" that need to be resolved. Lately I just haven't had the time to work on this, but hopefully it will clear up soon so that I can jump back on this.

/ Mirza from Oslo :)

kacf:
Superseded by #262.

@mirzak: I've only added minimal changes: Some modifications to the commit messages to add changelog data, and some warnings about the experimental nature of this feature. Feel free to check it out!

mirzak:
Cool.

I have also been meaning to suggest that we add a big fat warning if we are to merge this, but you read my mind :).

I will take a closer look on #262 tonight.

PR #233, opened 2017-01-23
qemu: Rearrange memory layout to allow lower memory footprint.

kacf:
The memory apparently starts counting from 0x60000000. By having our
kernel and dtb files load to an address closer to that address, we can
reduce the memory needed to boot. The problem with the previous
address was that it was pointing outside of the address space if the
available memory was reduced.

Consequently, reduce the QEMU memory to 64MiB.


GregorioDiStefano:
Nice!

PR #234, opened 2017-01-23
Reduce demo sdimg size to 400MiB.

kacf:
Generally it's all that's needed, and this will reduce testing times
considerably because there will be much less writing.

At the time of writing, the rootfs fills about 94MiB, so if you double
that for the dual rootfs and add 100MiB+ for the other partitions,
400MiB seems like a sane default.

### DIFF
--- a/meta-mender-demo/conf/layer.conf
+++ b/meta-mender-demo/conf/layer.conf
@@ -11,3 +11,6 @@ BBFILES += "${LAYERDIR}/recipes-*/*/*.bb \
 BBFILE_COLLECTIONS += "mender-demo"
 BBFILE_PATTERN_mender-demo = "^${LAYERDIR}/"
 BBFILE_PRIORITY_mender-demo = "6"
+
+MENDER_STORAGE_TOTAL_SIZE_MB ?= "400"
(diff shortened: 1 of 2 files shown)


meta-mender-demo/conf/layer.conf
```diff
@@ -11,3 +11,5 @@ BBFILES += "${LAYERDIR}/recipes-*/*/*.bb \
 BBFILE_COLLECTIONS += "mender-demo"
 BBFILE_PATTERN_mender-demo = "^${LAYERDIR}/"
 BBFILE_PRIORITY_mender-demo = "6"
+
+MENDER_STORAGE_TOTAL_SIZE_MB ?= "300"
```
bboozzoo: Switch to `core-image-minimal` and your rootfs will be ~20MB with dropbear

kacf: We need `core-image-full-cmdline` for the tests though. Do you think it's alright to leave it at 300 for demo/test purposes?

bboozzoo: Is there a specific feature you need?

PR #239, opened 2017-01-27
Fixes due to changes in Poky

kacf:


kacf:
@bboozzoo: What do you think about the "Adapt bitbake_path..." patch? It's maybe a bit hacky, but on the other hand provides a very smooth and reusable solution.

kacf:
The same recipe will also be used from the jenkins scripts in order to locate the mender-artifact tool.

PR #244, opened 2017-02-01
generated new demo certs for api gateway & storage proxy

estenberg:
After we got the keygen utility, I generated new certs and keys and think we should use them in demo environment so that the keys generated by keygen are well-tested.

Generated new certs for API Gateway and Storage Proxy as output from the [keygen utility](https://github.com/mendersoftware/integration/blob/master/keygen).

Then appended them both into server.crt, as described in the docs: https://docs.mender.io/Artifacts/Building-for-production#preparing-the-client-certificates.

This must be merged together with integration PR: https://github.com/mendersoftware/integration/pull/152


estenberg:
There was just one cert in server.crt before, how is this possible? I thought we needed both API Gateway and Storage Proxy cert?

estenberg:
@GregorioDiStefano @kacf is this safe to merge, or are there other places these default keys/certs are stored?

kacf:
> There was just one cert in server.crt before, how is this possible? I thought we needed both API Gateway and Storage Proxy cert?

Not sure what you mean, there are clearly two certificates in the existing file, and you are providing two new ones as well.

GregorioDiStefano:
Looks good.

estenberg:
OK, thank you.
My oversight on the two certificates, clearly there were two already, like you say. :)

I am merging both PRs then.

PR #246, opened 2017-02-03
Modify checksum after adding new dependency to mender.

pasinskim:


pasinskim:
Mere with https://github.com/mendersoftware/mender/pull/181

kacf:
We dropped this dependency.

PR #251, opened 2017-02-08
Morty update

kacf:


kacf:
Failed on beaglebone because the resulting update is only 16MB, which is... suspiciously small.

kacf:
Will handle the Beaglebone later. Merging.

PR #252, opened 2017-02-08
Split Mender binaries into 1.0.0b1 and git versions.

kacf:

### DIFF
--- a/meta-mender-core/classes/mender-install.bbclass
+++ b/meta-mender-core/classes/mender-install.bbclass
@@ -63,6 +63,10 @@ MENDER_STORAGE_RESERVED_RAW_SPACE ??= "0"
 PREFERRED_VERSION_go-cross-arm ?= "1.7.%"
 PREFERRED_VERSION_go-native ?= "1.7.%"
 
+PREFERRED_VERSION_mender ?= "1.0.%"
+PREFERRED_VERSION_mender-artifact ?= "1.0.%"
+PREFERRED_VERSION_mender-artifact-native ?= "1.0.%"
+
 IMAGE_INSTALL_append = " \
     mender \
     ca-certificates \
--- a/meta-mender-core/recipes-mender/mender-artifact/mender-artifact_1.0.0b1.bb
+++ b/meta-mender-core/recipes-mender/mender-artifact/mender-artifact_1.0.0b1.bb
@@ -1,0 +1,9 @@
+require mender-artifact.inc
+
+SRC_URI = "git://github.com/mendersoftware/mender-artifact.git;protocol=https;branch=1.0.x"
+
+# Tag: 1.0.0b1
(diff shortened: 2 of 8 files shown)


kacf:
@pasinskim, @bboozzoo

kacf:
Currently testing [here](https://ci.cfengine.com/job/yoctobuild-pr/335/).

kacf:
Hmm, looks like this causes the version number to be 1.0.0b1 for *all* our builds, instead of a git revision for the dev builds. But I think we can fix that after the release.

kacf:
> Hmm, looks like this causes the version number to be 1.0.0b1 for all our builds, instead of a git revision for the dev builds. But I think we can fix that after the release.

I have attempted to fix this in https://github.com/mendersoftware/mender-qa/pull/80, but it needs to be merged after this one.

kacf:
Updated to fix acceptance test. Rerunning [here](https://ci.cfengine.com/job/yoctobuild-pr/337/).

meta-mender-core/classes/mender-install.bbclass
```diff
@@ -63,6 +63,10 @@ MENDER_STORAGE_RESERVED_RAW_SPACE ??= "0"
 PREFERRED_VERSION_go-cross-arm ?= "1.7.%"
 PREFERRED_VERSION_go-native ?= "1.7.%"
 
+PREFERRED_VERSION_mender ?= "1.0%"
```
pasinskim: Shouldn't we have here `PREFERRED_VERSION_mender ?= "1.0.%"`?

meta-mender-core/classes/mender-install.bbclass
```diff
@@ -63,6 +63,10 @@ MENDER_STORAGE_RESERVED_RAW_SPACE ??= "0"
...
+PREFERRED_VERSION_mender-artifact ?= "1.0%"
+PREFERRED_VERSION_mender-artifact-native ?= "1.0%"
```
pasinskim: Same as above. I am not sure if we need a dot after `1.0`.

meta-mender-core/recipes-mender/mender-artifact/mender-artifact_1.0.0b1.bb
```diff
@@ -0,0 +1,9 @@
+require mender-artifact.inc
+
+SRC_URI = "git://github.com/mendersoftware/mender-artifact.git;protocol=https;branch=1.0.x"
+
+# Tag: 1.0.0b1
+SRCREV = "77326b288c70cd713e7ad15d2a084b6ee797e8ff"
```
pasinskim: Nice!

meta-mender-core/classes/mender-install.bbclass
kacf: You're right, we should have a dot. Will fix!

PR #254, opened 2017-02-14
Include .._EXTRA_SPACE in image size calculations.

kacf:
For example core-image-minimal will add space to this unconditionally
if systemd is enabled.


kacf:
DO NOT merge. This is currently broken and only for review right now.

PR #263, opened 2017-03-19
add missing expand argument to getVar

mirzak:
This fixes an expansion error during parsing of mender_git and
mender-artifact_git recipes.


bboozzoo:
IIRC there was a patch to bitbake that made `getVar()` expansion on by default (the second parameter usually passed as True), see here: https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?id=c4e2c59088765d1f1de7ec57cde91980f887c2ff 

IMO the change is not needed for master, as we're supposed to work with Poky/OE-core master where they have already transitioned to not passing `True`.

mirzak:
Aah I see.

I was using morty on Poky/OE-core and some other repository's (for stability) while using master of meta-mender doing some work. That should explain why I got errors.

Agreed that this is not needed on master after seeing the upstream Poky/OE-core changes that where made regarding this. Will keep this locally while working then :)

PR #264, opened 2017-03-20
Turn too-big-update warning into error.

kacf:
Also remove IMAGE_ROOTFS_EXTRA_SPACE reference, since we handle that
automatically now, and instead suggest changing
MENDER_STORAGE_TOTAL_SIZE_MB.


kacf:
@mirzak: Like we discussed earlier.

mirzak:
Looks good to me.

Though I have not had a chance to test the "automatically handle IMAGE_ROOTFS_EXTRA_SPACE" yet to see if my issues disappear.

kacf:
Thanks!

This is being tested in Jenkins, but some unrelated issue is preventing it from passing. I will pick this up once that's fixed.

PR #279, opened 2017-04-26
Small fixes

balister:


meta-mender-core/recipes-core/systemd/systemd_%.bbappend
```diff
@@ -1,4 +1,4 @@
-PACKAGECONFIG += "networkd resolved"
+PACKAGECONFIG_append = "networkd resolved"
```
kacf: Doesn't `_append` require a space before `networkd`?

balister: I used this to "validate":
[balister@thuvia build]$ bitbake -e systemd | grep ^PACKAGECONFIG
PACKAGECONFIG="xz                    ldconfig                    xkbcommon                    rfkill                    binfmt                    randomseed                    machined                    backlight                    vconsole                    quotacheck                    hostnamed                    myhostname sysusers                    hibernate                    timedated                    timesyncd                    localed                    ima                    smack                    logind                    firstboot                    utmp                    polkit  networkd resolved"

So I got 2 spaces after polkit. I see one space, I guess the LF adds another. At any rate, you are correct, consulting the manual says there should be a leading space to sbe safe from recipe changes. Good catch.

bboozzoo: I'm trying to find a reason why we even enable `networkd` & `resolved` in meta-mender-core. Aside from shipping networkd config files, which should really be a part of meta-mender-demo, there's nothing that ties us to networkd. Similarly, `resolved` is enabled for the build, but unless you add `resolve` to `nsswitch.conf` it's not going to be used. I've posted a patch that adds resolve to nsswitch http://lists.openembedded.org/pipermail/openembedded-core/2016-September/126738.html but didn't bother to follow up after systemd got updated. Also, IIRC you won't be even able to use `resolved` if you build with musl (as there was no nsswitch support). 

Lastly, there's no real dependency on systemd either.

balister: I was wondering why you had to mod systemd config also. It would be nice to minimize changes to recipes like this.

kacf:
Thanks, fixed the space and merged manually.

PR #280, opened 2017-04-27
fstab: fix support for SD card supported BSPs

jacobbarsoe:
default settings in fstab already include auto.
Remove redundant auto flag as it will fail auto
mounting during boot.


jacobbarsoe:
done deal

kacf:
Thanks! Merged manually to morty and will cherry-pick to master.

PR #284, opened 2017-05-02
fstab: fix support for SD card supported BSPs

kacf:
defaults settings in fstab already include auto.
Remove redundant auto flag as it will fail auto
mounting during boot.

Changelog: Commit


kacf:
Don't merge! Testing only.

PR #285, opened 2017-05-03
mender: Make mender wait till network configuration is done

mfischer:
Add Wants=network-online.target to the service file to make
mender wait for the network to be up, otherwise it might fail.

Changelog: Title

Signed-off-by: Moritz Fischer <moritz.fischer@ettus.com>

### DIFF
--- a/meta-mender-core/recipes-mender/mender/files/mender.service
+++ b/meta-mender-core/recipes-mender/mender/files/mender.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=Mender OTA update service
 After=systemd-resolved.service
+Wants=network-online.target
 
 [Service]
 Type=idle


kacf:
It's intentional that Mender starts without depending on the network service. When you make an update, it will keep trying to contact the server to verify that the network is working before deciding to commit or roll back. But if you add this dependency, then Mender won't start if the network doesn't start, and the rollback won't be performed.

What's the failure you're having exactly?

mirzak:
Jumping in here :). 

Agreed with @kacf that this is not the solution. 

I have experienced a similar issue though due to a very fast poll time (the one that is used i meta-mender-demo). 

If I use a WiFi on my device it can sometimes take a while to get a network connection, and sometimes Mender fails to update due to the fact that it can not rapport state to the server due to WiFi not been registered yet and it will rollback. 

But this issue is only when poll timeout is very low and shold not really pose a problem in "production" where timeouts are in minutes instead of seconds.

mirzak:
I even fixed it a while back https://github.com/mirzak/meta-mender-demo/commit/09c2d5d3b6e7638052450e57bc58812cb5a598c3

PR #286, opened 2017-05-04
MEN-1173: Add acceptance test for signed updates

kacf:


tests/acceptance/common.py
```diff
@@ -388,3 +388,26 @@ def path_restore():
     request.addfinalizer(path_restore)
 
     return os.environ['PATH']
+
+@pytest.fixture(scope="session")
+def signing_key(request):
+    subprocess.check_call(["openssl", "genrsa", "-out", "private.pem", "2048"])
```
pasinskim: This is probably not changing a lot, but maybe we can have some pregenerated and hardcoded keys instead of generating new ones all the time.

tests/acceptance/test_update.py
```diff
@@ -78,6 +82,35 @@ def get_env_checksums(bitbake_variables):
...
 
+    @staticmethod
+    def corrupt_middle_byte(fd):
+        # Corrupt the middle byte in the contents.
+        middle = int(os.fstat(fd.fileno()).st_size / 2)
+        fd.seek(middle)
+        middle_byte = int(fd.read(1).encode("hex"), base=16)
+        fd.seek(middle)
+        # Flip lowest bit.
+        fd.write("%c" % (middle_byte ^ 0x1))
```
pasinskim: Just a heads up here. The signature file is base64 encoded data you are reading from file and then you are flipping some "random" bit. Are you sure we won't end up with invalid base64 data at the end?

tests/acceptance/test_update.py
```diff
@@ -206,6 +239,172 @@ def test_network_based_image_update(self, successful_image_update_mender, bitbak
...
+                                            key=False,
+                                            checksum_ok=True,
+                                            update_written=True,
+                                            success=True),
+                              SignatureCase(label="Not signed, key present",
+                                            signature=False,
+                                            signature_ok=False,
+                                            key=True,
+                                            checksum_ok=True,
+                                            # update_written should be False, see MEN-1198.
```
pasinskim: Soon will be :)

tests/acceptance/test_update.py
kacf: Yep, base64 uses six lowest bits as payload, so flipping the lowest one should be safe.

tests/acceptance/common.py
kacf: Yes, that makes sense, I will switch to that.

kacf:
Updated and rerunning.

PR #290, opened 2017-05-05
MEN-1182: Default artifact demo signatures

kacf:


kacf:
I have a feeling this might break some of the server tests. I guess we'll see.

kacf:
This breaks a lot of stuff, we need to rethink this, see https://tracker.mender.io/browse/MEN-1182.

PR #292, opened 2017-05-15
Add tests for unsigned image of version 1.

kacf:
This is important so that version 1 can not be used as a backdoor
circumventing the signature verification.

### DIFF
--- a/tests/acceptance/test_update.py
+++ b/tests/acceptance/test_update.py
@@ -102,13 +102,22 @@ class SignatureCase:
     update_written = False
     success = True
 
-    def __init__(self, label, signature, signature_ok, key, checksum_ok, update_written, success):
+    def __init__(self,
+                 label,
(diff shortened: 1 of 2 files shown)


tests/acceptance/test_update.py
```diff
@@ -301,13 +332,18 @@ def test_signed_updates(self, sig_case, bitbake_path, bitbake_variables, signing
...
         # Generate artifact with or without signature.
         if sig_case.signature:
-            sign_args = "-k %s" % signing_key.private
-        else:
-            sign_args = ""
+            artifact_args += " -k %s" % signing_key.private
+
+        # Generate artifact with specific version. None means default.
+        if sig_case.artifact_version is not None:
+            artifact_args = " -v %d" % sig_case.artifact_version
```
kacf: Crap, this should be `+=`.

PR #293, opened 2017-05-15
[RFC] auto-reboot: Add systemd service to force an automatic reboot.

drewmoseley:
This allows us to simulate a bad device update to test rollback
capability.

Signed-off-by: Drew Moseley <drew.moseley@mender.io>

### DIFF
--- a/meta-mender-demo/recipes-mender/auto-reboot/auto-reboot_1.0.bb
+++ b/meta-mender-demo/recipes-mender/auto-reboot/auto-reboot_1.0.bb
@@ -1,0 +1,19 @@
+DESCRIPTION = "Systemd Service script to force a reboot to test automatic rollbacks by the Mender client."
+HOMEPAGE = "https://mender.io"
+LICENSE = "Apache-2.0"
+
+SRC_URI = " \
+	file://auto-reboot.service \
+	file://LICENSE \
+"
+LIC_FILES_CHKSUM = "file://${S}/../LICENSE;md5=85c94fa2073dc7058c4c77ba6203061c"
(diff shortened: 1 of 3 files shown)


drewmoseley:
@kacf any interest in this or something similar?

kacf:
Ah, we are testing rollback functionality already! Check [this test](https://github.com/mendersoftware/meta-mender/blob/master/tests/acceptance/test_update.py#L117-L143).

drewmoseley:
That only seems to work with QEMU right?  This recipe allows it to be tested/demoed on the Beaglebone using the full backend integration. Admittedly it's heavier-weight as it requires a Yocto rebuild but I have used it successfully for a demo that I put together.

kacf:
No, all the client acceptance tests run for the Beaglebone as well (well, except for the fact that it's out of order right now, but that's unrelated to the test). We also have similar "broken image" tests in the integration repository, so this should be pretty well covered already.

drewmoseley:
@kacf is there any way to use that artifact deployed through a backend instance to test the entire workflow?  I was able to do the standalone deployment but ultimately I'd like a demo setup through the backend.

Would this kind of change be better submitted through meta-mender-demo?

kacf:
I'm not sure I understand what you're after... Ping me on Slack when you get the chance!

drewmoseley:
OK.  I am able to test the rollback now using an image created similar to what is done here: https://github.com/mendersoftware/meta-mender/blob/master/tests/acceptance/test_update.py#L117-L143

This is no longer needed.

PR #294, opened 2017-05-17
MENDER_DATA_PART_DIR: copy files with rsync

dwsilver:
The MENDER_DATA_PART_DIR variable is not documented, but seems intended to enable preloading the data partition with the contents of a directory. The current implementation recursively copies the directory and its contents, resulting in several copies of the files. 

This change uses rsync to copy the contents of the directory so the contents of the directory pointed to by MENDER_DATA_PART_DIR get copied to the data partition once as expected.

### DIFF
--- a/meta-mender-core/classes/mender-sdimg.bbclass
+++ b/meta-mender-core/classes/mender-sdimg.bbclass
@@ -110,7 +110,7 @@ IMAGE_CMD_sdimg() {
     mkdir -p "${WORKDIR}/data"
 
     if [ -n "${MENDER_DATA_PART_DIR}" ]; then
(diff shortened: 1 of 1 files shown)


kacf:
Thanks! Indeed this is for populating the data directory. The reason it's not documented is because we want to move away from it, but this requires upstream support from Yocto and is work in progress.

I tweaked the patch slightly and resubmitted [here](#299). We'll see how it does in our test farm.

dwsilver:
Great, thanks! Can you make sure this gets cherry-picked to the morty branch once it lands?

kacf:
Sure.

kacf:
#299 merged in master and morty, and on its way to pyro.

PR #300, opened 2017-05-26
Add example recipe to show how to deploy files into Mender persistent…

drewmoseley:
… data partition

Signed-off-by: Drew Moseley <drew.moseley@mender.io>

### DIFF
--- a/meta-mender-demo/recipes-mender/hello-mender/hello-mender.bb
+++ b/meta-mender-demo/recipes-mender/hello-mender/hello-mender.bb
@@ -1,0 +1,20 @@
+DESCRIPTION = "Mender test recipe for persistent data files"
+LICENSE = "Apache-2.0"
+
+inherit deploy
+
+do_compile() {
+    echo 'Hello Mender config file' > hello-mender.cfg
+    echo 'persistent data test' > persistent.txt
+}
+
+do_install() {
+    install -d ${D}${sysconfdir}
+    install -m 0644 hello-mender.cfg ${D}${sysconfdir}
+}
+
+do_deploy() {
+    install -d ${DEPLOYDIR}/persist
+    install -m 0644 persistent.txt ${DEPLOYDIR}/persist
+}
+addtask do_deploy after do_compile before do_build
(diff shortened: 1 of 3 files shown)


drewmoseley:
@kacf please take a look at this.  This is my initial attempt at deploying files to the persistent data partition using MENDER_DATA_PART_DIR.  Is this consistent with what you were expecting?

meta-mender-demo/recipes-mender/hello-mender/hello-mender.bb
```diff
@@ -0,0 +1,20 @@
+DESCRIPTION = "Mender test recipe for persistent data files"
+LICENSE = "Apache-2.0"
+
+inherit deploy
+
+do_compile() {
+    dd if=/dev/urandom of=persistent.dat bs=4k count=100
+    dd if=/dev/urandom of=hello-mender bs=4k count=100
```
kacf: Maybe use something predictable so it's easier to see that the right content made it onto the `/data` partition?

drewmoseley:
@kacf I updated this to provide known contents for verification rather than relying on /dev/urandom.

I had not considered what to do about documenting this; it's fairly straightforward as is.  Do we want do put this in the docs given that you are planning a better upstream solution?

drewmoseley:
On further review, we don't want to merge this as is.  Given the bbappend to the core-image-full-cmdline recipe, this will cause any build that doesn't include the hello-mender recipe to fail.  I suppose we could add hello-mender in meta-mender-demo/recipes-core/images/core-image-full-cmdline.bbappend but I'm not sure that's the right approach.  Thoughts?

kacf:
I think that would be the right approach, yes. Why are trying to stick to bug fixes only in pyro; given that this is mostly for demonstration, and doesn't add actual value to the image, I think we should keep it in master only.

drewmoseley:
@kacf I updated this to include the package in the core-image-full-cmdline build from the meta-mender-demo layer.  That should resolve the build failure here and allow this to be merged.

drewmoseley:
Doc updates for this are available here: https://github.com/mendersoftware/mender-docs/pull/162

drewmoseley:
I added the changelog details.

I have no problem with the change you recommend in the mender-sdimg.bbclass.

kacf:
Great, if you can add another commit where you make the chown fix I described, the next test run should succeed.

kacf:
Famous last words :-)

drewmoseley:
chown patch added and tested.

kacf:
Thanks! I'll make sure this lands in pyro as well.

kacf:
Actually, only the second patch, since we decided to keep the demo recipe in master only.

PR #309, opened 2017-06-09
Fix sdimg being generated outside of sstate mechanism.

kacf:
Changelog: Fix a bug which caused sdimg generation to bypass the
sstate mechanism. This could lead to broken builds in some corner
cases, and it also means that sdimg images would never be cleaned up.

### DIFF
--- a/meta-mender-core/classes/mender-sdimg.bbclass
+++ b/meta-mender-core/classes/mender-sdimg.bbclass
@@ -166,7 +166,7 @@ EOF
     echo "### End of contents of wks file ###"
 
     # Call WIC
-    outimgname="${DEPLOY_DIR_IMAGE}/${IMAGE_NAME}.sdimg"
+    outimgname="${IMGDEPLOYDIR}/${IMAGE_NAME}.sdimg"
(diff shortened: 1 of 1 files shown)


kylemanna:
This fixed the issue I was running in to an described [on the mailing list](https://groups.google.com/a/lists.mender.io/forum/#!topic/mender/yt51xwaQy0o) with respect to accumulating `sdimg`s.

Test case:
1. Examine contents of `build/deploy` directory
2. `bitbake -c cleansstate menderized-image && bitbake menderized-image`
3. Verify old `sdimg` files (post patch) were deleted on `cleansstate`

Without this patch the `sdimg` files would require manual clean-up and were quite large.

kacf:
Thanks for testing @kylemanna!

kylemanna:
What's the policy on back-porting?  Would be nice to get this pulled in to the `morty` branch.  My test was on the `morty:HEAD` with the single patch cherry-picked.

Can send pull request if it helps.

kacf:
We only support the latest poky release currently, but we accept pull requests for older branches. I merged yours!

kylemanna:
Thanks.

PR #310, opened 2017-06-09
MEN-1014: Add mender-state-scripts class and example.

kacf:
The new mender-state-scripts class can be used in a recipe to turn
that recipe into a state script recipe. Any script or binary that the
recipe provides using the class will become part of the Mender update
as a state script.

There are several ways to include them:

* Put the script/binary in ${MENDER_STATE_SCRIPTS_DIR} during the
  do_compile stage.

* Put an archive in SRC_URI that extracts to
  ${PN}-${PV}/mender-state-scripts, where ${PN}-${PV} are the recipe
  name and version.

* Set the variable MENDER_STATE_SCRIPTS manually to a directory that
  contains the scripts.

Note that the three methods should not be mixed.

Changelog: Commit

### DIFF
--- a/meta-mender-core/classes/mender-artifactimg.bbclass
+++ b/meta-mender-core/classes/mender-artifactimg.bbclass
@@ -37,18 +37,22 @@ IMAGE_CMD_mender () {
         bberror "MENDER_DEVICE_TYPES_COMPATIBLE variable cannot be empty."
     fi
 
+    extra_args=
+
     if [ -n "${MENDER_ARTIFACT_SIGNING_KEY}" ]; then
-        signing_args="-k ${MENDER_ARTIFACT_SIGNING_KEY}"
-    else
-        signing_args=
+        extra_args="$extra_args -k ${MENDER_ARTIFACT_SIGNING_KEY}"
+    fi
+
+    if [ -d "${DEPLOY_DIR_IMAGE}/mender-state-scripts" ]; then
+        extra_args="$extra_args -s ${DEPLOY_DIR_IMAGE}/mender-state-scripts"
     fi
 
     mender-artifact write rootfs-image \
         -n ${MENDER_ARTIFACT_NAME} -t "$devs_compatible" \
-        $signing_args \
+        $extra_args \
         -u ${IMGDEPLOYDIR}/${IMAGE_BASENAME}-${MACHINE}.${ARTIFACTIMG_FSTYPE} \
         ${MENDER_ARTIFACT_EXTRA_ARGS} \
-        -o ${IMGDEPLOYDIR}/${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.mender \
+        -o ${IMGDEPLOYDIR}/${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.mender
 }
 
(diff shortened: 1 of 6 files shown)


kacf:
Note that this depends on https://github.com/mendersoftware/mender-artifact/pull/76.

kacf:
@drewmoseley: Here is the implementation of what I tried to document in https://github.com/mendersoftware/mender-docs/pull/166. Dunno if you wanna look at the gory details, but does the overall approach make sense?

meta-mender-core/classes/mender-artifactimg.bbclass
```diff
@@ -37,18 +37,22 @@ IMAGE_CMD_mender () {
...
     fi
 
     mender-artifact write rootfs-image \
         -n ${MENDER_ARTIFACT_NAME} -t "$devs_compatible" \
-        $signing_args \
+        $extra_args \
         -u ${IMGDEPLOYDIR}/${IMAGE_BASENAME}-${MACHINE}.${ARTIFACTIMG_FSTYPE} \
         ${MENDER_ARTIFACT_EXTRA_ARGS} \
-        -o ${IMGDEPLOYDIR}/${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.mender \
+        -o ${IMGDEPLOYDIR}/${IMAGE_NAME}.mender
```
drewmoseley: Did you deliberately drop ${IMAGE_NAME_SUFFIX} here?

meta-mender-core/classes/mender-state-scripts.bbclass
```diff
@@ -0,0 +1,85 @@
...
+# Default is to look in these two directories for scripts.
+MENDER_STATE_SCRIPTS ?= "${S}/mender-state-scripts ${MENDER_STATE_SCRIPTS_DIR}"
+
+do_create_default_script_dirs() {
+    # Doing this automatically has two advantages: The user can install directly
+    # into the directories with no preparation. And, we can complain if any of
+    # the targets in MENDER_STATE_SCRIPTS doesn't exist, because a directory
+    # that exists, but is empty, is different from not existing at all.
+
+    mkdir -p ${S}/mender-state-scripts ${B}/mender-state-scripts
```
drewmoseley: Should this be "mkdir -p ${MENDER_STATE_SCRIPTS}"?

meta-mender-core/classes/mender-artifactimg.bbclass
kacf: No, that's not intentional! Well spotted. This was some debug code from earlier, so I have removed it.

meta-mender-core/classes/mender-state-scripts.bbclass
kacf: I think not, actually. The intention is that the variable can point to a nonexistent directory, and this should be an error. So we can't create it or the error would be silenced. This `mkdir` is specifically about our two defaults only.

meta-mender-core/classes/mender-artifactimg.bbclass
kacf: Well, actually *restored* it, not removed it.

kacf:
Fixed review issues, rerunning [here](https://ci.cfengine.com/job/yoctobuild-pr/604).

PR #314, opened 2017-06-14
qemu raw flash

bboozzoo:
This PR adds support for raw flash to qemu. You need to use qemu master (or 2.9) to be able to use generated image.

Short summary:

There's a separate machine config `vexpress-qemu-flash`, as its setup is significantly different. The machine config already defines some MENDER variables (see vexpress-qemu-flash.conf) as we know the size of the storage upfront.

There's a number of u-boot patches that are queued up right here for now: https://github.com/bboozzoo/u-boot/commits/bboozzoo/vexpress-a9-ubi-mender (they will be added once I verify that things are working). The 
`vexpress_ca9x4: enable booting from ubifs` does some significant changes to uboot config and flash layout, these are:
* enable UBI, UBIFS, MTD on CFI, MTD concatenation (so our sepate 64MB banks (`nor0`, `nor1`) will appear as single 128MB `nor2` device)
* defines the following MTD partitions: `mtdparts=40000000.flash:1m(u-boot)ro,1m(u-boot-env),-(ubi)`
* uboot environment starts at the 1MB offset from the bottom of the flash (before it was at the top of the flash - 2 * env size)
* environment has been trimmed to just the necessary minimum

Patch `meta-mender-core/install-ubi: [DO NOT MERGE] comment out warnings...` will be dumped before this is merged.

There's some warnings from `configcheck` task when building linux-yocto. I don't know why, but configcheck may be picking incorrect config for verification. Needs investigation.

I had to add some workarounds for u-boot-fw-utils and generating `fw_env.config`. The code generating `fw_env.config` would end up using incorrect device (defaulting to `MENDER_STORAGE_DEVICE`) and generated incorrect config. @mirzak did not have this problem as he was using `u-boot-toradex-fsl-fw-utils`, where he could selectively include `meta-mender-core` files and ended up shipping a premade `fw_env.confg`. I also don't want to introduce more `MENDER_*` varaibles, and I cannot selectively include files, hence the workaround to provide a default `fw_env/.config.default`. (BTW adding `do_compile_append_vexpress-qemu-flash` ended up being added before `do_compile_append` code)

Also, there's a problem with data partition, for now I get:
```
[    4.930933] UBIFS error (ubi0:2 pid 718): validate_sb: bad superblock, error 4
[    4.932212]  magic          0x6101831
[    4.932403]  crc            0x293fadba
[    4.932599]  node_type      6 (superblock node)
[    4.932813]  group_type     0 (no node group)
[    4.933005]  sqnum          1
[    4.933149]  len            4096
[    4.933350]  key_hash       0 (R5)
[    4.933530]  key_fmt        0 (simple)
[    4.933716]  flags          0x0
[    4.933880]  big_lpt        0
[    4.934031]  space_fixup    0
[    4.934178]  min_io_size    8
[    4.934329]  leb_size       262016
[    4.934490]  leb_cnt        17
[    4.934624]  max_leb_cnt    1024
[    4.934762]  max_bud_bytes  8388608
[    4.934910]  log_lebs       4
[    4.935043]  lpt_lebs       2
[    4.935168]  orph_lebs      1
[    4.935301]  jhead_cnt      1
[    4.935435]  fanout         8
[    4.935583]  lsave_cnt      256
[    4.935728]  default_compr  1
[    4.935878]  rp_size        0
[    4.936006]  rp_uid         0
[    4.936139]  rp_gid         0
[    4.936296]  fmt_version    4
[    4.936542]  time_gran      1000000000
[    4.936925]  UUID           BD9A5147-7BB4-4BA1-9DE9-C88045D28D5F
```
@kacf @pasinskim

### DIFF
--- a/meta-mender-qemu/Dockerfile
+++ b/meta-mender-qemu/Dockerfile
@@ -20,7 +20,8 @@ FROM alpine:3.4
 
 # Install QEMU
 RUN apk update && apk upgrade && \
-    apk add qemu-system-arm bash e2fsprogs-extra python3 && \
+    apk add qemu-system-arm util-linux \
+            bash e2fsprogs-extra python3 && \
     rm -rf /var/cache/apk/*
 
 ARG VEXPRESS_IMAGE=scripts/docker/empty-file
--- a/meta-mender-qemu/Dockerfile
+++ b/meta-mender-qemu/Dockerfile
@@ -1,8 +1,8 @@
 # Usage of Docker image.
 #
 # While building:
-# --build-arg VEXPRESS_IMAGE=<sdimg>
-#       sdimg image to add to the Docker image
+# --build-arg VEXPRESS_IMAGE=<sdimg|vexpress-nor>
+#       image to add to the Docker image
 # --build-arg UBOOT_ELF=<u-boot.elf>
 #       U-Boot ELF image to add to Docker image.
 #
@@ -27,8 +27,8 @@ RUN apk update && apk upgrade && \
 ARG VEXPRESS_IMAGE=scripts/docker/empty-file
 ARG UBOOT_ELF=scripts/docker/empty-file
 
-ADD $UBOOT_ELF ./u-boot.elf
-ADD $VEXPRESS_IMAGE ./core-image-full-cmdline-vexpress-qemu.sdimg
+COPY $UBOOT_ELF ./u-boot.elf
+COPY $VEXPRESS_IMAGE .
 
 ADD scripts/mender-qemu ./
 ADD scripts/docker/entrypoint.sh ./
@@ -37,6 +37,7 @@ ADD scripts/docker/setup-mender-configuration.py ./
 # Delete images if they are empty. This is to save space if the artifacts are
 # mounted on /mnt/build instead.
 RUN if [ `stat -c %s core-image-full-cmdline-vexpress-qemu.sdimg` -eq 0 ]; then rm -f core-image-full-cmdline-vexpress-qemu.sdimg; fi
+RUN if [ `stat -c %s core-image-full-cmdline-vexpress-qemu.vexpress-nor` -eq 0 ]; then rm -f core-image-full-cmdline-vexpress-qemu.vexpress-nor; fi
 RUN if [ `stat -c %s u-boot.elf` -eq 0 ]; then rm -f u-boot.elf; fi
 
 RUN chmod +x entrypoint.sh mender-qemu
--- a/meta-mender-qemu/conf/machine/vexpress-qemu-flash.conf
+++ b/meta-mender-qemu/conf/machine/vexpress-qemu-flash.conf
@@ -1,0 +1,24 @@
+#@TYPE: Machine
+#@NAME: vexpress-qemu
+
+#@DESCRIPTION: Machine configuration for QEMU running vexpress with system on MTD device
+
+require conf/machine/include/qemu.inc
+require conf/machine/include/vexpress.inc
+
+IMAGE_FSTYPES_append = " ubifs ubi"
+
+MACHINE_ESSENTIAL_EXTRA_RDEPENDS = "kernel-image kernel-devicetree"
+MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += "mtd-utils mtd-utils-ubifs mtd-utils-jffs2"
+
+MKUBIFS_ARGS = "-m 1 -e 262016 -c 1024"
+UBINIZE_ARGS = "-p 256KiB -m 1 -O 64"
+
+# NOTE: normally we'd pick up overrides for vexpress-qemu, disable them for now
+# to avoid incorrect u-boot patches being applied (on a side note, maybe rework
+# u-boot patches to be applied for any machine?)
+
+#MACHINEOVERRIDES =. "vexpress-qemu-flash:vexpress-qemu:"
+
+# Needed to skip particular QA checks that don't apply to U-boot's binary.
+INSANE_SKIP_u-boot += "ldflags textrel"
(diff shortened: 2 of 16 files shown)


bboozzoo:
Pushed the missing u-boot patches.

Not sure if useful, but I'm using the following script to generate NOR flash contents for qemu:

```
#!/bin/sh

rm -f nor0 nor1
dd if=/dev/zero bs=1M count=128 | tr '\000' '\377' > nor-full
dd if=core-image-minimal-vexpress-qemu-flash.ubimg of=nor-full bs=1M seek=2 conv=notrunc
split -b $((64*1024*1024)) -a 1 -d nor-full nor
ls -l nor0 nor1
```

qemu is stared like this:
```
qemu-system-arm -M vexpress-a9 -m 128M \
    -kernel ./u-boot.elf -net nic,macaddr=52:54:00:b1:4e:ab -net user \
     -nographic \
    -drive file=nor0,if=pflash,format=raw \
    -drive file=nor1,if=pflash,format=raw
```

meta-mender-qemu/conf/machine/vexpress-qemu-flash.conf
```diff
@@ -22,3 +22,15 @@ UBINIZE_ARGS = "-p 256KiB -m 1 -O 64"
...
+MENDER_DATA_PART_SIZE_MB = "16"
+# combined NOR flash is 128MB
+MENDER_STORAGE_TOTAL_SIZE_MB = "128"
+# ??
+MEDNDER_BOOT_PART_SIZE_MB = "1"
+# align to PEB size 256k
+MENDER_PARTITION_ALIGNMENT_KB = "256"
+#MENDER_STORAGE_RESERVED_RAW_SPACE = "524288"
+
+MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET = "0x100000"
```
bboozzoo: MTD is partitioned like this:
```
1m(u-boot)ro,
1m(u-boot-env)ro,
-(ubi)
```
`meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc` hijacks `CONFIG_ENV_OFFSET` and sets it to the value of `MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET` thus the offset must match the partitioning for u-boot to work. Just for the record, the offset is a it's counted from the beginning of the flash.

The same offset is used in `meta-mender-core/recipes-bsp/u-boot/u-boot-fw-utils-mender.inc` where it's used to generate `fw_env.config`. However, there is a partition defined for uboot's env, and it will appear as /dev/mtd1 covering flash address space 0x100000 to 0x200000. Using the same value there would yield incorrect configuration.

Frankly, I'm not really fond of any of these settings being autoconfigured. In cases other than building for MMC you're more likely to get it wrong and produce something unusable.

meta-mender-qemu/Dockerfile
```diff
@@ -16,7 +16,7 @@
 # -e TENANT_TOKEN=<token>
 #       Use token as tenant token for client.
 
-FROM alpine:3.4
+FROM alpine:latest
```
bboozzoo: Hm noticed that qemu 2.9 is in edge already, we might not need this change after all.

meta-mender-qemu/conf/machine/vexpress-qemu-flash.conf
```diff
@@ -22,3 +22,15 @@ UBINIZE_ARGS = "-p 256KiB -m 1 -O 64"
...
+
+# 16MB for data partition
+MENDER_DATA_PART_SIZE_MB = "16"
+# combined NOR flash is 128MB
+MENDER_STORAGE_TOTAL_SIZE_MB = "128"
+# ??
+MEDNDER_BOOT_PART_SIZE_MB = "1"
+# align to PEB size 256k
+MENDER_PARTITION_ALIGNMENT_KB = "256"
+#MENDER_STORAGE_RESERVED_RAW_SPACE = "524288"
```
bboozzoo: removed

meta-mender-qemu/conf/machine/vexpress-qemu-flash.conf
kacf: Ok, let's leave it like this for now then.

bboozzoo:
@GregorioDiStefano updated dockerfile, entrypoint and mender-qemu. Entryponing and mender-qemu are a bit more flexible now.

Also, reconfiguration is not supported for flash images, so the script will error out.

kacf:
> `Changelog: message`

Did you want to include the whole message? Use `Changelog: Commit`.

bboozzoo:
updated

meta-mender-qemu/recipes-core/systemd/files/01-logs-to-console.conf
```diff
@@ -0,0 +1,2 @@
+[Journal]
+ForwardToConsole=true
```
kacf: Won't this be unbearably verbose for a demo environment? I think Mender logs every few seconds when it checks the server.

I can see the value of having this in the tests though, maybe we can do this in the client initialization in the test framework?

bboozzoo: Ok, let's drop it for now an figure out a way to add it the same way we modify tokens/server url and such.

bboozzoo:
systemd patch was dropped, waiting for tests to complete before merging

bboozzoo:
merge?

meta-mender-qemu/Dockerfile
```diff
@@ -27,8 +27,8 @@ RUN apk update && apk upgrade && \
 ARG VEXPRESS_IMAGE=scripts/docker/empty-file
 ARG UBOOT_ELF=scripts/docker/empty-file
 
-ADD $UBOOT_ELF ./u-boot.elf
-ADD $VEXPRESS_IMAGE ./core-image-full-cmdline-vexpress-qemu.sdimg
+COPY $UBOOT_ELF ./u-boot.elf
+COPY $VEXPRESS_IMAGE .
```
kacf: I think this includes the entire directory, doesn't it? Maybe `core-image-full-cmdline-vexpress-qemu.*` would be preferable? If that's valid syntax, that is.

bboozzoo: no, the line copies `$VEXPRESS_IMAGE` to `.` under the same name as it is now

kacf: Doh, of course. I completely misread that.

bboozzoo:
thanks

PR #315, opened 2017-06-16
add option to specify ArtifactVerifyKey in mender config script

GregorioDiStefano:
Changelog: none

### DIFF
--- a/meta-mender-qemu/Dockerfile
+++ b/meta-mender-qemu/Dockerfile
@@ -10,7 +10,7 @@
 # -v $BUILDDIR:/mnt/build:ro
 #       Use BUILDDIR from a poky build as image input.
 # -v <config-dir>:/mnt/config:ro
-#       Use server.crt from config-dir, if it exists.
+#       Use server.crt and/or artifact-verify-key.pem from config-dir, if it exists.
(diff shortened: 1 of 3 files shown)


GregorioDiStefano:
@kacf

meta-mender-qemu/Dockerfile
```diff
@@ -11,6 +11,8 @@
 #       Use BUILDDIR from a poky build as image input.
 # -v <config-dir>:/mnt/config:ro
 #       Use server.crt from config-dir, if it exists.
+# -v <file>:/mnt/config/artifact-verify-key.pem:ro
```
kacf: Can you mount a single file? I have to admit that I haven't tried, I was assuming the folder would be mounted and the file would just be in there. In that case it would share mount point with `server.crt`, but there should be a comment about the verify key, of course.

GregorioDiStefano: Yes, you can

GregorioDiStefano: One second!...

GregorioDiStefano: Nope, only possible in docker-compose, damn it.

PR #328, opened 2017-06-28
qemu raw flash acceptance tests

bboozzoo:
Currently contains also what's already up for review in #323 will rebase when that one is merged.

Work in progress, I'm testing with both `vexpress-qemu` and `vexpress-qemu-flash`. You will need a modification to mednersoftware/mender-qa scripts in order to build for `vexpress-qemu-flash` as the tests don't set `MACHINE` on their own.

Most of `test_build` tests are skipped, because they rely heavily on ext* images. All of mender-artifact tests are all usable. `test_rootfs` contains a separate tests for `ubimg` and `sdimg`. All of `test_sdimg` are skipped for non-sdimg images. `test_update` was not updated yet.

@kacf @GregorioDiStefano @pasinskim

### DIFF
--- a/meta-mender-core/classes/mender-install.bbclass
+++ b/meta-mender-core/classes/mender-install.bbclass
@@ -54,7 +54,7 @@ MENDER_DATA_PART_DIR ?= ""
 MENDER_DATA_PART_SIZE_MB ?= "128"
 
 # Size of the first (FAT) partition, that contains the bootloader
-MENDER_BOOT_PART_SIZE_MB ?= "16"
+MENDER_BOOT_PART_SIZE_MB ??= "16"
 
 # For performance reasons, we try to align the partitions to the SD
 # card's erase block. It is impossible to know this information with
--- a/meta-mender-core/classes/mender-install.bbclass
+++ b/meta-mender-core/classes/mender-install.bbclass
@@ -120,3 +120,8 @@ MENDER_CALC_ROOTFS_SIZE = "${@mender_calculate_rootfs_size_kb(${MENDER_STORAGE_T
 # But subtract IMAGE_ROOTFS_EXTRA_SPACE, since it will be added automatically
 # in later bitbake calculations.
 IMAGE_ROOTFS_SIZE ?= "${@eval('${MENDER_CALC_ROOTFS_SIZE} - (${IMAGE_ROOTFS_EXTRA_SPACE})')}"
+
+# Set hard limit on maximum rootfs size. Calculated rootfs size is used when
+# partitioning the disk image (be it SD card or UBI image), and defines an upper
+# bound of the space allocated for rootfs partition/volume.
+IMAGE_ROOTFS_MAXSIZE ?= "${MENDER_CALC_ROOTFS_SIZE}"
--- a/meta-mender-qemu/recipes-core/systemd/files/01-logs-to-console.conf
+++ b/meta-mender-qemu/recipes-core/systemd/files/01-logs-to-console.conf
@@ -1,0 +1,2 @@
+[Journal]
+ForwardToConsole=true
--- a/tests/acceptance/common.py
+++ b/tests/acceptance/common.py
@@ -38,24 +38,25 @@ def if_not_bbb(func):
             func()
     return func_wrapper
 
+def start_qemu(qenv=None):
+    """Start qemu and return a subprocess.Popen object corresponding to a running
+    qemu process. `qenv` is a dict of environment variables that will be added
+    to `subprocess.Popen(..,env=)`.
 
-# Return Popen object
-def start_qemu(latest_sdimg):
-    if pytest.config.getoption("bbb"):
-        return
+    Once qemu is stated, a connection over ssh will attempted, so the returned
+    process is actually a qemu instance with fully booted guest os.
 
-    fh, img_path = tempfile.mkstemp(suffix=".sdimg", prefix="test-image")
-    # don't need an open fd to temp file
-    os.close(fh)
+    The helper uses `meta-mender-qemu/scripts/mender-qemu` to start qemu, thus
+    you can use `VEXPRESS_IMG`, `QEMU_DRIVE` and other environment variables to
+    override the default behavior.
 
-    # Make a disposable image.
-    shutil.copy(latest_sdimg, img_path)
-
+    """
     env = dict(os.environ)
-    # *.sdimg is supported by mender-qemu and needs no special handling
-    env["VEXPRESS_IMG"] = img_path
-    # use snapshot mode
-    proc = subprocess.Popen(["../../meta-mender-qemu/scripts/mender-qemu", "-snapshot"], env=env)
+    if qenv:
+        env.update(qenv)
+
+    proc = subprocess.Popen(["../../meta-mender-qemu/scripts/mender-qemu", "-snapshot"],
+                            env=env)
 
     try:
         # make sure we are connected.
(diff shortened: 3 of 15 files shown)


bboozzoo:
To build & test this you'll need:
- qemu version mentioned here https://github.com/mendersoftware/mender-qa/pull/110#issuecomment-313594265
- updated poky (uboot patches apply to 2017.05 which is currently in master, while meta-mender builds are still stuck with older version)

meta-mender-qemu/recipes-core/systemd/files/01-logs-to-console.conf
```diff
@@ -0,0 +1,2 @@
+[Journal]
+ForwardToConsole=true
```
kacf: I recall that we talked about this before: I think this is too verbose to include by default; this will also affect our demo. If this is something we could add to the tests only, it would be great. Maybe during test setup?

tests/acceptance/common.py
```diff
@@ -516,3 +516,24 @@ def add_to_local_conf(prepared_test_build, string):
...
+
+    """
+    mark = request.node.get_marker('only_with_image')
+    if mark is not None:
+        images = mark.args
+        current = bitbake_variables.get('IMAGE_FSTYPES', '').strip().split(' ')
+        if not any([img in current for img in images]):
+            pytest.skip('no supported filesystem in {} ' \
+                        '(supports {})'.format(', '.join(current),
+                                               ', '.join(images)))
```
kacf: Nice marker!

tests/acceptance/common.py
```diff
@@ -466,7 +466,8 @@ def prepared_test_build_base(request, bitbake_variables):
     build_dir = tempfile.mkdtemp(prefix="test-build-", dir=os.environ['BUILDDIR'])
 
     def cleanup_test_build():
-        run_verbose("rm -rf %s" % build_dir)
+        if not pytest.config.getoption('--keep-build-dir'):
+            run_verbose("rm -rf %s" % build_dir)
```
kacf: Definitely +1!

I was always just too much in a hurry to add it myself, but very useful.

tests/acceptance/test_mender-artifact.py
```diff
@@ -247,4 +247,10 @@ def test_artifacts_rootfs_size(self, versioned_mender_image, bitbake_path, bitba
...
         size_from_artifact = int(match.group(1))
         size_from_build = int(bitbake_variables["MENDER_CALC_ROOTFS_SIZE"]) * 1024
-        assert(size_from_artifact == size_from_build)
+        # some filesystems (eg. ubifs) may use compression or empty space may
+        # not be a part of the image, in which case the image will be smaller
+        # or equal to MENDER_CALC_ROOTFS_SIZE
+        assert size_from_artifact <= size_from_build
+        # assume that the difference will not be more than 30% of calculated
+        # size
+        assert size_from_artifact >= 0.7 * size_from_build
```
kacf: We've actually had quite a number of "off-by-one" errors in this area, so I would argue that the exact match is pretty important for the filesystem types where it's possible to test this. Can we make this dependent on which filesystem type is being tested?

tests/acceptance/common.py
```diff
@@ -228,8 +228,8 @@ def determine_active_passive_part(bitbake_variables):
     elif mount_output.find(b) >= 0:
         return (b, a)
     else:
-        raise Exception("Could not determine active partition. Mount output: %s"
-                        % mount_output)
+        raise Exception("Could not determine active partition. Mount output:\n {}" \
+                        "\nwas looking for {}".format(mount_output, (a, b)))
```
kacf: I see this a lot. Is there a particular reason for using `format()` over `%`?

meta-mender-core/classes/mender-install.bbclass
```diff
@@ -120,3 +120,8 @@ MENDER_CALC_ROOTFS_SIZE = "${@mender_calculate_rootfs_size_kb(${MENDER_STORAGE_T
 # But subtract IMAGE_ROOTFS_EXTRA_SPACE, since it will be added automatically
 # in later bitbake calculations.
 IMAGE_ROOTFS_SIZE ?= "${@eval('${MENDER_CALC_ROOTFS_SIZE} - (${IMAGE_ROOTFS_EXTRA_SPACE})')}"
+
+# Set hard limit on maximum rootfs size. Calculated rootfs size is used when
+# partitioning the disk image (be it SD card or UBI image), and defines an upper
+# bound of the space allocated for rootfs partition/volume.
+IMAGE_ROOTFS_MAXSIZE ?= "${MENDER_CALC_ROOTFS_SIZE}"
```
kacf: Where is this used? AFAICS this is not a standard poky variable.

tests/acceptance/test_build.py
```diff
@@ -94,7 +94,7 @@ def test_image_rootfs_extra_space(self, prepared_test_build, bitbake_variables):
         """Test that setting IMAGE_ROOTFS_EXTRA_SPACE to arbitrary values does
         not break the build."""
 
-        add_to_local_conf(prepared_test_build, 'IMAGE_EXTRA_ROOTFS_SPACE_append = " + 640 - 222 + 900"')
+        add_to_local_conf(prepared_test_build, 'IMAGE_ROOTFS_EXTRA_SPACE_append = " + 640 - 222 + 900"')
```
kacf: Haha, really? So all this time we've been testing nothing!

tests/acceptance/3rdparty/mendertesting/check_commits.sh
```diff
@@ -20,11 +20,11 @@ fi
 if [ -n "$1" ]
 then
     echo "Checking range: $@:"
-    git log "$@"
+    git --no-pager log "$@"
     commits="$(git rev-list --no-merges "$@")"
 else
     echo "Checking range: ${COMMIT_RANGE}:"
-    git log "$COMMIT_RANGE"
+    git --no-pager log "$COMMIT_RANGE"
```
kacf: This should be fixed upstream. Can you submit the same thing to https://github.com/mendersoftware/mendertesting ?

bboozzoo: On it, https://github.com/mendersoftware/mendertesting/pull/20

meta-mender-core/classes/mender-install.bbclass
bboozzoo: It's not really documented (I suppose an upstream patch would be good), but it's a part of image.bbclass https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/classes/image.bbclass#n512

tests/acceptance/common.py
bboozzoo: IIRC `.format()` is supposed to be a preferred way of string formatting. There are some nice tricks you can do with format that you cannot do with `%`.

meta-mender-qemu/recipes-core/systemd/files/01-logs-to-console.conf
bboozzoo: Maybe we could do reverse, have logs enabled by default and disable them with a switch in `setup-mender-configuration.py` (and a corresponding env variable in docker-compose)?

kacf: If it's not the default I don't think anyone would use it. How verbose is it though? I just fear that putting this up on a big screen during a presentation at ELC, it's gonna look bad if the screen is flooded with log messages.

bboozzoo: It's not that verbose (subjective isn't it?), i'll try to record something with asciinema

bboozzoo: Here: https://asciinema.org/a/71qu9JDI21yqOapq1INsY2Duz

Now, because the state changes are logged directly to console, they will also appear in docker logs during integration tests (super useful). For acceptance, the tests that use qemu disable mender agent service, so no logs appear.

The only thing that slightly annoys me is that journald will replay dmesg when it starts. I could not find a way to stop this behavior.

I would propose that we keep this as a default and have an env switch that can be applied when you run a docker container. For local development you can just comment out the whole append or change `ForwardToConsole` to false.

tests/acceptance/test_mender-artifact.py
bboozzoo: i'll try to split these tests into image specific ones

bboozzoo: Should be fixed now, ext[234] are checked against calculated rootfs size, ubifs size is verified to be within a sane range.

bboozzoo:
adding @oleorhagen

PR #331, opened 2017-07-05
move systemd bbappend to meta-mender-demo

mirzak:
Bringing in meta-mender-core should make minimal changes to features
that are considered distribution features. In this case we should not
try and enforce usage of networkd just because Mender is used, since it
does not really have a hard dependency on it and this is better suited
for the meta-mender-demo layer.

Changelog: None


bboozzoo:
Looks good to me (we actually have no real dependency on systemd anyway).

Don't know about @kacf though. IIRC I brought this topic up before, but we still ended up keeping systemd  pieces in meta-mender.

mirzak:
Yeah, beside the systemd service for Mender and that "depends" on systemd for a smooth out of the box experience.

The parts that this patch moves are related to network configuration. 

I recently worked in a project where we used meta-mender-core and I had to add an bbappend that removed networked and resolved from systemd packageconf because meta-mender-core brought them in, but I did not want them. Which felt like an unnecessary step for me.

bboozzoo:
FWIW I had opened #93 that was adding SysV init support, see https://github.com/mendersoftware/meta-mender/pull/93/commits/0ed16200d8c39c524bb0f531e28ac4fd45804db9 for details.

kacf:
This actually became build number 666 in Jenkins. Bad sign? :-)

kacf:
Thanks @mirzak!

PR #332, opened 2017-07-05
raspberrypi: use bootargs passed on by boot firmware

mirzak:
Currently we completly ignore what boot firmware feels that we
should use as bootargs and we define our own. By doing so we
lose some functionality. This patch attempts to improve this.

The bootargs looks like this now:

8250.nr_uarts=0 bcm2708_fb.fbwidth=1184 bcm2708_fb.fbheight=624 bcm2708_fb.fbswap=1 vc_mem.mem_base=0x3dc00000 vc_mem.mem_size=0x3f000000 dwc_otg.lpm_enable=0 console=ttyS0,115200 rootfstype=ext4
rootwait root=/dev/mmcblk0p3

Changelog: Title


kacf:
Thanks @mirzak!

PR #333, opened 2017-07-10
MEN-1071: Adds a tenant-token variable for use with custom tokens.

oleorhagen:
Also adds a simple test for testing the tenant-token new variable.

Changelog: title

### DIFF
--- a/meta-mender-qemu/recipes-connectivity/openssh/openssh_%.bbappend
+++ b/meta-mender-qemu/recipes-connectivity/openssh/openssh_%.bbappend
@@ -1,0 +1,5 @@
+
+do_install_append () {
+	install -d ${D}${systemd_unitdir}/system/network.target.wants
+  ln -s ../sshdgenkeys.service ${D}${systemd_unitdir}/system/network.target.wants/
+}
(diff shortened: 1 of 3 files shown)


oleorhagen:
@kacf

tests/acceptance/test_build.py
```diff
@@ -101,6 +101,28 @@ def test_image_rootfs_extra_space(self, prepared_test_build, bitbake_variables):
...
+        run_bitbake(prepared_test_build)
+
+        built_rootfs = latest_build_artifact(prepared_test_build['build_dir'], ".ext[234]")
+
+        subprocess.check_call(["debugfs", "-R",
+                                   "dump -p /etc/mender/mender.conf mender.conf", built_rootfs])
+
+        try:
+            output = subprocess.check_output(["debugfs", "-R", "cat mender.conf", built_rootfs])
+            assert(output.find("authtentoken"))
```
kacf: Did this pass for you? The first `debugfs` call should put the `mender.conf` file on the local file system, so the proper next call would be `output = subprocess.check_output(["cat", "mender.conf"])`.

But I think I instead would restructure this a bit, since it's easy to validate it as JSON. My suggestion:

```
try:
    with open("mender.conf") as fd:
        data = json.load(fd)
    assert data['TenantToken'] == 'authtentoken'
```

PR #338, opened 2017-07-14
mender: get rid of IMAGE_DEPENDS

drewmoseley:
Replaced usage of IMAGE_DEPENDS_<type> with explicitly added
dependencies to do_image_<type> tasks.

This is related to upstream Yocto changes.  For reference:
    https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?id=f93d58378fb4095a18c1403de03ca0e13ce465cd
    https://bugzilla.yoctoproject.org/show_bug.cgi?id=11302


drewmoseley:
Let's just use the version over #341 instead

PR #340, opened 2017-07-17
Update tests to check for version file checksum.

pasinskim:
Now manifest file contains the checksum of the version file
and tests need to be updated.

Changelog: None


pasinskim:
This is to update tests after changing the implementation here https://github.com/mendersoftware/mender-artifact/pull/80

PR #341, opened 2017-07-24
A few minor changes for master.

drewmoseley:


drewmoseley:
I think @mirzak should review as well, especially for the RPI specifics.

Note that this also is needed against pyro and I'll resubmit there once it has been reviewed here.

This also contains the contents from #338 so if that one is reviewed first this will be simplified to just the one commit.

drewmoseley:
Let's hold off on merging.  I've got one more minor commit coming that adapts the RPi UBoot patch to apply cleanly against the current master sources.

drewmoseley:
OK.  Patch adjustment patch submitted.  As far as I can tell this is ready for merge.  I think we are still seeing CI failures but as I recall those are for issues unrelated to this specific merge request.

GregorioDiStefano:
@drewmoseley you forgot to signoff

drewmoseley:
@GregorioDiStefano thanks for the reminder.  These commits have been updated with Changelog entries.

PR #343, opened 2017-07-31
Toradex Colibri iMX7 support - WIP

bboozzoo:
This is a WIP PR adding Colibri iMX7 support and a setup for integrating with CI pipieline.

The first part of the PR is basically adding mender integration, some cleanups etc.

There's quite a bit of problems with current setup. First, meta-toradex-nxp supposedly does not support master. The most recent release that is mentioned is morty. I ran into some problems building u-boot-toradex (these were solved rather easily) and u-boot-toradex-fsl-fw-utils. Especially the latter recipe was a pain in the ass, as it's missing some patches that used to be applied by Yocto. I have these changes in my local tree.

The second part is adding `colibri-imx7-mender` machine configuration. The configuration is based on `colibri-imx7` as defined in mender-toradex-nxp. The idea is that CI builds use `colibri-imx7-mender` machine as it sets up mender storage variables and provides smaller updates to recipes (may change in future).

@kacf @pasinskim @drewmoseley @GregorioDiStefano


tests/meta-mender-toradex-nxp-ci/README
```diff
@@ -0,0 +1,64 @@
...
+  branch: master
+
+  URI: git://git.openembedded.org/openembedded-core
+  layers: meta
+  branch: master
+
+  URI: git://git.yoctoproject.org/xxxx
+  layers: xxxx
+  branch: master
+
```
drewmoseley: Doesn't this also depend on meta-toradex-nxp?

tests/meta-mender-toradex-nxp-ci/README
```diff
@@ -0,0 +1,64 @@
...
+=================
+
+  I. Adding the meta-mender-toradex-ci layer to your build
+ II. Misc
+
+
+I. Adding the meta-mender-toradex-ci layer to your build
+=================================================
+
+--- replace with specific instructions for the meta-mender-toradex-ci layer ---
```
drewmoseley: I guess this comment should be removed.

tests/meta-mender-toradex-nxp-ci/README
```diff
@@ -0,0 +1,64 @@
+This README file contains information on the contents of the
```
drewmoseley: Why is this in the tests directory?

bboozzoo: Because it will be used for testing only. I don't see any other reasonable method of setting mender storage variables than in machine config. This layer was crated specifically to define `colibri-imx7-mender` machine config and isolate it from `meta-mender-toradex-nxp`.

The `README` is obviously a template added `yocto-layer` which I have not updated yet.

bboozzoo:
review anyone?

tests/meta-mender-toradex-nxp-ci/recipes-bsp/u-boot/files/colibri-imx7/fw_env.config
```diff
@@ -0,0 +1,10 @@
+# Configuration file for fw_(printenv/setenv) utility.
+# Up to two entries are valid, in this case the redundant
+# environment sector is assumed present.
+# Notice, that the "Number of sectors" is not required on NOR and SPI-dataflash.
+# Futhermore, if the Flash sector size is ommitted, this value is assumed to
+# be the same as the Environment size, which is valid for NOR and SPI-dataflash
+
+# MTD device name       Device offset   Env. size       Flash sector size       Number of sectors
+/dev/mtd3               0x00000000      0x00020000      0x20000                 4
+/dev/mtd3               0x00020000      0x00020000      0x20000                 4
```
mirzak: Above is not completely correct. 

The field "Number of sectors" indicates how many NAND sectors to occupy for the U-boot ENV, in the above case 0x80000 (512 kB) will be used for each part. This must match the CONFIG_ENV_RANGE variable in the U-boot conf (which I believe is also setup to 512 kB in colibri-imx7.h).  CONFIG_ENV_RANGE is used so that it can tolerate bad blocks in the ENV area.

Above will result in that the "second" environment will be overwritten in case of bad blocks on the "first" environment part.

One possible solution is to increase the size of the u-boot-env MTD part (1MB) and change the device offset of the second env area to 0x80000

or

decrease the number of sectors to 2 (256 kB) both in fw_env.config and in u-boot conf, but you still need to adjust the device offset for the second area but you then do not change the size of u-boot-env MTD part.

or 

one can omit the "Number of sectors" field and CONFIG_ENV_RANGE from u-boot, then it will only occupy CONFIG_ENV_SIZE.

bboozzoo: Thanks for spotting this. Colibri iMX7 is set up with ENV_SIZE == ENV_SECT_SIZE (128kB). I believe ENV_RANGE will default to ENV_SIZE if not defined (and it's not defined for mx7 either). Redundant copy is currently set up to start right after the main one.

I wouldn't want to change the partitioning just for Mender. I suppose we could push back the redundant environment by another 128kB, so it would start at 0x3c0000 (or 0x40000 in /dev/mtd3), and set number of sectors to 2. ENV_SIZE remains unchanged, ENV_RANGE becomes 2*ENV_SIZE (to be added to clibri_imx7.h). 

With the mender view on things, things get a little complicated though. The setup code assumes that it knows better and sets up the redundant environment right after the main one, at MENDER_BOOTENV_SIZE distance (128kB). In theory I could manipulate MENDER_BOOTENV_SIZE to be 256kB, this would give the necessary space for ENV_RANGE, and just have ENV_SIZE set to 128kB. However, then there's a check for MENDER_BOOTENV_SIZE == ENV_SIZE, which will not be true anymore.

bboozzoo: @mirzak can you double check something for me?

 I'm looking at vf support, the BOOTENV_SIZE is 0x20000, ENV_OFFSET is 0x180000 (I suppose you've set MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET to the same value when building locally), ENV_RANGE is 0x80000 (4 * ENV_SIZE). VF eraseblock is 128kB, so I'm guessing you've set MENDER_PARTITION_ALIGNMENT_KB to 128 as well.

The generated `config_mender_defines.h` will define:
```
#define MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET_1 0x180000 /* CONFIG_ENV_OFFSET */
#define MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET_2 0x1a0000 /* ??? */
```
Then the 'generic mender bootcode' patch will set:
```
+#define CONFIG_ENV_OFFSET        MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET_1
+#define CONFIG_ENV_OFFSET_REDUND MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET_2
```
The offset of redundant environment will be 0x1a0000 (0x20000 wrt. /dev/mtd2 or u-boot-env start), while it should be 0x200000 (0x80000 wrt. /dev/mtd2, as defined in `fw_env.config`)

bboozzoo:
@kacf @drewmoseley  https://github.com/mendersoftware/meta-mender/pull/343/commits/d22be6b63408052b4e2d20162a694761334d4d7a we'll probably need to cherry-pick this to master once you review, otherwise it's not possible to have a working image with U-Boot using CONFIG_ENV_RANGE set to anything else than CONFIG_ENV_SIZE

tests/meta-mender-toradex-nxp-ci/recipes-bsp/u-boot/files/colibri-imx7/fw_env.config
kacf: > With the mender view on things, things get a little complicated though. The setup code assumes that it knows better and sets up the redundant environment right after the main one, at MENDER_BOOTENV_SIZE distance (128kB). In theory I could manipulate MENDER_BOOTENV_SIZE to be 256kB, this would give the necessary space for ENV_RANGE, and just have ENV_SIZE set to 128kB. However, then there's a check for MENDER_BOOTENV_SIZE == ENV_SIZE, which will not be true anymore.

Then I think we should take `CONFIG_ENV_RANGE` into account in the calculations. Basically do the same for it as for `ENV_SIZE`, make a `MENDER_ENV_RANGE` and make sure it passes the equality check, and use it to calculate the offset at which to place the environment. Surely this is not the only board that will be affected by `CONFIG_ENV_RANGE`.

tests/meta-mender-toradex-nxp-ci/conf/machine/colibri-imx7-mender.conf
```diff
@@ -41,6 +41,12 @@ IMAGE_ROOTFS_MAXSIZE = "307200"
 # meta-mender-core/recipes-bsp/u-boot/u-boot-fw-utils-mender.inc we need ship
 # our own config instead.
 MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET = "0x380000"
+# board configuration is set up with CONFIG_ENV_RANGE = 2 * CONFIG_ENV_SIZE,
+# hence we need to override the offset of redundant environment otherwise the
+# default value (as calculated by u-boot-mender-common.inc) would end up inside
+# the backup space of the main environment
+MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET_1 = "${MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET}"
+MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET_2 = "0x3c0000"
```
kacf: I think this part should be unnecessary if we take `CONFIG_ENV_RANGE` into account in our calculations, as per my other comment.

bboozzoo: Basically, what you're suggesting is:
* add MENDER_BOOTENV_RANGE, defaults to unset
* update the integration patch to define CONFIG_ENV_RANGE iff MENDER_BOOTENV_RANGE is set
* verify that MENDER_BOOTENV_RANGE is a multiple of MENDER_BOOTENV_SIZE
* update `meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc` to account for MENDER_BOOTENV_RANGE when calculating the offset of redundant environment
* add check that the environments do not overlap

is that right? Looks doable.

OTOH, I would still keep the weak assignment to `MENDER_UBOOT_ENV_STORAGE_OFFSET_{1,2}` (+ a check that environments don't overlap). Should the user decide to override the offsets manually they surely knows better where they want the environment to be placed. Mender integration code should probably not interfere with this.

kacf: > is that right? Looks doable.

Yes, exactly!

> OTOH, I would still keep the weak assignment to MENDER_UBOOT_ENV_STORAGE_OFFSET_{1,2} (+ a check that environments don't overlap). Should the user decide to override the offsets manually they surely knows better where they want the environment to be placed. Mender integration code should probably not interfere with this.

Yes, absolutely. Make the automagic work 80% of the time, and then make it easily overridable when it's being stupid! :-)

bboozzoo:
Pushed 2 commits adding BOOTENV_RANGE setting.

kacf:
Everything is passing here, so merge at will!

PR #353, opened 2017-08-04
Fix parse error when PREFERRED_VERSION is unset

drewmoseley:
Changelog: Title


drewmoseley:
Per internal discussions with @bboozzoo and @GregorioDiStefano there seem to be some other issues here.  We may need to rethink the mender_preferred_version_from_* functions.

kacf:
Absolutely, thanks!

I hate the whole recipe, to be honest. It's way too complex. But I have not found another way to support all at once:

* Building version tags from version specific recipes (`mender_1.2.0.bb`)
* Building temporary build tags from the git recipe (`mender_git.bb`)

The autodetection seems necessary to accomplish this. I'm all for simplifying this though, if anybody has an idea how.

PR #354, opened 2017-08-08
meta-mender-core: do not warn about ubifs being unsupported

bboozzoo:
@estenberg @kacf @GregorioDiStefano


estenberg:
Looks good, I think we should merge after verifying tests still pass.

drewmoseley:
Looks good to me.  I think we should definitely cherry-pick this back to pyro since that is being used as the Yocto release for Mender 1.2

estenberg:
I cherry-picked to pyro in 56f655ff73efda0ccd5ccf23893eaf7c2f6f10d9.

There is a side-effect that Mender 1.1 also will have this warning removed as it is based on pyro, but I think that is a small issue so we can ignore it.

PR #356, opened 2017-08-11
Toradex acceptance tests & CI setup

bboozzoo:
This PR builds on top of #343 with some significant additions:

#### CI layers

Under `tests/` you will find Yocto layers for building CI images, for both test host (i.e. the default image that you start your tests from) and test image (the one that's undergoing QA). The layers are:

* meta-mender-ci

Base CI layer that defines `poky-mender-ci` distro, which is basically Poky + system - random unused features + ipk packages. You can/should use this distro bo build images for both test host and QA. 

The layer also defines `mender-image-test-host` image, which is `core-image-full-cmdline` + debug tweaks + package management (so that you could make use of the ipk repos), OpenSSH (sshd and ssh client) + rsync + image partitioning tools (parted, mkfs.ext[234], mtd-utils) + `mender-qa`.

`mender-qa` is a recipe & a package that is supposed to provide some basic functionality for deploying the test image and activating it (i.e. so that after reboot we use that image). There's some default implementation for respective subcommands, and board CI layers are supposed to provide the actual board-specific code. The package needs to be installed in QA images too to provide a method of switching back to the test host image.

On top of this, there's a bbappend for system to enable DHCP on eth* by default.

This layer should be included when building QA images along with board specific CI layers *(some exceptions may apply)*

* meta-mender-toradex-nxp-ci

This is a Toradex NXP specific layer. The layer comes with a patch for `u-boot-toradex` that will, by default, boot from SD card, and boot from NAND flash only if `mender_qa_run_tests=1`. It is imperative that the layer is included when building for both test host and QA images (otherwise u-boot-fw-utils may get out of sync wrt. default configs).

The layer also provides an implementation of `deploy-test-image` that basically installs new u-boot, image (ubimg) and wipes u-boot-env (hence u-boot default env must be in sync)

Maybe it would be better to rework u-boot to also load `uEnv.txt` from SD card like it's done for BBB. This would remove the tight coupling, but I have not tried to do so yet. I suggest we start what's right now and see whether it turns out to be inconvenient.

* meta-mender-beaglebone-ci

Beaglebone specific layer. I have not worked with this too much, but got the basic image deploy & switch functionality working. There's an implementation for `deploy-test-image` and `activate-test-image` commands. Image switching uses `uenvcmd` like it does in current setup, but with a minor change in partitions (since we're running a test host image and not the Debian installation that came with BBB).

There's also a bbappend for packagegroup-core-boot which pulls in mender-qa. IMO this should me moved to `meta-mender-ci` so that when including the base CI layer you'd have mender-qa installed by default.

#### tests

Tests were updated where possible. I've tried to add a common board setup fixture that is controlled by `--board-type` command line switch. Also, since I don't have access to your S3, I have used minio and thus added more switches that add relevant support.

@GregorioDiStefano @kacf @pasinskim


tests/meta-mender-beaglebone-ci/COPYING.MIT
```diff
@@ -0,0 +1,17 @@
...
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
```
kacf: Should use Apache 2.0 license here, our default license.

tests/meta-mender-beaglebone-ci/README
```diff
@@ -26,3 +26,41 @@ General
...
+
+* Use mender-image-test-host to build a test host image
+
+* Test host image is running from eMMC (/dev/mmcblk1 aka `mmc dev 1` under U-Boot)
+
+* When built, just dd the image onto the card, boot, once in the running system do:
+
+     dd if=/dev/mmcblk0 of=/dev/mmcblk1 bs=1M
+
+* Set MACHINE = "beaglebone-mender" device when building for test host.
```
kacf: Why is it not enough to use the CI distro?

tests/acceptance/common.py
```diff
@@ -299,7 +299,76 @@ def boot_from_internal():
         append("/uboot/uEnv.txt", bootline)
         reboot()
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="module")
```
kacf: You can probably use `scope="session"` here, no? This should be valid for the whole run.

tests/acceptance/common.py
```diff
@@ -299,7 +299,76 @@ def boot_from_internal():
...
+
+    print('board type:', bt)
+    if bt == "qemu":
+        return qemu_running(request, clean_image)
+    elif bt == "bbb":
+        return setup_bbb(request)
+    elif bt == "colibri-imx7":
+        return setup_colibri_imx7(request, clean_image)
+    else:
+        pytest.skip('board type not configured')
```
kacf: I would use `fail` here. If board type is somehow misconfigured in Jenkins, I'd rather know about it than get a silent skip. If the test is deselected I don't think the fixture will run anyway, so it should not be a problem for local testing.

tests/acceptance/common.py
```diff
@@ -299,7 +299,76 @@ def boot_from_internal():
...
+    run("mender-qa activate-test-image")
+
+    reboot()
+
+def common_board_cleanup():
+    run("mender-qa activate-test-image off")
+    reboot()
+
+
+@pytest.fixture(scope='module')
```
kacf: Again, can be `scope='session'`.

tests/acceptance/common.py
```diff
@@ -299,7 +299,76 @@ def boot_from_internal():
...
+
+    def board_cleanup():
+        def board_cleanup_impl():
+            common_board_cleanup()
+        execute(board_cleanup_impl, hosts=conftest.current_hosts())
+
+    request.addfinalizer(board_cleanup)
+
+
+@pytest.fixture(scope="module")
```
kacf: Again, can be `scope='session'`.

tests/acceptance/common.py
bboozzoo: fixed

tests/acceptance/common.py
bboozzoo: fixed

tests/acceptance/common.py
bboozzoo: fixed

tests/acceptance/common.py
bboozzoo: fixed

tests/meta-mender-beaglebone-ci/COPYING.MIT
bboozzoo: I suppose we don't need a license file here as there's one at the top already.

tests/meta-mender-beaglebone-ci/README
bboozzoo: We're not setting any machine specific configuration for BBB yet, but it's already being done for Toradex. I've added `beaglebone-mender` just to have a placeholder for any such settings. For instance, should we change the testing to use internal eMMC rather than the SD card (which probably makes more sense for real life application), the related settings would go to `beaglebone-mender.conf`.

kacf:
Everything is passing here as well, merge when you're ready!

bboozzoo:
Rebased on top of master. Waiting for the checks to complete and will merge.

bboozzoo:
merging

PR #361, opened 2017-08-15
Rewrite mender.conf generation in Python.

drewmoseley:
This has several advantages:

* It allows settings to be merged with a SRC_URI provided mender.conf.

* It's much more tolerant towards missing variables.

* It doesn't need a template file with exactly matched variable
  strings in it.

Changelog: It's now possible to provide a mender.conf file in SRC_URI,
and this will be merged with settings taken from bitbake variables,
with the bitbake variables taking precedence.

Modified-by: Drew Moseley <drew.moseley@northern.tech>
* Added second parameter to getVar() calls as the Morty branch still
requires it.

(cherry picked from commit 49bdabf199f8256b7ce3eefc3d23990085044e25)


drewmoseley:
@kacf please take a look.  I had to modify the original patch to work with Morty.  Due to https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?id=1fce7ecbbb004 and related patches I had to explicitly add the second parameter to the getVar() function calls but otherwise it's the same as in master.

PR #367, opened 2017-09-01
Add all new Mender 1.1.1 and 1.2.0 components to recipes.

kacf:
Changelog: Title


drewmoseley:
@kacf should this also remove the beta versions of the recipes?

kacf:
Yes, I think so. I was considering leaving them in just to keep them around for next time, but we can always get them from history.

PR #371, opened 2017-09-06
Remove godep from meta-mender layer.

kacf:
We have not used it in a long time, and the version is ancient.

Changelog: Title


bboozzoo:
merge?

PR #372, opened 2017-09-11
MEN-1412 Fix broken header checksum verification.

pasinskim:
Add an acceptance test for checking that the update is failing
if header checksum is invalid.

Changelog: None

### DIFF
--- a/tests/acceptance/test_update.py
+++ b/tests/acceptance/test_update.py
@@ -95,6 +95,7 @@ class SignatureCase:
     key = False
     key_type = ""
     checksum_ok = True
+    header_checksum_ok = True
 
     update_written = False
     success = True
@@ -106,6 +107,7 @@ class SignatureCase:
                  key,
                  key_type,
                  checksum_ok,
+                 header_checksum_ok,
                  update_written,
                  artifact_version,
                  success):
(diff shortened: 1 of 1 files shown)


pasinskim:
This one depends on https://github.com/mendersoftware/mender-artifact/pull/95 and https://github.com/mendersoftware/mender/pull/245

Tests started here: https://mender-jenkins.mender.io/job/mender-builder/2302/

tests/acceptance/test_update.py
```diff
@@ -461,6 +488,13 @@ def test_signed_updates(self, sig_case, bitbake_path, bitbake_variables):
                                 os.remove(data_file)
                         finally:
                             os.chdir("..")
+                    if not sig_case.header_checksum_ok:
+                        # Corrupt the header checksum (the same as corrupting header itself)
```
kacf: Actually, it's not, because if you change the checksum, then the signature check fails, but this we already know. What we need to do is to keep the checksum the same, so that the signature check passes, but change the actual header, which would fail when matching the checksum against the file.

pasinskim:
@kacf build is passing now here: https://mender-jenkins.mender.io/job/mender-builder/2388

PR #373, opened 2017-09-14
Check for creation of Mender artifact file

drewmoseley:
This is a workaround for the lack of an error return code
from mender-artifact v1.0.x.


drewmoseley:
@kacf what do you think of this?  if we include this, then we don't have to update the 1.0.x branch of mender-artifact, at least for the failure scenarios I've seen.

drewmoseley:
Any idea why the CI build is failing?  It seems to be having trouble in the populate-lic phase of the mender recipe but I  cannot reproduce that locally.

kacf:
The base reason is that the checksum is wrong for master builds. But the slightly deeper reason is that it's lacking the magic to detect which version it's building. I have attempted to cherry-pick the necessary changes in #374, on top of your change. Let's see how it goes.

PR #374, opened 2017-09-18
Morty backports for Jenkins building

kacf:


kacf:
@drewmoseley: I had to add a missing changelog tag for your initial commit, now the build is fine, but it fails in acceptance testing. This will have to be investigated, but I'm pretty sure this has nothing to do with this PR. I'm not sure how much time you want to spend on it. This is already in a better state than it was before, so we could merge as is.

drewmoseley:
Looks OK to me.  Since these are all cherry picks, I didn't investigate each patch in detail.

PR #375, opened 2017-09-18
MEN-595: Organize Mender features by DISTRO_FEATURES, not classes.

kacf:
```
commit e3f2dd878302397ce25267951e9cd61b4d2ad39e
Date:   Mon Sep 18 10:21:18 2017

    MEN-595: Organize Mender features by DISTRO_FEATURES, not classes.
    
    The main benefit is being able to turn features on and off
    selectively. With classes this is not possible once the class has been
    included, but with DISTRO_FEATURES, one can disable a feature by using
    DISTRO_FEATURES_BACKFILL_CONSIDERED.
    
    Also add the convenience variables MENDER_FEATURES_ENABLE and
    MENDER_FEATURES_DISABLE, that map to DISTRO_FEATURES_BACKFILL and
    DISTRO_FEATURES_BACKFILL_CONSIDERED, respectively. The only difference
    is that you can only add Mender features to them. This has no
    functional advantage, it is only to have an easier-to-understand
    variable, since DISTRO_FEATURES_BACKFILL_CONSIDERED is so badly named.
    
    Changelog: Commit
    
commit 0193434d72dd0669a8c01cf17a062ae4bf28d976
Date:   Fri Sep 15 13:08:34 2017

    Don't autodetect sdimg fstype, use ARTIFACTIMG_FSTYPE.
    
    Changelog: None
    
commit e7e7f688d7408466043f017b61a7dca20863946e
Date:   Fri Sep 15 10:52:05 2017

    Remove automatic dependency and depend on ARTIFACTIMG_FSTYPE.
    
    It is more accurate, since this is what the generator actually uses.
    
    Changelog: None
    
```


kacf:
This is mainly in preparation for grub support, since our number of optional features is growing.

kacf:
@bboozzoo, @drewmoseley

kacf:
Documentation is in https://github.com/mendersoftware/mender-docs/pull/260.

drewmoseley:
If I'm reading this and the docs correctly, the old(current) method of inheriting mender-full will continue to work without modification correct?  Is that expected to be the case moving forward or do you anticipate obsoleting that at some point?

kacf:
> If I'm reading this and the docs correctly, the old(current) method of inheriting mender-full will continue to work without modification correct? Is that expected to be the case moving forward or do you anticipate obsoleting that at some point?

Correct. At first I was considering obsoleting it, but I think it's a nice mechanism to get the defaults. Now that features can be turned off selectively I don't think it hinders anything.

drewmoseley:
I'm fine with this.  I'd like to hear @bboozzoo's opinion.

bboozzoo:
raw flash acceptance failed because of this:
```
MMC:   MMC: 0
*** Warning - bad CRC, using default environment

In:    serial
Out:   serial
Err:   serial
Net:   smc911x-0
Hit any key to stop autoboot:  2  1  0 
Booting from NOR...
ubi0: attaching mtd3
ubi0: scanning is finished
UBI init error 12
=> No handlers could be found for logger "paramiko.transport"
qemu-system-arm: terminating on signal 15 from pid 6255 (/usr/bin/python)
```

kacf:
> raw flash acceptance failed because of this

Yep, I'll take a look! I'm surprised it even got that far on the first attempt!

kacf:
> raw flash acceptance failed because of this

So this revealed some very annoying behavior: Variable overrides (like suffixing with `_mender-ubi`) trumps all other assignments, no matter how weak the assignment is. For example:

```
MYVAR_mender-ubi ??= "wrong"
MYVAR = "right"
```

The value will be "wrong" after the above assignments, assuming that `mender-ubi` is an active override, which I find pretty surprising. This makes providing alternative defaults much less convenient. Since we can't include files conditionally, I don't see any other way to resolve this than using python snippets to set the variables after testing whether the feature is enabled. I'm open to other ideas though.

kacf:
Pushed two new patches to fix the flash tests.

PR #376, opened 2017-09-19
rpi: Fix DTB handling for Mender-based Raspberry Pi boards.

drewmoseley:
On the Raspberry Pi platform, the DTB is generate by the start.elf firmware
and we should defer to that rather than loading the "standard" DTB file.
This is fixed in the meta-raspberrypi layer here:
    https://www.mail-archive.com/yocto@yoctoproject.org/msg32465.html

Additional links with details:
    https://www.raspberrypi.org/forums//viewtopic.php?f=107&t=134018
    http://dius.com.au/2015/08/19/raspberry-pi-uboot/

### DIFF
--- a/meta-mender-raspberrypi/README.md
+++ b/meta-mender-raspberrypi/README.md
@@ -2,7 +2,7 @@
 
 This Yocto layers contains recipes which enables support of building Mender client for Raspberry Pi boards.
 
-**NOTE!**. To be able to support update of Linux kernel and DTB these are installed to `/boot` on rootfs. This breaks support of user configurations based on `config.txt` because the Raspberry Pi boot firmware requires that the DTB file is in the same partition as the boot firmware files (mmcblk0p1). Boot firmware normally patches the DTB file based on configurations in `config.txt`.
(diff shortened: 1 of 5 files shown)


mirzak:
@drewmoseley. Here are my thoughts. 

First of you missed the Changelog :).
 
I am all for the change that is happening here. But just to document some consequences (maybe we should update the README with a note as well, we probably should update the README anyway since there are some notes there that are no longer true with this change). 

The biggest consequence of this change is that the DTB is no longer updated when you update with Mender since the DTB that is inside the "boot part" is used by the start.elf firmware. It could probably be possible to update it now that we have statescripts, but until it is implemented it will not be updated. And this was the main reason why I choose not to implement it like this in the beginning, but as time has shown there has been some "dropped" features because of this that confuse the users (even fooled me :)). So it might be better not to update the DTB to gain a setup that closer to the "stock" image. But as mentioned we should document this.

Also start.elf does not generate a DTB from scratch, but it uses the DTB that resides in the "boot part" as base and does some magic on it before it is passed on to U-boot/Kernel. So we are still using a DTB that is generated by Yocto.

Some random review comment below :).

I am not a big fan of providing links in the commit message and IMO this is better suited as text in the pull request.

drewmoseley:
@mirzak thanks for the feedback.  I'll update the README and push some updates here.

Regarding the links, I would normally include them in the PR, however, I'm concerned that down the line there will be questions as to why this change was done since it is fairly non-standard in Linux and I wanted to make sure the details were visible from a checked out repo.

drewmoseley:
I updated the commit with README updates and a Changelog.

kacf:
> Regarding the links, I would normally include them in the PR, however, I'm concerned that down the line there will be questions as to why this change was done since it is fairly non-standard in Linux and I wanted to make sure the details were visible from a checked out repo.

+1 from me. It's true that they may become stale at some point, but when reading history it is much better to have them available. Text in PRs basically gets lost as soon as the PR is merged. Personally I rarely include any text in the PR that isn't already in the commit message(s).

PR #380, opened 2017-09-22
MEN-1431: Make UEFI image generator: uefiimg.

kacf:
This is the first step in booting via UEFI. This just creates an image
using the GPT partition table format, which UEFI requires. There are
some shortcomings in this implementation, which will be resolved in
upcoming tasks:

* The boot partition source in the wks file should be changed to
  'bootimg-efi'. However, this can't be done until we have a valid
  bootloader available like grub. Hence the EFI partition is not valid
  at the moment.

* Right now you have to set IMAGE_BOOT_FILES to a dummy file in order
  for wic not to complain. This will also go away once we get a real
  EFI bootloader in place.

We also tweak some variable assignments after it was revealed that
removing mender-uboot from DISTRO_FEATURES caused some problems.

Changelog: title

### DIFF
--- a/meta-mender-core/classes/mender-artifactimg.bbclass
+++ b/meta-mender-core/classes/mender-artifactimg.bbclass
@@ -10,7 +10,9 @@ MENDER_ARTIFACT_SIGNING_KEY ?= ""
 
 do_image_mender[depends] += "mender-artifact-native:do_populate_sysroot"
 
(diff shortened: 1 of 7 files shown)


kacf:
Merging, since the previous build passed, latest update was just tweaking commit order, and this fixes a regression currently in the meta-mender layer.

PR #381, opened 2017-09-23
Setup image dependencies for SDIMG on Raspberry Pi platforms

drewmoseley:

### DIFF
--- a/meta-mender-core/classes/mender-part-images.bbclass
+++ b/meta-mender-core/classes/mender-part-images.bbclass
@@ -117,6 +117,7 @@ addtask do_rootfs_wicenv after do_image before do_image_uefiimg
 
 do_image_sdimg[depends] += "${@d.getVarFlag('do_image_wic', 'depends', False)} wic-tools:do_populate_sysroot dosfstools-native:do_populate_sysroot mtools-native:do_populate_sysroot rsync-native:do_populate_sysroot"
 do_image_uefiimg[depends] += "${@d.getVarFlag('do_image_wic', 'depends', False)} wic-tools:do_populate_sysroot dosfstools-native:do_populate_sysroot mtools-native:do_populate_sysroot rsync-native:do_populate_sysroot"
+do_image_sdimg[depends] += " ${@bb.utils.contains('SOC_FAMILY', 'rpi', 'bcm2835-bootfiles:do_populate_sysroot', '', d)}"
 
 # This isn't actually a dependency, but a way to avoid sdimg and uefiimg
 # building simultaneously, since wic will use the same file names in both, and
--- a/meta-mender-raspberrypi/README.md
+++ b/meta-mender-raspberrypi/README.md
@@ -29,8 +29,6 @@ in addition to `meta-mender` dependencies.
         MENDER_PARTITION_ALIGNMENT_KB = "4096"
         MENDER_BOOT_PART_SIZE_MB = "40"
 
-        do_image_sdimg[depends] += " bcm2835-bootfiles:do_populate_sysroot"
-
         # raspberrypi files aligned with mender layout requirements
         IMAGE_BOOT_FILES_append = " boot.scr u-boot.bin;${SDIMG_KERNELIMAGE}"
         IMAGE_INSTALL_append = " kernel-image kernel-devicetree"


drewmoseley:
@kacf @mirzak putting this into local.conf doesn't seem to do anything.  I experimented a bit and verified that if I cleaned the bcm2835-bootfiles and rebuilt then it didn't actually get included.  Putting it in the mender-sdimg.bbclass conditionally on the SOC_FAMILY seems like the only way to handle it.

I'm testing this on both pyro and morty as well.

drewmoseley:
I wonder if the other custom settings needed for RPI could be included conditionally in a similar manner. That would cut down on the chances of user error.  @kacf what do you think?

mirzak:
Checked the meta-raspberrypi [1], they add dependencies on do_deploy and not do_populate sysroot. do_deploy makes more sense, but do not know which one is correct. Maybe this is causing the problems?

@drewmoseley, but do we need to update this on pyro and morty?

As I understand it the image depends only need to be changed on master and the old method should still work on pyro and morty, that is:

```
IMAGE_DEPENDS_sdimg += " bcm2835-bootfiles"
```

Regarding the "other" RPi settings, most of the can be dropped on master branch at least. meta-raspberrypi has introduced [2] a new config (`RPI_USE_U_BOOT`) which should be enough. That means that we do not need to set `KERNEL_IMAGETYPE` and we do need to append to `IMAGE_BOOT_FILES` since this is now taken care of the RPI machine configuration. 

[1]. https://github.com/agherzan/meta-raspberrypi/blob/master/conf/machine/include/rpi-base.inc#L100-L103
[2]. https://github.com/agherzan/meta-raspberrypi/commit/e9bb7f0c6d3f77312970034ca0a63b3df0e3ba8e#diff-c02a029d89d92b2ddeb5e523a8cd535b

drewmoseley:
@mirzak I modeled the initial change on https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?id=f93d58378fb4095a18c1403de03ca0e13ce465cd which used do_populate_sysroot.

I think you are right that the old style IMAGE_DEPENDS_* will still work on the pyro and morty branches but as far as I can tell the new still setting does work.  If we put the logic into the class then it doesn't matter if it differs between branches but if it is something the users need to add manually to their local.conf files, then I'm concerned that it becomes more error prone.

mirzak:
@drewmoseley, I ran some tests my self on pyro branch only though. 

But too me it seems to work when putting:

    do_image_sdimg[depends] += " bcm2835-bootfiles:do_populate_sysroot"

In local.conf. 

I ran

     bitbake -c cleansstate bcm2835-bootfiles

between my tries. And when removing the dependency from sdimg the bootfiles are not deployed. I wonder if something has happened on master that broke it?

But either way, regarding putting all custom configurations in bbclasses for the purpose of minimizing user error does not seem right to me. And their are quite a lot of additions that are required by meta-mender as well, so the risk for user error remains.

There are other mechanisms in Yocto that can be utilized for this, one example is providing custom TEMPLATECONF files, this way we setup up everything for the users. I utilize this a lot in other projects. 

What you end up doing is something like this:

   $ export TEMPLATECONF=${PWD}/layers/meta-mender/meta-mender-raspberrypi/conf
   $ . layers/openembedded-core/oe-init-build-env build

Then the `oe-init-build-env` script will copy the template files from our location instead of using the default ones from poky. And the files provided by us can contain all the custom options required for Mender and RPi.

I usually also setup a "repo" manifest for my projects as well. See [1] and [2]. I would gladly setup something similar for meta-mender + meta-raspberrypi

What do you think about this?

[1]. https://github.com/hostmobility/hostmobility-bsp-platform
[2]. https://github.com/Freescale/fsl-community-bsp-platform

drewmoseley:
When I tested the other day I used the output of "bitbake -g core-image-full-cmdline" to figure out if the dependency was getting set and not the actual builds.  I wonder if the "-g" stuff somehow misses the dependency in local.conf?

It does look like the dependency in local.conf does work (with either IMAGE_DEPENDS_ or  do_image_sdimg[depends]) so the urgency is off.  Interesting it seems to work even with that completely removed.  I wonder if the IMAGE_BOOT_FILES stuff in rpi-base.inc is superseding this dependency?

You may be right about not using the class file for other settings and the TEMPLATECONF stuff may well be what we should use instead.  I'll look at it as I've never used it.

But for _this_ change, I still like the idea of moving it into the image class since ultimately that is what it is modifying.

kacf:
> But for this change, I still like the idea of moving it into the image class since ultimately that is what it is modifying.

It is, but in the wrong layer. If it can stay in the raspberry layer, I think it should, so that we don't clutter the core layer with board specific stuff.

drewmoseley:
There's no way to "append" a class in a separate layer that I am aware of.  That's why I have in the main class conditional on the SOC_FAMILY variable.  Is there another mechanism I am overlooking to allow us to store this in meta-mender-raspberrypi?

mirzak:
> You may be right about not using the class file for other settings and the TEMPLATECONF stuff may well be what we should use instead. I'll look at it as I've never used it.

I started working on something. Take a look at [1] and [2] to get an idea of what I mean. 

[1]. https://github.com/mirzak/mender-raspberrypi-manifest
[2]. https://github.com/mirzak/meta-mender/commit/ba3ff2c84ae0a6307ff130af0244cdcf6a35e873

kacf:
No, I think you're right, but it doesn't need to be in that class, since the class functionality is global during evaluation. So I'm pretty sure it can be anywhere. The simplest place is in the `conf/layer.conf` file of the raspberry-pi layer (or some class file inherited from it). You can at least try it, if it doesn't work, we'll put it where you originally proposed.

drewmoseley:
I recall storing some variables in layer.conf previously.  I'll give it a shot.  I'm not sure how that is viewed by the Yocto community though.  Seems like only layer settings should be in there.

kacf:
Probably true. I'm not sure how the best practice is for layer-global settings like this. You could inherit a new class instead, which is specific to raspberry pi, just to get it a bit more separated. But we're bike shedding a bit though. If it's a lot of trouble to put it elsewhere, let's just put it in `mender-sdimg.bbclass`.

drewmoseley:
OK.  This will need to be adjusted based on https://github.com/mendersoftware/meta-mender/pull/387.  I'll rebase after that one is merged.

drewmoseley:
Rebased against the latest HEAD.  @kacf please review.

drewmoseley:
@kacf Changelog added.

drewmoseley:
@kacf in thinking about this a bit more, I wonder if it would be better to use MENDER_FEATURES_ENABLE and something like mender-raspberrypi?  We could then include the RPI_USE_U_BOOT stuff as well and drop the user required lines to basically just 2.

@mirzak I _think_ this will also address your concerns from above.

Thoughts?

kacf:
I think it's unneeded. We already have all the required information, we know that it's RPi from `SOC_FAMILY`, and we also know that U-Boot is being used from the `mender-uboot` feature being enabled in `DISTRO_FEATURES`. So we can test for those two things. I think the U-Boot dependency should be there regardless of board actually, it's valid for all boards, not just RPi. So something like:

```
def sdimg_deps(distro_features, soc_family):
    list = ""
    if 'rpi' in soc_family:
        list += " bcm2835-bootfiles:do_populate_sysroot"
    if 'mender-uboot' in distro_features:
        list += " u-boot:do_deploy"
    return list
do_image_sdimg[depends] += "${@sdimg_deps('${DISTRO_FEATURES}', '${SOC_FAMILY}'}"
```

kacf:
I don't understand why the tests fail, it doesn't look related to these commits. I've opened [this PR](https://github.com/mendersoftware/meta-mender/pull/393) just to debug the tests.

kacf:
Merging for now, we can always improve it more later.

PR #382, opened 2017-09-26
[RFC] Go support changes

drewmoseley:
I needed these changes to get a master-based build to succeed.  I did see at least one relevant change in poky (https://git.yoctoproject.org/cgit.cgi/poky/commit/?id=07ef4dd93dcb21418647932f3324ee09bc514895) that would require some of this.  I didn't dig too deeply and these are likely not the best solution but I needed something that would build.

Any idea what the right fix is for these?


kacf:
Hmm, I think the rsync could be removed completely actually, since they now point `${B}` to `${S}`. I'm planning to bring our master branch up to date this week, so I will get some more time to look at this, and might cherry-pick your patches. I'll close this one, since I will put a new one up anyway.

PR #384, opened 2017-09-26
Fix custom set IMAGE_ROOTFS_SIZE not being respected.

kacf:
This partially reverts variable assignment depending on distro
features.

What we want is for the assignment to happen if, and only if:
* mender-install feature is enabled
* there is no previous assignment

However it seems impossible to do this correctly, because:
* If we use an override, then even a weak default assignment overrides
  any other hard assignment.
* If we use inline Python to detect the distro feature, then the code
  executes at the end instead of where it occurs, which means our
  assignment is overridden by earlier defaults.
* It is not possible in the Python code to tell whether the currently
  set value is a default or not.
* Using a Python function in the assignment itself, we can execute at
  the right moment, but we get an infinite recursion if we try to
  access the already set value of the variable, which we'd need in
  order to set a new one conditionally.

Therefore we have to set the default always, regardless of distro
feature. Most likely this isn't a big issue since it is only a
default, which can be overridden by using the '=' operator.

Changelog: Title


kacf:
Build was accidentally cancelled, but I'll merge this, since this is a revert to previous behavior, and this has been verified both by myself and an external user.

PR #387, opened 2017-09-28
Bringing poky/master up to date

kacf:


kacf:
@drewmoseley: I have cherry-picked your patches here. Will try this together with some support changes in mender-qa and poky.

drewmoseley:
Thanks for this.  FWIW I pulled this into my setup today and verified that all my builds succeeded.  I didn't do a runtime test on them but that shouldn't be a huge issue.

PR #389, opened 2017-10-02
pre-generate ssh keys before booting mender-qemu

GregorioDiStefano:
Changelog: none


GregorioDiStefano:
oops, was looking at an old sdimg, paths updated

PR #394, opened 2017-10-11
Pyro update

kacf:

PR #398, opened 2017-10-12
Fix parse error in MENDER_CERT_LOCATION when PREFERRED_VERSION is unset.

kacf:
```
commit e2e69ebaf8c7dd5ebeae24a913c0faa1d25d180a
Date:   Thu Oct 12 13:55:01 2017

    Fix parse error in MENDER_CERT_LOCATION when PREFERRED_VERSION is unset.
    
    Also add a long desired test, which tests all common combinations of
    PREFERRED_VERSIONs. This tests that we can build all tagged versions
    of Mender, something that isn't normally checked in Jenkins, because
    it builds using PREFERRED_VERSION and a SHA. It also checks that
    different types of PREFERRED_VERSION, including with and without
    "pn-", and including being missing altogether, builds successfully.
    
    Changelog: Title
    
commit 92aa055d76bca814c7d8ff7e799eea1924576b24
Date:   Thu Oct 12 13:54:04 2017

    tests: Keep a backup of original local.conf instead of grepping for changes.
    
    Changelog: None
    
```

### DIFF
--- a/meta-mender-core/recipes-mender/mender/mender_git.bb
+++ b/meta-mender-core/recipes-mender/mender/mender_git.bb
@@ -6,15 +6,22 @@ require mender.inc
 # SHA.
 def mender_autorev_if_git_version(d):
     version = d.getVar("PREFERRED_VERSION")
+    if version is None or version == "":
+        version = d.getVar("PREFERRED_VERSION_%s" % d.getVar('PN'))
     if version is not None and "git" in version:
         return d.getVar("AUTOREV")
     else:
         return "f6ffa190892202263fdb75975059fbb201adab6a"
 SRCREV ?= '${@mender_autorev_if_git_version(d)}'
 
-def mender_branch_from_preferred_version(pref_version):
+def mender_branch_from_preferred_version(d):
     import re
-    match = re.match(r"^[0-9]+\.[0-9]+\.", pref_version)
+    version = d.getVar("PREFERRED_VERSION")
+    if version is None or version == "":
+        version = d.getVar("PREFERRED_VERSION_%s" % d.getVar('PN'))
+    if version is None:
+        version = ""
+    match = re.match(r"^[0-9]+\.[0-9]+\.", version)
     if match is not None:
         # If the preferred version is some kind of version, use the branch name
         # for that one (1.0.x style).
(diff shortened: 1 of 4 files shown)


kacf:
@drewmoseley: If you want to cherry-pick this to morty, you will probably have to add a second argument `True` to all the `getVar` calls.

kacf:
I'm cherry-picking this [to pyro](https://github.com/mendersoftware/meta-mender/pull/394) as well.

drewmoseley:
@kacf is there any kind of debug logging when it is testing the various values of the PREFERRED_VERSION variables? I didn't see anything obvious.

Also, how specifically does this patch fix the MENDER_CERT_LOCATION parse error? I see the updates to mender_branch_from_preferred_version() in mender_git.bb and mender-artifact_git.bb but it's not clear to me how that will fix the MENDER_CERT_LOCATION variable.  What am I missing?

drewmoseley:
re: the debug logging, never mind.  I was just looking in the wrong place; I see it now.

meta-mender-core/recipes-mender/mender/mender_git.bb
```diff
@@ -6,23 +6,30 @@ require mender.inc
...
+    match = re.match(r"^[0-9]+\.[0-9]+\.", version)
     if match is not None:
         # If the preferred version is some kind of version, use the branch name
         # for that one (1.0.x style).
         return match.group(0) + "x"
     else:
         # Else return master as branch.
         return "master"
-MENDER_BRANCH = "${@mender_branch_from_preferred_version('${PREFERRED_VERSION}')}"
+MENDER_BRANCH = "${@mender_branch_from_preferred_version(d)}"
```
kacf: @drewmoseley: The key is here. If `PREFERRED_VERSION` was unset, it would trigger a parse error, so now we pass the always defined `d` instead, and fetch the variable from there, with proper checking for `None`.

I'm not completely certain why this only started showing up once we made `MENDER_CERT_LOCATION` dependent on a function call. `MENDER_CERT_LOCATION` started depending on `MENDER_BRANCH`, so I suspect it has to do with recursive calls having undefined values. By using `d` directly instead, we avoid the undefined values.

PR #399, opened 2017-10-13
systemd: Move epoch settings back into meta-mender-core.

drewmoseley:
Changelog: none


drewmoseley:
I think this is more appropriate in meta-mender-core.

PR #403, opened 2017-10-25
u-boot: update mender_boot_part_name when mender_boot_part changes

mirzak:
mender_boot_part_name was never updated once it had been written from
the default environment.

Which led to that we never updated the Linux kernel and dtb on
a NAND/UBI setup, becuase we would always load the Linux kernel from
ROOTFS_A regardless of current active part.

While at it also removed the "persistant" storage of mender_boot_part_name
and the reason for this is that the persistant value would not match the
actual value used to boot the device since we alter it in "mender_setup"
which does not re-save the environment.

Changelog: Title

Reported-by: Dan Walkes <danwalkes@trellis-logic.com>

### DIFF
--- a/meta-mender-core/recipes-bsp/u-boot/patches/0002-Generic-boot-code-for-Mender.patch
+++ b/meta-mender-core/recipes-bsp/u-boot/patches/0002-Generic-boot-code-for-Mender.patch
@@ -101,7 +101,7 @@ new file mode 100644
 index 0000000000000000000000000000000000000000..b4fcdb8148311362222415ae13b37efcfc87d682
 --- /dev/null
 +++ b/include/env_mender.h
-@@ -0,0 +1,76 @@
+@@ -0,0 +1,79 @@
 +/*
 +  Copyright 2017 Northern.tech AS
 +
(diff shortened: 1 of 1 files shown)


mirzak:
The background of this can be found on the mailinglist [1].

[1]. https://groups.google.com/a/lists.mender.io/forum/?nomobile=true#!topic/mender/y3orr23h2L4

kacf:
Running through tests.

mirzak:
> Looks good to me! I don't think this will fix existing devices though, because the already saved environment will be used instead of this one. This is a weakness of the current U-Boot approach.

Yeah, this is true.

Two things:
- Should we mark UBI support as "broken" in the current release in some doc file? Because it really is :(.
- Should a test be added that exposes this kind of error?

kacf:
> Should we mark UBI support as "broken" in the current release in some doc file? Because it really is :(.

I think it's a normal bug like any other, albeit a serious one. You can still update 99% of the fs correctly.

The fact that it doesn't fix existing devices is annoying though, we should think hard if there's some way we can fix that.

> Should a test be added that exposes this kind of error?

Ideally, yes, but I'm not sure how to separate one kernel from another. How could you tell which kernel you're running after you've booted without compiling two different kernels?

I'm open to ideas to both of the above. :-)

kacf:
Thanks in any case!

mirzak:
> I think it's a normal bug like any other, albeit a serious one. You can still update 99% of the fs correctly.

Yeah, just wanted to get your opinion. 

> The fact that it doesn't fix existing devices is annoying though, we should think hard if there's some way we can fix that.

Will try harder. Though would be interesting to know how widely used UBI/UBIFS support is, in production that is. 

But would still be interesting to handle updates that happen to "ENV"  area. I have always struggled with this in "homegrown" solutions.

> Ideally, yes, but I'm not sure how to separate one kernel from another. How could you tell which kernel you're running after you've booted without compiling two different kernels?

I simple solution to this is to "invalidate" the inactive partition (erase or corrupt it) which would prevent "accidental" accesses to it by simply failing. Would at least cover this case, and might cover other "cross rootfs access" problems.

kacf:
> I simple solution to this is to "invalidate" the inactive partition (erase or corrupt it) which would prevent "accidental" accesses to it by simply failing. Would at least cover this case, and might cover other "cross rootfs access" problems.

How about this one: https://github.com/mendersoftware/meta-mender/pull/415

PR #406, opened 2017-10-30
Document how to make new recipes for new releases.

kacf:
Changelog: None


drewmoseley:
@kacf should we create a similar readme in the mender-artifact recipe dir?  Should this be backported?

kacf:
This is primarily for use with the release checklist, which points directly to the master version. As such it's not important to backport it, unless you see a need.

For maintenance reasons I would prefer not to have two distinct versions of this file in mender and mender-artifact, but I can add a symlink to the latter.

kacf:
#409

PR #409, opened 2017-11-01
Add recipe instructions to mender-artifact by linking to mender. 

kacf:
Changelog: None


kacf:
Hmm, github doesn't indicate that the added file is a symlink, but it is.

PR #412, opened 2017-11-07
Fix dtb handling morty

drewmoseley:


drewmoseley:
@kacf anything of concern here?  These are all changes I made for the RPi demo work.  Most of them are cherry picks so I think they are ok.

drewmoseley:
Hmm.  Build failure in volatile-binds.  Likely due to https://github.com/mendersoftware/meta-mender/pull/412/commits/31384813b0dc36d21fe4a761204c71887c3779b3.  I have not rebuilt since I synced my sources this afternoon.  Maybe something upstream has changed.  I'll run a local build and see what I can find.

kacf:
> Hmm. Build failure in volatile-binds. Likely due to 3138481. I have not rebuilt since I synced my sources this afternoon. Maybe something upstream has changed. I'll run a local build and see what I can find.

You can almost certainly omit that cherry-pick. AFAIK, the change that prompted it was introduced in poky/pyro only. Morty will not need the same treatment.

kacf:
Has 0b4f7ca been applied to master? Seems like it would be useful there as well.

Otherwise this looks good to me!

drewmoseley:
@kacf actually I added the volatile-binds patch because the morty branch started having the same issue as seen in pyro.  I'll dig a bit more to see if I can find what changed before merging.

drewmoseley:
Regarding https://github.com/mendersoftware/meta-mender/commit/0b4f7caa259c1cf5039af3399da7bd5821ebc1db, it wasn't needed in master/pyro but off the top of my head, I don't recall why.  I'll take a look at that as well.

drewmoseley:
@kacf regarding https://github.com/mendersoftware/meta-mender/commit/0b4f7caa259c1cf5039af3399da7bd5821ebc1db I can't find a specific commit in UBoot that makes that unnecessary in pyro and morty but it's pretty obvious when it happens as the variable expansion is done incorrectly and the boot fails.  As long as we are not seeing issues in those branches I would recommend _not_ applying that patch.

kacf:
Ok.

drewmoseley:
@kacf I don't see the volatile-binds build failure locally but I'm using the HEAD of morty.  Is the CI build using an older hash from the morty branch?

kacf:
@drewmoseley: Oh probably! I didn't think about that. It pulls from the same repo as the rest of the jobs: https://github.com/mendersoftware/poky/tree/morty. You can push to the morty branch there to update what gets built.

kacf:
Feel free to force push, this is *only* used for build purposes.

drewmoseley:
OK.  I updated our poky fork.  We'll see if this build succeeds.

kacf:
The build error I see there now seems to be because the mender recipe that's being used is old (1.0.1), while the checked out version is master. You'll have to either add the Git recipe or a more recent stable recipe.

drewmoseley:
@kacf I don't follow.  How does the CI system use a checked out version that is different from what the recipes provide?

drewmoseley:
OK.  I added the latest _git recipes modified to work with morty.  Hopefully that will get us past the CI build failures at the very least.

kacf:
```
ERROR: core-image-full-cmdline-1.0-r0 do_image_mender: Size of rootfs is greater than the calculated partition space (110534656 > 109051904). This image won't fit on a device with the current storage configuration. Try reducing IMAGE_OVERHEAD_FACTOR if it is higher than 1.0, or raise MENDER_STORAGE_TOTAL_SIZE_MB if the device in fact has more storage.
```

You might want to cherry-pick [this one](https://github.com/mendersoftware/meta-mender/commit/071c2f6e610adc7ecfbcac5b7b20aec918f697b3).

drewmoseley:
Well, we at  least made it past the build test now.

kacf:
Probably the tests are very out of date. I would consider simply dumping all files inside `tests` in the master branch into the branch, don't even bother cherry-picking each commit.

kacf:
There might be a couple of build tests failing due to differences in the branches, but I think those can be selectively disabled with `@pytest.mark.skip` over the test function. I'm thinking especially of `test_build.py`.

kacf:
We'll see how it goes!

drewmoseley:
No luck

kacf:
Oh, right, morty was before recipe specific sysroots. Man, one forgets how much has changed.

Can you try replacing the occurrence of `prepare_recipe_sysroot` with `unpack` in `meta-mender/tests/acceptance/common.py`?

drewmoseley:
OK.  I added that as a separate commit.  If it works I can squash that before merging to avoid bisect breakage.

kacf:
```
ERROR: Nothing PROVIDES 'mender-test-dependencies'
```

Oh, it's probably a missing commit. I don't have time to figure out which one right now, but shouldn't be hard to find.

drewmoseley:
I've resolved as many of the acceptance test failures as I know how to fix.  Any suggestions on the others?

kacf:
There are a lot of format errors, which I don't quite get, since it should compile `mender-artifact` from master, even though only the recipe for 1.0.0 is in morty. But maybe you can try to upgrade the `mender-artifact` recipe? Because it certainly is old, and the whole 1.x series has been unsupported for a while now.

drewmoseley:
No luck with the newer recipes.  I _think_ the problem is that the golang support in morty is incompatible with the newer recipes.

drewmoseley:
Hmm.  Now the tests are not automatically running.  Any idea why @GregorioDiStefano or @kacf ?

kacf:
I think it is because DigitalOcean ran out of High CPU droplets. Unfortunately we haven't found a way to have it automatically revert to Normal CPU droplets.

If this happens (the symptom in Jenkins is that there are jobs in the queue, but no droplets are spawned even after a long time), just launch the "normalcpu-emergency-launcher" job. It's a dummy job whose only task is to make sure build slaves get spawned. Sorry, we will try to automate this, but Jenkins and Digital Ocean are giving us surprises like this constantly.

For now I've started a new job for you.

kacf:
> No luck with the newer recipes. I think the problem is that the golang support in morty is incompatible with the newer recipes.

The commits below are the ones I have found that are directly related to the recent Go build chain changes. You can try to revert them if present, or manually make sure that morty contains the old versions.

```
commit 6602f3a4cdcf84577acc661dc78d69ce967b2c4b
Date:   Thu Sep 28 14:41:55 2017

    Don't copy sources into GOPATH, it's not needed by latest poky/master.
    
    It's already checked out at the right place.
    
    Changelog: Fix build failure after poky switched to checking out Go
    sources under the full GOPATH.
    
commit f6b15a65d016f663dff35ece91b173c806ee62b2
Author: Drew Moseley <drew.moseley@northern.tech>
Date:   Tue Sep 26 03:29:29 2017

    mender: Use GO environment variable to launch the compiler
    
    Changelog: Title
    
commit eeef780a405372233cc60da91df7d9a03f81dbe2
Date:   Thu Sep 28 18:05:00 2017

    mender-artifact: Fix build failure due to poky golang source directory changes.
    
    Yocto's golang support has changed so that the src directories
    are no longer correct.  Use the new full paths.
    
    Changelog: Title
    
    Based on a patch by Drew Moseley <drew.moseley@northern.tech>

commit b9812917e50c7a9fcd3e1882a17aa240a517f24d
Author: Drew Moseley <drew.moseley@northern.tech>
Date:   Tue Sep 26 03:25:41 2017

    mender-artifact: Fix build failure due to poky golang support changes.
    
    The current poky creates a symlink for src pointing at the original
    location such that when we try to rsync we end up deleting both
    the original and the new.
    
    Changelog: Title
    
```

drewmoseley:
Copying the comment from https://github.com/mendersoftware/integration/pull/403#issuecomment-347514765:
> It is worth noting that the latest error you got in mendersoftware/meta-mender#412, this one, appears to be legitimate failure, that test should definitely pass, on all versions going back to 1.1.0. If it fails there may be a real bug in the build pipeline somewhere, maybe it's not picking up the right version of the mender-artifact tool? If you keep having trouble with this one, let me know, I can take a look.

@kacf I'm hesitant to spend much more time on this but if you have a suggestion to quickly address this acceptance failure then at least we can get this work committed to the morty branch just in case it is useful to a user at some point. I'd hate to abandon potentially useful work.  I've tried a bunch of changes to the mender-artifact versions and have not had any luck.

kacf:
> I'm hesitant to spend much more time on this but if you have a suggestion to quickly address this acceptance failure then at least we can get this work committed to the morty branch just in case it is useful to a user at some point. I'd hate to abandon potentially useful work. I've tried a bunch of changes to the mender-artifact versions and have not had any luck.

I don't really have a quick fix no. And this failure in particular worries me because it is an indication that the tool is producing an invalid artifact. This probably won't work in production. I'll can see if I can figure out something tomorrow.

drewmoseley:
Nice.  Through the qemu acceptance tests so that's moving in the right direction.

kacf:
Do you need any more help from me here, or are you good to go with the current status? :-)

drewmoseley:
I'm good with this as is.  It's really just getting the last bits merge for posterity at this point so I think the integration failures are acceptable.  Quoting @kacf from https://github.com/drewmoseley/meta-mender/pull/1:
> The integration tests are still failing, but I think we can leave this alone for now. In morty, the Docker generation was still in its infancy, so we'd have to bring in up to speed. I think the acceptance tests are already a good indicator that it's working as intended, even if the full test suite isn't.

Merging and calling this done.

kacf:
Huzzah!

PR #418, opened 2017-11-10
add rpi3 support

GregorioDiStefano:
Changelog: none

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ The following layers in this repository are officially supported by Mender:
 
 * `meta-mender-core` (Core components of Mender)
 * `meta-mender-beaglebone` (Beaglebone board support)
+* `meta-mender-raspberrypi` (Raspberry Pi board support, actively testing RPi3)
 * `meta-mender-qemu` (QEMU emulator board support)
 
 The rest of the layers are community supported, and the response to issues may vary.


GregorioDiStefano:
@kacf @estenberg

README.md
```diff
@@ -10,6 +10,7 @@ The following layers in this repository are officially supported by Mender:
 
 * `meta-mender-core` (Core components of Mender)
 * `meta-mender-beaglebone` (Beaglebone board support)
+* `meta-mender-raspberrypi` (Raspberry Pi board support, actively testing RPi3)
```
kacf: Do we need to mention "actively testing"? This is true for the others too.

mirzak: Jumping in here :). 

IMO it should at least say that only RPi 3 (32-bit?, you can build it with 64-bit support as well) is part of the CI environment.

GregorioDiStefano: @kacf only rpi3 is being tested, but the layer supports RPI 0/1/2 as well.

estenberg: For now I would say
* `meta-mender-beaglebone` (Beaglebone Black reference device)
* `meta-mender-raspberrypi` (Raspberry Pi 3 reference device, community support for other RPi versions)

If we are only testing 32-bit RPi3 and this makes a difference indeed we should have rather have "Raspberry Pi 3 32-bit reference device, ..."

drewmoseley: Agree.  We need to clarify that only RPi3/32-bit has been formally tested.  I've tried to build the 64-bit  and it fails on morty and pyro, even without the Mender integration.  I've used RPi2 pretty extensively so I think it's sound.  RPi0 I don't think will work yet because U-Boot doesn't seem to have support for it in the upstream, although I have seen a patch for it.

kacf:
Outdated.

PR #419, opened 2017-11-16
Fix incorrect line number test for Mender artifact version 2

drewmoseley:


kacf:
Changelog! :-)

drewmoseley:
The build failure here is not due to this change.  It's a bug in the upstream meta-raspberrypi layer.  An open PR with the fix for it is here: https://github.com/agherzan/meta-raspberrypi/pull/156

kacf:
Yep, I see it!

Keep in mind that fixes that don't change external behavior don't need a changelog entry. When I say "Changelog!" I mean it needs a `Changelog: None` to indicate this fact! :-) All (relevant) tests passed otherwise, so I just fixed that and merged this manually.

PR #422, opened 2017-11-22
tests: Use device type number instead of device string.

kacf:


kacf:
Dammit, this doesn't work for Flash... Will need some additional tweaking.

kacf:
Turns out this wasn't the problem after all. Closing.

PR #427, opened 2017-11-28
MEN-1387: Implement heuristic automatic patching of U-Boot.

kacf:
The main tools in the process are uboot_auto_configure.sh and
uboot_auto_patch.sh. The former sets up the necessary prerequisites,
and the latter does the actual patching.

The prerequisites include making a copy of the source and then
compiling a host version of the fw tools. This gives access to a
direct dump of the default environment, as well as build information
that helps with locating files that need to be patched.

The last step in uboot_auto_configure.sh is calling
uboot_auto_patch.sh, which uses the information from the prerequisite
steps to go through all the required patch steps. This is basically
everything listed at [1].

As part of this, we delete the patch files for QEMU and Beaglebone,
since they are not needed anymore. QEMU needs a tiny new patch, which
is necessary only because we ourselves have created the vexpress-qemu
machine type, and we need to define load addresses for the kernel and
dtb file. Raspberry Pi unfortunately has to keep its manual patching
because of its use of the boot.scr and cmdline.txt features, which we
currently don't patch correctly.

Testing this feature is tricky and time consuming. A couple of tests
have been added to test that the MENDER_UBOOT_AUTO_CONFIGURE variable
and other u-boot recipe related steps are working correctly. In
addition there is also a test which compile tests all U-Boot boards,
and this test is monstruous, taking more than an hour even on a 20
core machine. To deal with this, the test has revisions listed inside
it that can be used to skip the test for revisions that have already
been tested. Since the test only tests U-Boot source code and our
patching tools, it will only ever change status if either the U-Boot
version has changed, or one of our patching tools for U-Boot. Mender
source code changes have no effect on it.

[1] https://docs.mender.io/1.2/devices/integrating-with-u-boot

Changelog: Implement heuristic automatic patching of U-Boot.
It can be turned on and off by setting `MENDER_UBOOT_AUTO_CONFIGURE`
to `1` and `0`, respectively, in a `u-boot.bbappend` file. It is on by
default. If the automatic patching is unsuccessful, there is a special
bitbake target that can be used to extract the generated patch and use
it as a basis for a manual patch. It can be invoked with `bitbake -c
save_mender_auto_configured_patch <u-boot-recipe>`, where
`<u-boot-recipe>` is either `u-boot` or the fork of U-Boot that your
board uses.


kacf:
@drewmoseley, @pasinskim: This is a big one, but maybe you have time to take a look? :-)

kacf:
@drewmoseley: Btw, I haven't checked how well this works with the Toradex layer. Maybe you'd like to give that a spin? In fact, if you have time to try this branch with some boards, that would be great. Basically just remove any patches you have for the board, and let the script work its magic.

drewmoseley:
@kacf is there a chance this will work with Toradex which uses raw flash?

kacf:
@drewmoseley: Oh you're right, it won't. There is [a separate task](https://tracker.mender.io/browse/MEN-1597) to make this work for Flash based boards. So the proper fix for Toradex for now is to add the `MENDER_UBOOT_AUTO_CONFIGURE = "0"`.

kacf:
Added that to the toradex layer.

meta-mender-core/recipes-bsp/u-boot/files/uboot_auto_configure.sh
```diff
@@ -0,0 +1,139 @@
...
+# Prepare env tools for host platform.
+bash $SUB_X "$SCRIPT_DIR/uboot_auto_patch.sh" --patch-config-file
+$MAKE HOSTCC="$BUILDCC -DMENDER_AUTO_PROBING" CC="$BUILDCC -DMENDER_AUTO_PROBING" $ENV_TARGET
+
+# Prepare a fake environment to make work fw_printenv properly. Doesn't have
+# to be valid, just existing.
+dd if=/dev/zero of=fake-env.txt count=1 bs=$(printf %d "$BOOTENV_SIZE")
+cat > fw_env.config <<EOF
+fake-env.txt 0 $BOOTENV_SIZE
+fake-env.txt 0 $BOOTENV_SIZE
```
pasinskim: Is this for the original and redundant environment? Or should be there just once?

meta-mender-core/recipes-bsp/u-boot/files/uboot_auto_patch.sh
```diff
@@ -0,0 +1,359 @@
...
+    kconfig_repl="$(definition_for_kconfig_files "${2:-}" "${3:-}")"
+
+    # c_repl refers to replacement text in C files.
+    # c_sed_op refers to operation on C files.
+    # kconfig_repl refers to replacement text in Kbuild files.
+    if [ -n "${3:-}" ]; then
+        # Replace.
+        c_sed_op="s%.*%$c_repl%"
+        should_exist=1
+    elif [ -n "${2:-}" ]; then
```
pasinskim: Isn't this exactly as the condition above? Can't you do `if [ -n "${3:-}" ] || [ -n "${2:-}" ]  ; then`?

tests/acceptance/test_uboot_automation.py
```diff
@@ -0,0 +1,168 @@
...
+import subprocess
+
+# Make sure common is imported after fabric, because we override some functions.
+from common import *
+
+class TestUbootAutomation:
+    def revision_already_checked(self):
+        # Update these two to skip the check. You can obtain them with the same
+        # Git command as used below.
+        expected_poky_rev = "e553de00a372b193c64c267853a84c608d1aac6d"
```
pasinskim: Will this mean we will need to update this every time we will update poky? If so can we make some explicit comment saying that?

tests/acceptance/test_uboot_automation.py
```diff
@@ -0,0 +1,168 @@
...
+        total = 0.0
+        for file in os.listdir(env['LOGS']):
+            total += 1
+            with open(os.path.join(env['LOGS'], file)) as fd:
+                if "AutoPatchFailed\n" in fd.readlines():
+                    failed += 1
+
+        # PLEASE UPDATE the version you used to find this number if you update it.
+        # From version: v2017.09
+        measured_failed_ratio = 747.0 / 1176.0
```
pasinskim: The one problem I can see here is that we will need to update this somehow from time to time. What if new version of u-boot will cause increase/decrease which is in range (and we are not updating the number here), and then the following one will be out of boundaries. I know this is rather hard problem to solve, but something to think of.

pasinskim: Maybe we could not fail the test but just print the message? Can we have some soft failures implemented somehow?

meta-mender-core/recipes-bsp/u-boot/files/uboot_auto_configure.sh
kacf: Indeed. It just needs to be present twice in the file to satisfy the redundant environment requirement. Its contents don't matter.

meta-mender-core/recipes-bsp/u-boot/files/uboot_auto_patch.sh
kacf: Yes, you're right, well spotted! Will fix.

tests/acceptance/test_uboot_automation.py
kacf: Ok, I will improve the comment.

tests/acceptance/test_uboot_automation.py
kacf: > What if new version of u-boot will cause increase/decrease which is in range (and we are not updating the number here)

That's why I want it to fail. A significant change in either direction may signal a bug: Either that U-Boot has changed and the tool is now misbehaving, or that we are succeeding in patching suspiciously many boards, which may signal that there is a bug in the test itself. Both cases should be checked out, and if it looks legitimate, updating the number should be done to account for the new status. It's a bit crappy to have a test like this that we have to keep updated, but I don't see how else to test a heuristic tool like this.

I can try to improve the comment a bit more.

meta-mender-core/recipes-bsp/u-boot/files/uboot_auto_configure.sh
```diff
@@ -0,0 +1,139 @@
...
+	Source directory of U-Boot (will be modified!)
+--tmp-dir=<temp dir>"
+	Temporary directory to use while working
+--debug
+	Lots of debug output
+
+Note that all the arguments above can also be passed as environment variables
+that are capitalized and dashes changed to underscores.
+
+--debug
```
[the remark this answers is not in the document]
kacf: Will fix!

meta-mender-core/recipes-bsp/u-boot/files/uboot_auto_configure.sh
```diff
@@ -0,0 +1,139 @@
...
+    cat <<EOF
+$(basename "$0")
+
+Script to produce a patch for U-Boot so it will work with Mender.
+
+--config=<U-Boot machine config>
+	The U-Boot config specification that corresponds to your board. This
+	usually ends with "_defconfig".
+--src-dir=<U-Boot src dir>
+	Source directory of U-Boot (will be modified!)
```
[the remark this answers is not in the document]
kacf: Alright, adding that the diff from the original is the resulting patch.

meta-mender-core/recipes-bsp/u-boot/files/uboot_auto_configure.sh
```diff
@@ -0,0 +1,139 @@
...
+--tmp-dir=<temp dir>"
+	Temporary directory to use while working
+--debug
+	Lots of debug output
+
+Note that all the arguments above can also be passed as environment variables
+that are capitalized and dashes changed to underscores.
```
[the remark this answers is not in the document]
kacf: Fixed!

kacf:
> One other thing we should do is announce this somehow and maybe inform people that beaglebone layer is no longer needed.

Indeed, I think this deserves its own announcement email on the list.

PR #429, opened 2017-11-28
Remove recipes for Mender 1.0.x series.

kacf:
It has been out of support for a long time already.

Changelog: Title


kacf:
@estenberg, @drewmoseley: You don't see a problem with this, right?

I will keep this in master/rocko only, it will not be backported to pyro or older.

drewmoseley:
Sounds good to me.

estenberg:
:+1:

PR #434, opened 2017-12-02
Toradex master

drewmoseley:
This is the final bit to cleanup Toradex for now in master


kacf:
Changelog and signoff missing from the reverts. I think reverts can probably be omitted from the check, but I haven't implemented that yet, so you'll need them for now. I have added a personal TODO to have the checker omit reverts in the future.

Otherwise this looks good.

drewmoseley:
This is ready for merge.

PR #436, opened 2017-12-07
Fix incorrect test cleanup causing boot failure on physical boards.

kacf:
Changelog: None


kacf:
@GregorioDiStefano: Here's why your presentation failed! :-)

PR #438, opened 2017-12-10
meta-mender-raspberrypi: increase layer priority to 10

mirzak:
Currently meta-raspberrypi has BBFILE_PRIORITY set to 9, which means that our layer will be processed before the layer we append. 

This is not intended and we should have higher priority as there might be changes in meta-raspberrypi that need to be applied before we can apply ours. 

Changelog: None


mirzak:
Agreed.

Though noticed that this pull-request is wrong, I only want the first patch in the pull-request. meta-raspberrypi does yet support Pi Zero so can not merge that yet.

Will also drop the second patch.

kacf:
Thanks @mirzak!

PR #440, opened 2017-12-14
Fix device types compatible

kacf:


mirzak:
Do not have possibility to test this currently but it does look good

PR #442, opened 2017-12-14
meta-mender-core: Allow dtbos in KERNEL_DEVICETREE

scottellis:
Raspberry Pi kernel recipes use the KERNEL_DEVICETREE
variable to specify the dtbo overlays to build, not just
the board dtbs.

Ignore these dtbo overlays when assigning MENDER_DTB_NAME.

The existing behavior of choosing the last dtb in the list
if more then one is found is maintained.

Not sure this is a particularly good idea. It might be better
to abort if multiple dtbs are found since using the wrong
dtb for uboot will probably not work.

Signed-off-by: Scott Ellis <scott@jumpnowtek.com>

### DIFF
--- a/meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
+++ b/meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
@@ -91,8 +91,8 @@ do_provide_mender_defines() {
     DTB_COUNT=$(echo $MENDER_DTB_NAME | sed 's/\s/\n/g' | wc -l)
 
     if [ "$DTB_COUNT" -ne 1 ]; then
-        bbwarn "Found more then one dtb specified in KERNEL_DEVICETREE. Only one should be specified. Choosing the last one."
-        MENDER_DTB_NAME=$(echo $MENDER_DTB_NAME | sed 's/\s/\n/g' | tail -1)
+        bbfatal "Found more then one dtb specified in KERNEL_DEVICETREE. Only one should be specified."
+        exit 1
     fi
 
     cat > ${S}/include/config_mender_defines.h <<EOF
--- a/meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
+++ b/meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
@@ -81,9 +81,18 @@ do_provide_mender_defines() {
             ;;
     esac
 
-    MENDER_DTB_NAME=$(echo "${KERNEL_DEVICETREE}" | grep -o '[^ ]*$')
-    if [ "$MENDER_DTB_NAME" != "${KERNEL_DEVICETREE}" ]; then
-        bbwarn "More than one dtb file specified in KERNEL_DEVICETREE. Only one should be specified. Choosing the last one."
+    MENDER_DTB_NAME=$(echo "${KERNEL_DEVICETREE}" | sed 's/\s/\n/g' | grep -v '.dtbo' | grep -o '[^ ]*$')
+
+    if [ -z "${MENDER_DTB_NAME}" ]; then
+        bbfatal "Did not find a dtb specified in KERNEL_DEVICETREE"
+        exit 1
+    fi
+
+    DTB_COUNT=$(echo ${MENDER_DTB_NAME} | sed 's/\s/\n/g' | wc -l)
+
+    if [ "${DTB_COUNT}" -ne 1 ]; then
+        bbwarn "Found more then one dtb specified in KERNEL_DEVICETREE. Only one should be specified. Choosing the last one."
+        MENDER_DTB_NAME=$(echo $MENDER_DTB_NAME | sed 's/\s/\n/g' | tail -1)
     fi
 
     cat > ${S}/include/config_mender_defines.h <<EOF
--- a/meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
+++ b/meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
@@ -83,14 +83,14 @@ do_provide_mender_defines() {
 
     MENDER_DTB_NAME=$(echo "${KERNEL_DEVICETREE}" | sed 's/\s/\n/g' | grep -v '.dtbo' | grep -o '[^ ]*$')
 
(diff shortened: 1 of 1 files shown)


meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
```diff
@@ -81,9 +81,18 @@ do_provide_mender_defines() {
...
+    MENDER_DTB_NAME=$(echo "${KERNEL_DEVICETREE}" | sed 's/\s/\n/g' | grep -v '.dtbo' | grep -o '[^ ]*$')
+
+    if [ -z "${MENDER_DTB_NAME}" ]; then
+        bbfatal "Did not find a dtb specified in KERNEL_DEVICETREE"
+        exit 1
+    fi
+
+    DTB_COUNT=$(echo ${MENDER_DTB_NAME} | sed 's/\s/\n/g' | wc -l)
+
+    if [ "${DTB_COUNT}" -ne 1 ]; then
```
kacf: Use `$DTB_COUNT`.

meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
```diff
@@ -81,9 +81,18 @@ do_provide_mender_defines() {
             ;;
     esac
 
-    MENDER_DTB_NAME=$(echo "${KERNEL_DEVICETREE}" | grep -o '[^ ]*$')
-    if [ "$MENDER_DTB_NAME" != "${KERNEL_DEVICETREE}" ]; then
-        bbwarn "More than one dtb file specified in KERNEL_DEVICETREE. Only one should be specified. Choosing the last one."
+    MENDER_DTB_NAME=$(echo "${KERNEL_DEVICETREE}" | sed 's/\s/\n/g' | grep -v '.dtbo' | grep -o '[^ ]*$')
```
kacf: I suggest using `sed -r 's/\s+/\n/g'`, then I think you can drop the final `grep -o`. That one is for getting rid of empty lines, right?

scottellis:
Did you try any of your suggestions removing the extra sed lines?

They don't work for me. It's pretty easy to test.

I can remove the {} around variables. That's fine.

What about aborting instead of just warning if multiple dtbs are found?


"Kristian Amlie" <notifications@github.com> said:

kacf:
> Did you try any of your suggestions removing the extra sed lines? They don't work for me. It's pretty easy to test.

I think it may have been because quotes were missing around the initial assignment, so the newlines were not included.

> What about aborting instead of just warning if multiple dtbs are found?

I think we should not abort. It's common to include aliases for dtbs (for instance Beaglebone does this, I think), where the underlying files are identical. It's still advised to use only one, but I think it should be allowed not to.

I had to test this myself, so I took the pull request, made some adjustments and resubmitted in #444. Let me know if anything is missing there that you need.

scottellis:
You still need to strip leading whitespace from KERNEL_DEVICETREE.

I added some debug to show.

This is fine

KERNEL_DEVICETREE = "bcm2710-rpi3-b.dtb \
    overlays/pi3-disable-bt.dtbo \
    overlays/pi3-disable-wifi.dtbo \
"

... log
WARNING: u-boot-1_2017.09-r0 do_provide_mender_defines: *** KERNEL_DEVICETREE: bcm2710-rpi3-b.dtb overlays/pi3-disable-bt.dtbo overlays/pi3-disable-wifi.dtbo
WARNING: u-boot-1_2017.09-r0 do_provide_mender_defines: *** MENDER_DTB_NAME: bcm2710-rpi3-b.dtb
WARNING: u-boot-1_2017.09-r0 do_provide_mender_defines: *** DTB_COUNT: 1
WARNING: u-boot-1_2017.09-r0 do_provide_mender_defines: *** Final MENDER_DTB_NAME: bcm2710-rpi3-b.dtb


This gives a misleading warning message because MENDER_DTB_NAME still has two lines

KERNEL_DEVICETREE = " \
    bcm2710-rpi-3-b.dtb \
    overlays/pi3-disable-bt.dtbo \
    overlays/pi3-disable-wifi.dtbo \
"
... log
WARNING: u-boot-1_2017.09-r0 do_provide_mender_defines: *** KERNEL_DEVICETREE:  bcm2710-rpi-3-b.dtb overlays/pi3-disable-bt.dtbo overlays/pi3-disable-wifi.dtbo
WARNING: u-boot-1_2017.09-r0 do_provide_mender_defines: *** MENDER_DTB_NAME:
bcm2710-rpi-3-b.dtb
WARNING: u-boot-1_2017.09-r0 do_provide_mender_defines: *** DTB_COUNT: 2
WARNING: u-boot-1_2017.09-r0 do_provide_mender_defines: Found more then one dtb specified in KERNEL_DEVICETREE. Only one should be specified. Choosing the last one.
WARNING: u-boot-1_2017.09-r0 do_provide_mender_defines: *** Final MENDER_DTB_NAME: bcm2710-rpi-3-b.dtb


Unfortunately the latter is how KERNEL_DEVICETREE is often defined.

This is the code I was running
...
    bbwarn "*** KERNEL_DEVICETREE: ${KERNEL_DEVICETREE}"

    MENDER_DTB_NAME="$(echo "${KERNEL_DEVICETREE}" | sed -r 's/\s+/\n/g' | grep -v '.dtbo')"
    bbwarn "*** MENDER_DTB_NAME: $MENDER_DTB_NAME"

    if [ -z "$MENDER_DTB_NAME" ]; then
        bbfatal "Did not find a dtb specified in KERNEL_DEVICETREE"
        exit 1
    fi

    DTB_COUNT=$(echo "$MENDER_DTB_NAME" | wc -l)
    bbwarn "*** DTB_COUNT: $DTB_COUNT"

    if [ "$DTB_COUNT" -ne 1 ]; then
        bbwarn "Found more then one dtb specified in KERNEL_DEVICETREE. Only one should be specified. Choosing the last one."
        MENDER_DTB_NAME="$(echo "$MENDER_DTB_NAME" | tail -1)"
    fi

    bbwarn "*** Final MENDER_DTB_NAME: $MENDER_DTB_NAME"
...




"Kristian Amlie" <notifications@github.com> said:

kacf:
Ah yes, you're right. I'll add that.

PR #443, opened 2017-12-15
tests: Standardize on long board names and remove unused functions.

kacf:
Changelog: None

Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>

PR #451, opened 2017-12-21
Matrixed jenkins

kacf:


kacf:
Built on #443. Not sure if I will merge that first or not. Depends on testing.

PR #452, opened 2017-12-21
mender: Quote bootargs for expansion with embedded variables. 

drewmoseley:
Changelog: none


drewmoseley:
@kacf I put a version of this in Morty since I saw some issues with it there.  I am seeing similar behavior on a raspberrypi-cm platform.  I'm not sure why it's needed on some platforms and not others.  I suspect there is a UBoot fix or configuration that may be contributing to things.  Thoughts?

kacf:
I think it's always better to quote than not to quote. I'll rerun this through Jenkins, I think it's just a hiccup, but given the nature of this fix it *could* be legit.

drewmoseley:
@kacf how quickly can we get this merged back into rocko?  I need it for services.

kacf:
I can do that right now. See #455.

PR #453, opened 2017-12-22
Update README.md

mterwoord:
- Removed beaglebone, as folder doesn't exist anymore.
- Added raspberry layer, because according to Mender 1.3 release notes, it's supported now.


kacf:
Good catch, thanks! Merged manually to master and rocko after adding tags.

Please submit to master in the future. We don't normally merge pull requests into release branches unless they are branch specific fixes.

PR #463, opened 2018-01-17
mender: Allow empty in mender-wait-for-timesync.

drewmoseley:
Changelog: None


kacf:
We're having some problems with Jenkins, let me rerun this in a day or two.

drewmoseley:
@kacf I need this for a customer deliverable.  Any chance it can be expedited?

kacf:
Alright. It should be fairly safe, since we don't build it by default.

PR #467, opened 2018-01-25
Rocko update

kacf:

PR #468, opened 2018-01-27
Morty update

drewmoseley:
This merge brings morty up to date with the current pyro level of functionality.  Rather than trying to backport all the missing changes, I started from the current pyro sources, and modified it as needed to make it work with Morty branches of the other repositories.  I've manually tested on Rpi3, Beaglebone and vexpress-qemu platforms.


drewmoseley:
@kacf can you take a look at this?  It's a pretty big change but if you look at the branch https://github.com/drewmoseley/meta-mender/tree/wip-morty-pyrobackport I think it's fairly obvious what is being done here.  Showing it here as a merge into morty makes it a bit difficult to understand in this PR directly but the contents are identical to what is in my wip branch.  I only did this extra merge step to avoid having to do some kind of force push.

drewmoseley:
Also, I'm guessing the automatic tests won't complete; as I recall there were some issues in the morty branches that never got fixed up,.

kacf:
Nice work! I think this is exactly the way it should be done, with the merge and everything.

About the tests, I'm ignoring the physical boards right now (they may anyway depend on some of the commits I have mentioned in [this comment](https://github.com/mendersoftware/meta-mender/pull/467#issuecomment-361165375)), let's look at the QEMU builds: It looks like the error may be legitimate actually. If you look at [the lines near the bottom](https://mender-jenkins.mender.io/job/mender-builder/350/label=normalcpu%20&&%20mender_qemu_sdimg%20&&%20mender_servers/console):

```
21:08:43 + cp -vL '/home/jenkins/workspace/yoctobuild/build-vexpress-qemu/tmp/deploy/images/vexpress-qemu/core-image-full-cmdline-vexpress-qemu.*' /home/jenkins/workspace/yoctobuild/vexpress-qemu
21:08:43 cp: cannot stat '/home/jenkins/workspace/yoctobuild/build-vexpress-qemu/tmp/deploy/images/vexpress-qemu/core-image-full-cmdline-vexpress-qemu.*': No such file or directory
```

Why is that file not found? I think it should be. In the past I've sometimes seen `build-vexpress-qemu/tmp/...` being named as `build-vexpress-qemu/tmp-glibc/...`. Maybe that's the reason? What happens when you build it locally?

drewmoseley:
@kacf sorry for the delay here. Just getting backto this.  I'm not seeing any CI failures here.  Did something change in the infrastructure that would fix them?  I've never seen tmp-glibc on our standard builds (even with vexpress-qemu and vexpress-qemu-flash) but I have seen it on some other community BSPs.  I'm not sure why that name can be modified but it's never caused me any grief.

drewmoseley:
Is it worth me cherry-picking your suggested commits here before merging or can I do that afterwards?  It looks like all the _builds_ completed successfully here.  Not sure if we have actual runtime tests on this branch any more.

kacf:
It looks like the runtime tests didn't run because the image wasn't found. But clearly it was built, so this must be just some path issue during copying.

I see it as very unlikely that anything runtime would break TBH, so I would just merge this, and then pull in more cherry-picks in the next round.

PR #472, opened 2018-02-05
MEN-1006: Increased the standard mender-retry-polling interval

oleorhagen:
Changelog:Title

### DIFF
--- a/meta-mender-demo/recipes-mender/mender/mender_%.bbappend
+++ b/meta-mender-demo/recipes-mender/mender/mender_%.bbappend
@@ -4,7 +4,7 @@ SRC_URI_append = " file://server.crt"
 
 MENDER_UPDATE_POLL_INTERVAL_SECONDS = "5"
 MENDER_INVENTORY_POLL_INTERVAL_SECONDS = "5"
-MENDER_RETRY_POLL_INTERVAL_SECONDS = "1"
+MENDER_RETRY_POLL_INTERVAL_SECONDS = "30"
 
 do_compile_prepend() {
   bbwarn "You are building with the mender-demo layer, which is not intended for production use"


meta-mender-demo/recipes-mender/mender/mender_%.bbappend
```diff
@@ -4,7 +4,7 @@ SRC_URI_append = " file://server.crt"
 
 MENDER_UPDATE_POLL_INTERVAL_SECONDS = "5"
 MENDER_INVENTORY_POLL_INTERVAL_SECONDS = "5"
-MENDER_RETRY_POLL_INTERVAL_SECONDS = "1"
+MENDER_RETRY_POLL_INTERVAL_SECONDS = "10"
```
pasinskim: Nice and simple, but I think we've been considering increasing this to 30 seconds.

oleorhagen:
@pasinskim fixed

oleorhagen:
Done

kacf:
Rerunning tests... again!

kacf:
Hmm, there is still some failures here. Everything with "pull/447/head" in them can be ignored (they are a remnant of earlier checking), but even master fails.

Will rerun yet again and compare with [this one](https://s3-eu-west-1.amazonaws.com/mender-testing-reports/integration-reports/1205/report.html).

kacf:
Dammit, `test_bootstrapping` is failing *consistently* and I don't understand why.

From the output it looks like it is using the pure Fabric `run` function, but it should be using the `run` function from our `common` module, which may explain why it times out after a very short delay. It is capped at 30 minutes, whereas this test times out after less than 3 minutes, which may explain the issue; maybe it simply haven't had time to establish the connection. I will try to investigate why, but I may not have time before the summit.

oleorhagen:
If u'd like, I can have a look on friday.

kacf:
Feel free! Notice in particular:

```
/usr/local/lib/python2.7/dist-packages/fabric/tasks.py:173: in run
return self.wrapped(*args, **kwargs)
tests/test_bootstrapping.py:53: in accept_devices
run("strings /data/mender/mender-store | grep -q 'authtoken'")
```

.. how it goes straight from `accept_devices` function into the `run` function in Fabric. There should be an intermediary function in `common.py` there, which I thought would be ensured by the import order, but I guess not.

GregorioDiStefano:
Should be fixed now.

PR #473, opened 2018-02-13
fix support of CONVERSIONTYPES for sdimg

mirzak:
The image_types.bbclass defines a set of conversion types that can be
applied to images (gz, xz, bz2 and more).

With the current state it is not possible to apply an conversion type
on the sdimg using:

	IMAGE_FSTYPES += "sdimg.gz"

Because the conversion types assume that that the image is named in the
following format:

	${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}

IMAGE_NAME_SUFFIX defaults to ".rootfs" which breaks things for us. By
setting IMAGE_NAME_SUFFIX to an empty string the expected image name
format is correct and the sdimg type can support all the conversiontypes
that are present in image_types.bbclass.

Above is really helpfull to compress the "sdimg" file as it mostly consist
of zero's and by compressing it is much easier to handle, transfer between
devices or between servers.

Note that this change does not generate a compressed sdimg by default, but
only adds support for it.

Changelog: None

### DIFF
--- a/meta-mender-core/classes/mender-part-images.bbclass
+++ b/meta-mender-core/classes/mender-part-images.bbclass
@@ -22,6 +22,20 @@ python() {
 inherit image
 inherit image_types
 
+# This normally defaults to .rootfs which is misleading as this is not a simple
+# rootfs image and causes problems if one wants to use something like this:
+#
+#    IMAGE_FSTYPES += "sdimg.gz"
+#
+# Above assumes that the image name is:
+#
+#    ${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${type}
+#
+# Which results in a empty "gz" archive when using the default value, in our
+# case IMAGE_NAME_SUFFIX should be empty as we do not use it when naming
+# our image.
+IMAGE_NAME_SUFFIX_sdimg = ""
+
 mender_part_image() {
     suffix="$1"
     part_type="$2"


drewmoseley:
That's amazing timing.  I was fighting with that issue over the weekend specifically for creating *.bmap files.

Does this change in this class only affect the sdimg files?

mirzak:
Amazing! :). I have actually hit this once before and hacked around it. This time I wanted to fix it properly.

From my testing it only seems to effect the "sdimg" files, good point though and caution should be taken to not break to much :). 

I am still able to generate other images types (which has been my testcase):

```
<image name>-20180213150631.ext4
<image name>-20180213150631.ext4.gz
<image name>-20180213150631.mender
<image name>-20180213150631.rootfs.manifest
<image name>-20180213150631.sdimg
<image name>-20180213150631.sdimg.bmap
<image name>-20180213150631.sdimg.gz
<image name>-20180213150631.testdata.json
<image name>-20180213150631.wic.gz
```

drewmoseley:
@mirzak this patch removes the "rootfs" suffix from all images, and doesn't just fix the conversion of sdimg files.  That feels like it will break many deployment and CI scripts since the file names for so many things will change.

While it's not logically a single rootfs, would we be better off adding ${IMAGE_NAME_SUFFIX} to the sdimg files?  Or maybe create a symlink so that we have both the old and the new style names?

FWIW, it looks like the ubimg class is using ${IMAGE_NAME_SUFFIX} already.

@kacf thoughts?

mirzak:
> @mirzak this patch removes the "rootfs" suffix from all images, and doesn't just fix the conversion of sdimg files. That feels like it will break many deployment and CI scripts since the file names for so many things will change.

Oooh yeaah. I see it now. Fooled by the fact that I was still able to create the images, thinking that they would fail.

The intention was not to remove it on all of them.

> While it's not logically a single rootfs, would we be better off adding ${IMAGE_NAME_SUFFIX} to the sdimg files? Or maybe create a symlink so that we have both the old and the new style names?

Yeah this did occur to me as well as a possible solution, and I do like the idea of symlink to keep both styles. 

Will do some more tests on this.

drewmoseley:
I have a patch I'm testing and will post a link to here in about 5 min.

mirzak:
This would be the correct way to remove it from the "sdimg" only:

     IMAGE_NAME_SUFFIX_sdimg = ""

Verified this time

drewmoseley:
OK.  And here's my version that just adds another symlink: https://github.com/mendersoftware/meta-mender/pull/474/commits/48367dd1fb155e427d6cc4a34e64319814213be8

drewmoseley:
I don't have a strong preference for which solution we merge however, if we use the IMAGE_NAME_SUFFIX_sdimg variable, then I think we should add that to the name when the file is created in case users later on try to change the value from blank.

mirzak:
I do not have a strong preference either, but I do have an interest of getting something merged in both master and rocko.

I do not really see the benefit of adding it to the name, it is supposed to be blank as we already have a descriptive suffix ("sdimg") and it is not really available for users to change.

kacf:
Why not merge both? They are not mutually exclusive AFAICS.

drewmoseley:
Including both sounds reasonable to me.  I've got my commit over in https://github.com/mendersoftware/meta-mender/pull/474.  I'll amend it and reword the commit log a bit so it's a bit clearer.  In the meantime, I think this one is good to go.

kacf:
Unrelated error, merging. Thanks!

mirzak:
Will this be cherry-picked to "rocko"?

kacf:
Yes, it will!

mirzak:
Thanks

PR #469, opened 2018-01-27
Misc Changes from recent Dragonboard port 

drewmoseley:

PR #474, opened 2018-02-13
More misc changes for master branch

drewmoseley:
A couple of (hopefully) simpler changes than those in #469 to get merged easier.


drewmoseley:
This will depend on https://github.com/mendersoftware/meta-mender/pull/473

drewmoseley:
I rebased and cleaned up the comment for the commit adding the extra symlink so as far as I can tell this is ready to go.  I'll monitor for CI issues.

kacf:
Yep, Raspberry Pi is broken probably due to master branch mismatch. We'll fix this later. Hence we are only seeing unrelated and expected failures here. Merging.

PR #477, opened 2018-02-16
MEN-1603: Warn on unused MENDER_* variables

alfrunes:
changelog: title


meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -176,6 +176,67 @@ python() {
...
+    if(isinstance(e, bb.event.ParseCompleted)):
+        from bb import data
+        import os
+        import re 
+        import json
+
+        path = d.getVar("LAYERDIR_MENDER")
+        path = os.path.join(path, "conf/mender-vars.json")
+
+        if( (os.path.isfile(path))):
```
oleorhagen: nitpick: extra space

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -176,6 +176,67 @@ python() {
...
+            mender_vars = {}
+            with open(path, "r") as f:
```
oleorhagen: if not using the with scope, why not make this a one-liner? (:

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -176,6 +176,67 @@ python() {
...
+                    #       whereas the value is arbitrary
+                    # RFC: might want to swap bb.warn() with bb.note() in
+                    #      this case, as this does not have to be so strict
+                        expected_expressions = []
+                        val = d.getVar(k)
+
+                        if isinstance (mender_vars[k], list):
+                            # item is a list of strings
+                            for regex in mender_vars[k]: # (can be a list of items)
+                                if( re.search(regex, val) == None ):
```
oleorhagen: same as above

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -176,6 +176,67 @@ python() {
...
+                            # item is a single string
+                            regex = mender_vars[k]
+                            if (re.search(regex, val) == None): 
+                                bb.warn("%s is initialized with value \"%s\"" % (k, val),\
+                                        "\nExpected: \"%s\"" % regex)
+
+
+        else: ## if(!os.path.isfile(path)): ##
+            # This should never run, but left it in here in case we #
+            # need to generate new json file template in the future #
```
oleorhagen: I like it!

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -176,6 +176,67 @@ python() {
...
+            with open(path, "r") as f:
+                mender_vars = json.load(f)
+
+            for k in d.keys():
+                if k.startswith("MENDER_"):
+                    if ( k not in mender_vars.keys() ):
+                    # Warn if user has defined some new (unused) MENDER_.* variables
+                        bb.warn("\"%s\" is not a real MENDER_ variable" % k)
+
+                    elif ( mender_vars[k] != "" ):
```
oleorhagen: try n be consistent with the spacing used elsewhere.

meta-mender-core/classes/mender-setup.bbclass
kacf: Isn't he using it...?

meta-mender-core/classes/mender-setup.bbclass
kacf: In general Python doesn't need parentheses around if statements. This goes for the whole snippet.

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -176,6 +176,67 @@ python() {
...
+        if( (os.path.isfile(path))):
+            mender_vars = {}
+            with open(path, "r") as f:
+                mender_vars = json.load(f)
+
+            for k in d.keys():
+                if k.startswith("MENDER_"):
+                    if ( k not in mender_vars.keys() ):
+                    # Warn if user has defined some new (unused) MENDER_.* variables
+                        bb.warn("\"%s\" is not a real MENDER_ variable" % k)
```
kacf: I suggest using the message: `"\"%s\" is not a recognized MENDER_ variable. Typo?" % k`

meta-mender-core/classes/mender-setup.bbclass
oleorhagen: It's no biggie, but I was thinking, why not simply:
```
 mender_vars = json.load(open(path, "r"))
```

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -176,6 +176,67 @@ python() {
...
+
+                    elif ( mender_vars[k] != "" ):
+                    # If certain keys should have associated some restricted value
+                    # (expressed in regular expression in the .json-file)
+                    # NOTE: empty strings (json-values) are only compared by key, 
+                    #       whereas the value is arbitrary
+                    # RFC: might want to swap bb.warn() with bb.note() in
+                    #      this case, as this does not have to be so strict
```
kacf: Comment should be indented to the same level as the following code.

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -176,6 +176,67 @@ python() {
...
+                        val = d.getVar(k)
+
+                        if isinstance (mender_vars[k], list):
+                            # item is a list of strings
+                            for regex in mender_vars[k]: # (can be a list of items)
+                                if( re.search(regex, val) == None ):
+                                    expected_expressions += [regex]
+                            if (len(expected_expressions) > 0): 
+                                bb.warn("%s is initialized with value \"%s\"" % (k, val),\
+                                        "\nExpected: \"%s\"" % ', '.join(expected_expressions))
```
kacf: Won't this trigger whenever any of the expressions don't match? I would have expected it to trigger only if all of the expressions don't match.

meta-mender-core/conf/mender-vars.json
```diff
@@ -0,0 +1,83 @@
...
+     "existance in the variable in current build", 
+     "For example: we can issue a warning if user does",
+     "not include mender-uboot and mender-install",
+     "if we give the following definition below:", 
+     "'MENDER_FEATURES_ENABLE': ['mender-uboot', 'mender-install']",
+     "(using single quote (') for double quote in the above string)"],
+    "MENDER_ARTIFACT_NAME": "",
+    "MENDER_BOOT_PART": "",
+    "MENDER_BOOT_PART_DEFAULT": "",
+    "MENDER_BOOT_PART_DEFAULT_mender-ubi": "",
```
kacf: We will need to handle this specially: String like `_mender-ubi` are called an override, and there is an infinite number of possibilities for such strings, because the string can refer to some property of the build. This one in particular is "if we are building mender with ubi", but it can be anything, for example `_64bit`. The result then ends up in `MENDER_BOOT_PART_DEFAULT` if the property is true.

So I think we should remove all the `mender-ubi` strings, and before doing the matching, simply drop everything from the variable from the point where is an underscore followed by a lowercase letter.

meta-mender-core/classes/mender-setup.bbclass
alfrunes: with statement implicitly calls f.close()

meta-mender-core/classes/mender-setup.bbclass
oleorhagen: haha, didn't even notice :see_no_evil:

meta-mender-core/classes/mender-setup.bbclass
oleorhagen: haha, kk. I give

alfrunes:
@kacf 
Sorry for being a bit late with this one...

PR #478, opened 2018-02-16
MEN-1597 first round: mtdparts automated

kacf:
```
commit e1f4ccb163a8cb98077520d825c19a13148976e9
Date:   Wed Feb 14 10:20:31 2018

    Fix incorrect reference to image and machine name in qemu launcher.
    
    Changelog: None
    
commit 43378e49a7fcbfa19a9610e47abb733dceb8e35e
Date:   Fri Feb 16 16:14:13 2018

    Make sure image deployment path is correct for symlink.
    
    We should be referring to the one in the working directory. In most
    cases it doesn't make any difference, but it does if another image is
    depending on this one.
    
    Changelog: None
    
commit 74c88d595e79cf12c9a6081ca6d09dd5fc93ab53
Date:   Fri Feb 16 16:16:37 2018

    Make sure mender only runtime depends on u-boot-fw-utils.
    
    No functional difference, but triggers less rebuilds.
    
    Changelog: None
    
commit 32f46a8944e45d1c7a8d175b9c3a5245b5078840
Date:   Fri Feb 16 17:14:10 2018

    MEN-1597: Promote mtdids and mtdparts information to Bitbake variables
    
    The motivation behind the change is allowing different parts of the
    build process to be influenced by their contents. This was not
    possible before because all the information resided in U-Boot source
    code, which is not accessible outside of that recipe.
    
    It's not completely automatic because one has to define mtdid of the
    storage one wants to use. Maybe this can be automated further by
    improving the auto patching from the sdimg, but even this might be
    tricky, so we might settle for simply making it easy to extract the
    information manually, and then use that. For now MENDER_MTDIDS has to
    be set to exactly one mtdid assignment, and the rest is automatic, or
    one can set it to multiple mtdids, but then MENDER_IS_ON_MTDID and
    MENDER_MTDPARTS have to be defined manually.
    
    Changelog: The handling of mtdids and mtdparts (`MTDIDS_DEFAULT` and
    `MTDPARTS_DEFAULT` defines in U-Boot source code) has changed. The
    defines now need to be patched out of U-Boot source code if they are
    present, and added to `MENDER_MTDIDS` and `MENDER_MTDPARTS` bitbake
    variables, respectively. The benefit is that build procedures that are
    influenced by their values no longer need to be hardcoded, such as the
    offset for the U-Boot environment, whose value will be automatically
    determined. It will also be used in the upcoming mtdimg type.
    
commit edd8866fb2151236795715deec5343d00ba15952
Date:   Fri Feb 16 17:33:14 2018

    MEN-1597: Add mtdimg image type.
    
    The mtdimg type is an image type meant to be flashed to the entire
    Flash device, unlike the ubimg, which should only be flashed to the
    ubi area of the mtd device. Either one can be used depending on need.
    
    vexpress-nor image, used in QEMU, was also changed to take advantage
    of the new mtdimg image.
    
    Changelog: Commit
    
```


kacf:
I've discovered several weaknesses here while continuing the U-Boot automation work. I will resubmit this as part of that.

PR #480, opened 2018-02-26
mender-setup: add possibility to override Mender fstab entries

mirzak:
Please see commit message for additional information.

This is probably not yet "mergable" as I have not taken in to account that I might break stuff with this. Wanted some feedback before I continue.

I initially tried to work around this with ROOTFS_POSTPROCESS_COMMAND_remove, but could not successfully remove the mender_update_fstab_file entry.

I could also work around this with a custom post process command with some "sed" magic, but I wanted to see if there is a interest to moving this "upstream" or/and at least make you aware of the "problem"

### DIFF
--- a/meta-mender-core/classes/mender-setup-image.inc
+++ b/meta-mender-core/classes/mender-setup-image.inc
@@ -21,12 +21,12 @@ mender_update_fstab_file() {
     echo "# Where the U-Boot environment resides; for devices with SD card support ONLY!" >> ${IMAGE_ROOTFS}${sysconfdir}/fstab
 
     if [ -n "${MENDER_BOOT_PART}" ]; then
-        echo "${MENDER_BOOT_PART}   /uboot               ${MENDER_BOOT_PART_FSTYPE}       defaults,sync    0  0" >> ${IMAGE_ROOTFS}${sysconfdir}/fstab
+        echo "${MENDER_BOOT_PART_FSTAB_ENTRY}" >> ${IMAGE_ROOTFS}${sysconfdir}/fstab
     else
         bbdebug 2 "MENDER_BOOT_PART not set. Not adding to fstab..."
     fi
 
-    echo "${MENDER_DATA_PART}   /data                ${MENDER_DATA_PART_FSTYPE}       defaults         0  0" >> ${IMAGE_ROOTFS}${sysconfdir}/fstab
+    echo "${MENDER_DATA_PART_FSTAB_ENTRY}" >> ${IMAGE_ROOTFS}${sysconfdir}/fstab
 }
 
 # Setup state script version file.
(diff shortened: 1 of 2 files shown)


mirzak:
Also while doing this I noticed these two lines:

- https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/classes/mender-setup-image.inc#L16
- https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/classes/mender-setup-image.inc#L34

Use both "_append" and "+=", a bit odd and I wonder if it was intentional.

kacf:
Nice, I like it on first glance, but I'll look at it a bit more detailed soon! Got some deadlines to meet! :-)

drewmoseley:
@mirzak I tend to go overboard and use both _append and +=.  It doesn't hurt and in the case there happened to be a previous _append entry in the same file would handle it properly.

kacf:
Finally getting back to this. Sorry for the delay.

I think this is pretty much good to go as is. Can you just add a changelog entry that mentions the change from one to the other?

mirzak:
Would it be worth to keep the old variables around for a while to not break stuff?

Something like:

```
# This configuration is used in fstab
MENDER_DATA_PART_FSTAB_ENTRY ??= "${MENDER_DATA_PART_FSTAB_ENTRY_DEFAULT}"
MENDER_DATA_PART_FSTAB_ENTRY_DEFAULT = "${MENDER_DATA_PART}    /data    ${MENDER_DATA_PART_FSTYPE}    defaults    0 0"
MENDER_BOOT_PART_FSTAB_ENTRY ??= "${MENDER_BOOT_PART_FSTAB_ENTRY_DEFAULT}"
MENDER_BOOT_PART_FSTAB_ENTRY_DEFAULT = "${MENDER_BOOT_PART}    /uboot     ${MENDER_BOOT_PART_FSTYPE}    defaults,sync    0 0"

# Legacy fstab options - these will be removed in the future
MENDER_DATA_PART_FSTYPE ??= "${MENDER_DATA_PART_FSTYPE_DEFAULT}"
MENDER_DATA_PART_FSTYPE_DEFAULT = "auto"
MENDER_BOOT_PART_FSTYPE ??= "${MENDER_BOOT_PART_FSTYPE_DEFAULT}"
MENDER_BOOT_PART_FSTYPE_DEFAULT = "auto"
```

mirzak:
I have updated the commit message with a changelog entry if you want to merge this as-is.

kacf:
Yeah, good point. We will only add this to master, so some behavior changes are expected, but you can add a message [here](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/classes/mender-setup.bbclass#L174-L177). Otherwise looks good as is! :-)

mirzak:
Added error message @kacf

kacf:
Starting test build.

mirzak:
Note to self (or someone else :)) to update docs

mirzak:
#594 fixes the same problem that I was trying to solve here. Closing this.

PR #486, opened 2018-03-17
More changes from Dragonboard port

drewmoseley:
I've split the previous PR into 2.  I think the troublesome commits are in this PR so we'll see what the test results show.


kacf:
Sigh, yes, I'm afraid so. In fact, no client builds are working ATM. I will try to look at the problem next week, but I have to prioritize and get some coding work done this week.

drewmoseley:
@kacf the only issue in the vexpress-qemu acceptance is "Could not reconnect to host"

There are several failures in the qemux86-64 acceptance tests.  Do you think they could actually be caused by this PR or is that stuff still WIP?

kacf:
> the only issue in the vexpress-qemu acceptance is "Could not reconnect to host"

Yes, I think this is exactly the issue that I saw with your PR earlier, and it has been consistent, so this is definitely a real error.

> There are several failures in the qemux86-64 acceptance tests. Do you think they could actually be caused by this PR or is that stuff still WIP?

I was also greeted by those exact 5 tests failing this morning. Turns out it was a [race condition](https://github.com/mendersoftware/meta-mender/commit/0df5e1b4d22750e3fbc2d74559580c99b0a5f22d), so you might want to rebase before your next build. That should fix qemux86-64.

drewmoseley:
@kacf @GregorioDiStefano do we need to do some cleanup:

> WARNING: The free space of /mnt/sstate-cache (10.162.0.3:/export/sstate-cache) is running low (0.998GB left)

kacf:
Indeed we do. Started special cleanup build [here](https://mender-jenkins.mender.io/job/mender-builder/542/).

drewmoseley:
@kacf I think this is finally ready.  At least all the tests have passed.  I don't have any means to test this on the master branch as the only BSP I know of that uses 10 or more partitions is the Dragonboard which only has a platform layer that has been tested on the Morty branch.

drewmoseley:
@kacf any thoughts here?

kacf:
The failure on vexpress-qemu-flash is almost certainly because of the `MENDER_DEFAULT_LOADADDR` being missing for the UBI case. But the question of `CONFIG_LOADADDR` in U-Boot is something I'm also struggling with. Should we use `loadaddr`, which is its variable name, or should we use `kernel_addr_r`? I've asked some question on the U-Boot mailing lists, but have gotten zero response. I propose that we talk about this because it's quite important for future development. I have added it to the client meeting agenda, where I've also posted some links for context.

drewmoseley:
OK.  I originally submitted this PR without https://github.com/mendersoftware/meta-mender/pull/486/commits/2121acf841ba0c58d6438ba78ee6f866612d1801 so I'll update it as such so at least we can get the other bits merged.

drewmoseley:
@kacf I just pushed an updated version rebased on the latest HEAD with the patch context lines updated. Nothing else has changed.

kacf:
Great, thanks for pushing this!

PR #489, opened 2018-04-05
Document raspberrypi0-wifi and raspberrypi-cm support.

drewmoseley:
Changelog: None


drewmoseley:
@kacf this is obviously pretty simple.  Note that raspberrypi0-wifi explicitly requires master but raspberrypi-cm works on older branches but it may not be worth making this change in older branches.

kacf:
Makes sense, we can let demand drive it. For now, I think master is enough.

PR #490, opened 2018-04-06
test_uboot_compile: Run maximum 8 jobs to avoid I/O saturation.

kacf:
Changelog: None


kacf:
Will solve this differently.

PR #494, opened 2018-04-10
More autopatching changes

drewmoseley:
In working on the Udooneo board locally, I discovered a few other changes we should consider.  Please review.


meta-mender-core/classes/mender-part-images.bbclass
```diff
@@ -74,14 +74,39 @@ mender_part_image() {
...
+            spl_end=$(expr $spl_align_kb \* 1024 + $spl_size)
+            if [ $spl_end -gt $(expr ${IMAGE_BOOTLOADER_OFFSET} \* 1024) ]; then
+                bberror "Size of SPL specified in SPL_BINARY is too large." \
+                        "It will be overwritten by the IMAGE_BOOTLOADER_FILE." \
+                        "Please raise it manually."
+            fi
+            cat >> "$wks" <<EOF
+# embed SPL
+part --source rawcopy --sourceparams="file=${WORKDIR}/${SPL_BINARY}" --ondisk mmcblk0 --align $spl_align_kb --no-table
+EOF
```
kacf: Why is this needed in addition to `IMAGE_BOOTLOADER_FILE`? Not sure I understand the context here.

meta-mender-core/classes/mender-part-images.bbclass
```diff
@@ -74,14 +74,39 @@ mender_part_image() {
...
+
         # Copy the files to embed in the WIC image into ${WORKDIR} for exclusive access
         install -m 0644 "${DEPLOY_DIR_IMAGE}/${IMAGE_BOOTLOADER_FILE}" "${WORKDIR}/"
 
         if [ $(expr ${IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET} % 2) -ne 0 ]; then
             bbfatal "IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET must be aligned to kB" \
                     "boundary (an even number)."
         fi
-        bootloader_align_kb=$(expr $(expr ${IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET} \* 512) / 1024)
+        bootloader_align_kb=$(expr $(expr $(expr ${IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET} \* 512) / 1024) + $spl_align_kb)
```
kacf: This I find a bit unintuitive. `IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET` should be absolute, IMHO.

Also, note that `expr` can take more than two operands, and respects normal operator precedence, so we only need one call here.

meta-mender-core/recipes-bsp/u-boot/patches/0002-Generic-boot-code-for-Mender.patch
```diff
@@ -168,6 +168,10 @@ index 0000000..705c39e
 +                                                                        \
 +    "mender_setup="                                                     \
 +    "setenv mender_kernel_root " MENDER_STORAGE_DEVICE_BASE "${mender_boot_part}; "    \
++    "if test ${mmcargs} != \"\"; " \
++    "then "                        \
++    "run mmcargs; "                \
++    "fi; "                         \
```
kacf: IMHO, we should keep this out of our boot script unless `mmcargs` is really ubiquitous, and encourage calling it outside, before getting into mmcargs. I think we should be careful about this to our central boot script to avoid accumulating baggage there that we will later have to support.

Is this solely because the auto-patcher doesn't respect it and hence it doesn't get called?

meta-mender-core/classes/mender-part-images.bbclass
drewmoseley: Some U-Boot systems require a multi-stage setup.  Within U-Boot there is SPL (secondary program loader) and TPL (tertiary program loader).  I don't see any support in Yocto for TPL but there is definitely support for SPL on certain platforms.  Beaglebone actually uses it via a file in the VFAT partition (filename MLO).  In the case of the udooneo platform, there is a small SPL binary located in the raw disk image that then loads the primary U-Boot binary.  See https://github.com/graugans/meta-udoo/blob/morty/scripts/lib/wic/canned-wks/udoo-sdcard.wks for context.

meta-mender-core/classes/mender-part-images.bbclass
drewmoseley: Hmm.  There was already two calls to expr so I was simply modeling my change on that.  Should I try to clean it up?

drewmoseley:
@kacf you commented about the need for running mmcargs but I can't find the comment here.  I wonder if maybe you deleted the comment.

I can't say whether that is widely used or not but it is definitely used on at least one of the boards I have been playing with lately.

My current thinking is that perhaps we should do away with the MENDER_BOOTCOMMAND entirely and handle everything through a boot.scr file.  That would allow the platform-specific Mender layer to modify the boot.scr rather than having it directly embedded in the patch.  I'm hoping to play with that a bit in the coming weeks.

meta-mender-core/classes/mender-part-images.bbclass
mirzak: This has actually been discussed on the mailing list, @kacf @drewmoseley.  

I have been in the situation where I need to support two files for boot-loader (SPL and u-boot.bin) as well. My workaround has been to simply `dd` the two seperate files in to one in a "post_install" step in the U-boot recipe and provide combined output file to `IMAGE_BOOTLOADER_FILE`.

What we also discussed on that particular thread was to add support for passing multiple files to `IMAGE_BOOTLOADER_FILE` instead of adding a second argument.

I believe the suggestion was something like this (do not recall the exact syntax but I think the idea was the same):

```
IMAGE_BOOTLOADER_FILE = "SPL,1; u-boot.bin,69;"
```

And that is filename together with an offset. That way we can support a TPL as well or other binaries that need to be put at an offset on the storage medium.

mirzak: Found the thread: https://groups.google.com/a/lists.mender.io/forum/#!searchin/mender/mirza$20krak/mender/hyHk9a6C4K8/A23cC8qLBAAJ

kacf:
> you commented about the need for running mmcargs but I can't find the comment here. I wonder if maybe you deleted the comment.

You can find it under that little "Show outdated" link. It happens when you push a new patch that changes the context around the diff that was originally commented on.

> I can't say whether that is widely used or not but it is definitely used on at least one of the boards I have been playing with lately.
> 
> My current thinking is that perhaps we should do away with the MENDER_BOOTCOMMAND entirely and handle everything through a boot.scr file. That would allow the platform-specific Mender layer to modify the boot.scr rather than having it directly embedded in the patch. I'm hoping to play with that a bit in the coming weeks.

I fully agree with this, but that doesn't really change the matter with regards to `mmcargs`. The commands will still be the same, and I'd still be careful about adding too many special calls in that code.

For now, I think my vote goes to keeping `mmcargs` in the platform specific patch.

meta-mender-core/classes/mender-part-images.bbclass
kacf: @mirzak: Oh yes, I had completely forgotten about that. Nice catch! I would definitely prefer this approach because it is much more generic.

Tweaking it ever so slightly, I would propose the syntax:

```
IMAGE_BOOTLOADER_FILE = "SPL;1 u-boot.bin;69"
```

for two reasons:

* Similarity in to [`IMAGE_BOOT_FILES`](https://www.yoctoproject.org/docs/latest/mega-manual/mega-manual.html#var-IMAGE_BOOT_FILES), with `src;location` format.
* Using only spaces for separating entries to ease shell parsing.

The great thing about this approach is that we can support as many special files as people want.

meta-mender-core/classes/mender-part-images.bbclass
kacf: > I'm not sure IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET is absolute in the case of an SPL type configuration. Does the link I posted above give you enough to comment?

Sort of, but I think the proposal highlighted by @mirzak is a better way to handle this.

kacf:
@drewmoseley: Do you want to put the the first patches in their own PR so we can get them merged quickly? I believe only the last two of the original patches were debated.

meta-mender-core/classes/mender-part-images.bbclass
drewmoseley: I like this suggestion.  Not sure if I'll have time to implement it anytime soon though.  I created https://tracker.mender.io/browse/MEN-1883 to track it.

meta-mender-core/classes/mender-part-images.bbclass
drewmoseley: Agreed

drewmoseley:
@kacf the first 4 patches are already in #469.  I'll try to update that one with the uncontroversial patches here.

Regarding mmcargs, I agree that adding board-specific stuff is going to be problematic since it won't be a bounded problem.  I've also seen recently boards that called "findfdt" to pick an appropriate FDT file.  I'm not sure how we could enable that in the platform specific patch since the mender core stuff completely replaces the bootcmd.  Are you suggesting a platform-specific version of MENDER_BOOTCOMMAND?  That may work but for the board in question I'm in no hurry so I may use it to do a POC of boot.scr instead.

kacf:
> Regarding mmcargs, I agree that adding board-specific stuff is going to be problematic since it won't be a bounded problem. I've also seen recently boards that called "findfdt" to pick an appropriate FDT file. I'm not sure how we could enable that in the platform specific patch since the mender core stuff completely replaces the bootcmd. Are you suggesting a platform-specific version of MENDER_BOOTCOMMAND? That may work but for the board in question I'm in no hurry so I may use it to do a POC of boot.scr instead.

No, on the contrary, I really would like `MENDER_BOOTCOMMAND` to stay as platform independent as possible, but `MENDER_BOOTCOMMAND` is currently only used for the auto-patched boards. Our manual approach still suggests putting the necessary calls and mender variables into an already existing `bootcmd`. My suggestion would be to save the patch instead, turn off auto-patching and make a manual adjustment based on the patch.

Alternatively, if we want to provide a hook, we could call some generically named function, say `mender_bootcmd_init`, which is supposed to be defined by the user, where he can do whatever he pleases. This might be a good catchall solution. Again, this is independent from `boot.scr` which I think is a good idea regardless.

drewmoseley:
I think all the other patches will drop off from this PR as they are included elsewhere.  I'll take a look when I return next week and rebase everything.

drewmoseley:
Reworked the mmcargs stuff to provide generic PRE and POST setup commands.  Still need to look into a better solution for SPL

meta-mender-core/recipes-bsp/u-boot/patches/0002-Generic-boot-code-for-Mender.patch
```diff
@@ -166,7 +166,13 @@ index 0000000..705c39e
 +                                                                        \
 +    "mender_dtb_name=" MENDER_DTB_NAME "\0"                             \
 +                                                                        \
++    "mender_pre_setup_commands=" MENDER_PRE_SETUP_COMMANDS "\0"         \
++                                                                        \
```
kacf: I think I would not include this in our patch, but rather encourage the user to patch it in themselves using `SRC_URI` and a patch file (for example using `CONFIG_EXTRA_ENV_SETTINGS`). This also saves us having to introduce a bitbake variable for it.

The calling code should still work.

meta-mender-core/recipes-bsp/u-boot/patches/0002-Generic-boot-code-for-Mender.patch
```diff
@@ -166,7 +166,13 @@ index 0000000..705c39e
...
 +    "mender_setup="                                                     \
++    "if test \"${mender_pre_setup_commands}\" != \"\"; "                \
++    "then "                                                             \
++    "run mender_pre_setup_commands; "                                   \
++    "fi; "                                                              \
```
kacf: If following the suggestion above, this should be the only remaining piece of this patch.

meta-mender-core/recipes-bsp/u-boot/patches/0002-Generic-boot-code-for-Mender.patch
```diff
@@ -185,7 +187,11 @@ index 0000000..ae601f8
 +    "setenv mender_uboot_root_name ${mender_boot_part_name}; "          \
 +    "setenv expand_bootargs \"setenv bootargs \\\\\"${bootargs}\\\\\"\"; "              \
 +    "run expand_bootargs; "                                             \
-+    "setenv expand_bootargs\0"                                          \
++    "setenv expand_bootargs; "                                          \
++    "if test \"${mender_post_setup_commands}\" != \"\"; "               \
++    "then "                                                             \
++    "run mender_post_setup_commands; "                                  \
++    "fi\0"                                                              \
```
kacf: Same comment for the post approach as the pre approach.

meta-mender-core/recipes-bsp/u-boot/patches/0002-Generic-boot-code-for-Mender.patch
kacf: The motivation was to avoid more variables for things where there already is a mechanism. But maybe you're right, we are already adding some code for this, might as well add it all.

But then I would at least rename the variables to `MENDER_UBOOT_PRE_SETUP_COMMANDS` and `MENDER_UBOOT_POST_SETUP_COMMANDS` to keep them somewhat namespaced (like `MENDER_UBOOT_AUTO_CONFIGURE`.

drewmoseley:
Closing.  Everything except SPL here has been accepted into other PRs.  I'll get back to cleaning up the SPL handling at some point.

PR #500, opened 2018-04-18
Morty backports 

drewmoseley:


drewmoseley:
@kacf the build failures here appear to be due to the inability to locate the meta-oe-go layer for the morty branch.  Is there any way this will work?

kacf:
Oops, this was removed by mistake during the rewrite of the build script. Simple fix: https://github.com/mendersoftware/mender-qa/pull/183

drewmoseley:
@kacf I managed to get all the acceptance tests to pass here. I seem to recall discussing with you that integration tests would be a bit dicey. Is it worth spending time on that for the older branches?

Also, I added https://github.com/mendersoftware/meta-mender/pull/500/commits/625e79bb02a92600d6c51627a7cde2b7ddfb6b1c to make it work better when running the tests in my home environment.  Any concerns? I'll move it to master once this is complete.

tests/acceptance/fixtures.py
```diff
@@ -253,11 +253,11 @@ def cleanup_test_build():
     local_conf = os.path.join(build_dir, "conf", "local.conf")
     local_conf_orig = local_conf + ".orig"
     with open(local_conf, "a") as fd:
-        fd.write('SSTATE_MIRRORS = " file://.* file://%s/sstate-cache/PATH"\n' % os.environ['BUILDDIR'])
+        fd.write('SSTATE_MIRRORS = " file://.* file://%s/PATH"\n' % bitbake_variables['SSTATE_DIR'])
+        fd.write('SSTATE_DIR = "${TOPDIR}/sstate-cache"\n')
+        fd.write('DL_DIR = "%s"\n' % bitbake_variables['DL_DIR'])
```
kacf: I'm probably missing something obvious, but I don't understand how it is supposed to work. Isn't the `SSTATE_DIR` assignment a no-op, IOW it is the same location as the default? `${TOPDIR}` refers to the build directory, which will be the test-specific directory in this case right?

Otherwise I'm all for the cause! :-)

kacf:
> I managed to get all the acceptance tests to pass here. I seem to recall discussing with you that integration tests would be a bit dicey. Is it worth spending time on that for the older branches?

Yeah, although the status of your branch isn't too bad. In fact they all passed except for one test. I remember that one being problematic, so I wouldn't worry too much about it. Looks like your branch is in pretty good shape!

tests/acceptance/fixtures.py
drewmoseley: Indeed.  I am fairly certain I intended to use 'bitbake_variables['SSTATE_DIR']' here so it will properly pull that, if it is modified (which I do locally to share one sstate directory across all my branches and builds).  I'll take a look and see if I can determine whether this is just an oversight on my part.

drewmoseley: Actually, I remember now. The tests use the existing SSTATE_DIR has an SSTATE_MIRROR to reuse them without polluting them with testing-based artifacts.  I think I put SSTATE_DIR here out of a sense of completion but you're right, it should be left at the default in case it changes.

drewmoseley:
OK.  I cleaned up the SSTATE_DIR stuff in https://github.com/mendersoftware/meta-mender/pull/500/commits/dfffecd0334027e9f1a0a75bdee6bfc4c3a99555.  If this test run succeeds I'll finally be able to call this done.

PR #505, opened 2018-04-25
mender: Cleanup IMAGE_FSTYPES.

drewmoseley:
Since we now explicitly use ARTIFACTIMG_FSTYPE instead of scanning
IMAGE_FSTYPES (and have sensible defaults) it should no longer be
necessary to explicitly set these.

Changelog: Title


drewmoseley:
Assuming https://github.com/mendersoftware/mender-docs/pull/397 is correct, then I think this cleanup may be correct.  I'm not sure about the conf files under the tests directory since I don't know how they are used but they don't seem to need the settings either.

kacf:
x86 is red, but some kind of error has sneaked into master that fails the same five tests on every PR. I have to fix this anyway, so I'm going to merge this, since all the rest is green.

PR #511, opened 2018-05-07
added layer dependency on mender layer

oleorhagen:
Changelog: Title


oleorhagen:
Unless any objections, I will merge this now.

mirzak:
LGTM with one nit-pick.

I am not sure if this qualifies as something that should go in to the change-log.

mirzak:
Looking at previous change-log entries, I would say this is OK as-is.

PR #521, opened 2018-05-31
Misc changes

drewmoseley:


mirzak:
Nice one, seen those dangling links as well but never got around to investigate and fix.

drewmoseley:
D'oh.  Once again forgot the changelog.  My main repos have a git hook that won't allow a commit without one but evidently that doesn't check on "git am" as I exported this patch from a temp repo.  It should be clean now.

PR #525, opened 2018-06-05
Accept MAC address for qemu through env var

0lmi:


kacf:
I went ahead and fixed it. Resubmitted in #526.

PR #528, opened 2018-06-18
QEMU: Limit client setup steps to run on first boot only.

kacf:
Changelog: Fix occasional inode corruption issue when re-launching a
previously launched and saved QEMU image.

Changelog: Title


kacf:
Known spurious failure, everything green otherwise.

PR #529, opened 2018-06-18
More board-generic changes from recent ports

drewmoseley:
Most of these commits will be submitted against master or have already been.


drewmoseley:
@kacf I think the latest commits should resolve the test issues.

drewmoseley:
@kacf I think this is ready to go now.  I have most of these commits on the other branches too and will be submitting when my local testing completes.  But I wanted to wrap up here since I have the Dragonboard and Gumstix integrations on morty as well.  Any concerns with anything I've done here?

drewmoseley:
Ahh. Crap.  I got so excited that the tests passed that I forgot I had not yet removed that.  Thanks for the doublecheck.

drewmoseley:
OK.  I guess we can ignore issues with qemux86-64 since that is explicitly _not_ available on this branch.  Merging.

kacf:
> OK. I guess we can ignore issues with qemux86-64 since that is explicitly not available on this branch. Merging.

Yep!

PR #531, opened 2018-06-18
Misc cleanup prs

drewmoseley:
This pulls together a number of items from my other open PRs that will not, I don't believe, have issues with CI failures.  Hopefully if that is the case we can get these merged in and then I can go back to the others and try and get their issues squared away.


tests/acceptance/common.py
```diff
@@ -345,19 +345,13 @@ def get_bitbake_variables(target, env_setup="true", export_only=False):
...
+        if ret.get('MENDER_TEST_MACHINE') is not None:
+            ret['MACHINE'] = ret.get('MENDER_TEST_MACHINE')
         else:
-            local_fd = open(os.path.join(os.environ['BUILDDIR'], "conf", "local.conf"))
-            for line in local_fd:
-                match = re.match('^ *MACHINE *\?*= *"([^"]*)" *$', line)
-                if match is not None:
-                    ret['MACHINE'] = match.group(1)
-            local_fd.close()
+            raise Exception("Could not determine MACHINE or MENDER_TEST_MACHINE value.")
```
kacf: Smart!

meta-mender-qemu/scripts/mender-qemu
```diff
@@ -51,8 +51,17 @@ QEMU_ARGS=""
...
+# Process them in the order specified in bitbake.conf.
+# Note: it's possible this will miss some other possible configurations
+# but we don't want to have to run a full bitbake parse here.
+
+# Determine MACHINE if set with '='
+EXTRACTED_MACHINE="${EXTRACTED_MACHINE:-$(cat $BUILDDIR/conf/site.conf $BUILDDIR/conf/auto.conf $BUILDDIR/conf/local.conf 2>/dev/null | grep 'MACHINE *=' | grep -v '#' | tail -1 | cut -d= -f2 | tr -d \" | tr -d \ )}"
+# Determine MACHINE if set with '?='
+EXTRACTED_MACHINE="${EXTRACTED_MACHINE:-$(cat $BUILDDIR/conf/site.conf $BUILDDIR/conf/auto.conf $BUILDDIR/conf/local.conf 2>/dev/null | grep 'MACHINE *?=' | grep -v '#' | tail -1 | cut -d= -f2 | tr -d \" | tr -d \ )}"
+# Determine MACHINE if set with '??='
+EXTRACTED_MACHINE="${EXTRACTED_MACHINE:-$(cat $BUILDDIR/conf/site.conf $BUILDDIR/conf/auto.conf $BUILDDIR/conf/local.conf 2>/dev/null | grep 'MACHINE *??=' | grep -v '#' | tail -1 | cut -d= -f2 | tr -d \" | tr -d \ )}"
```
kacf: > the only truly reliable way to figure it out is to run bitbake and let it parse the metadata and I don't think we want to do that on every invocation of this script.

Actually, we already do. Check out the `eval` above this section. I bet you can replace the `MACHINE` variable there with the new `MENDER_TEST_MACHINE` variable you added.

And following that, perhaps it should be renamed to `MENDER_MACHINE`, since it has uses outside of testing too.

tests/acceptance/test_update.py
```diff
@@ -142,7 +142,7 @@ def test_broken_image_update(self, bitbake_variables):
             subprocess.call("dd if=/dev/zero of=image.dat bs=1M count=0 seek=16", shell=True)
             subprocess.call("mender-artifact write rootfs-image -t %s -n test-update -u image.dat -o image.mender" % image_type, shell=True)
             put("image.mender", remote_path="/var/tmp/image.mender")
-            run("mender -rootfs /var/tmp/image.mender")
+            run("mender -rootfs /var/tmp/image.mender -f")
```
kacf: I'm curious why this is needed?

drewmoseley: If your build has state scripts in it, then the command line installation will fail.  My normal config adds some so I had to add this to run the tests locally.

drewmoseley:
@kacf I adapted the MENDER_MACHINE and mender-qemu patches per your recommendations.  I am running local tests now but since the MENDER_MACHINE has to be exported(*) rather than simply set as a standard variable, much of the sstate cache is invalidated.  This may be ready to go apart from the BBB build failure.

(*) I think this was due to some of the tests that run bitbake on mender-test-dependencies and other things rather than a full image file; in those cases the variable evidently is not visible if it is not exported.

PR #536, opened 2018-06-26
MEN-1657: Fix incorrect memory size in vexpress-qemu devicetree file.

kacf:
Our emulator script uses 256M, so the device tree file should mirror
that.

Changelog: Fix Linux kernel hanging when loaded via U-Boot/UEFI/GRUB
on vexpress-qemu*.


mirzak:
You could have passed `mem=` Linux kernel argument to limit the accessible memory, but do not know if that makes it any easier.

kacf:
I think my solution is better. It corrects the problem at the source, instead of just overriding it.

Interestingly, Linux will actually detect the correct size itself, as can be seen by running Linux through U-Boot, but I guess that detection happens later in the boot process.

PR #540, opened 2018-06-29
Overo

drewmoseley:
This is the Gumstix Overo board. This is only tested on the Morty branch and at the time of development the upstream BSP only supported that branch so for now it should stay only on Morty.

### DIFF
--- a/meta-mender-overo/conf/layer.conf
+++ b/meta-mender-overo/conf/layer.conf
@@ -1,0 +1,13 @@
+# Board specific layer configuration for meta-mender
+# Copyright 2018 Northern.tech AS
+
+# We have a conf and classes directory, add to BBPATH
+BBPATH .= ":${LAYERDIR}"
+
+# We have recipes-* directories, add to BBFILES
+BBFILES += "${LAYERDIR}/recipes-*/*/*.bb \
+	${LAYERDIR}/recipes-*/*/*.bbappend"
+
+BBFILE_COLLECTIONS += "mender-overo"
+BBFILE_PATTERN_mender-overo = "^${LAYERDIR}/"
+BBFILE_PRIORITY_mender-overo = "9"
--- a/meta-mender-overo/recipes-bsp/u-boot/files/fw_env.config.default
+++ b/meta-mender-overo/recipes-bsp/u-boot/files/fw_env.config.default
@@ -1,0 +1,8 @@
+# Configuration file for fw_(printenv/saveenv) utility.
+# Up to two entries are valid, in this case the redundant
+# environment sector is assumed present.
+# Notice, that the "Number of sectors" is ignored on NOR.
+
+# MTD device name       Device offset   Env. size       Flash sector size       Number of sectors
+/dev/mtd2               0x00000          0x20000         0x20000                 1
+/dev/mtd2               0x20000          0x20000         0x20000                 1
--- a/meta-mender-overo/recipes-bsp/u-boot/patches/0002-u-boot-Mender-integration-for-Overo-board.patch
+++ b/meta-mender-overo/recipes-bsp/u-boot/patches/0002-u-boot-Mender-integration-for-Overo-board.patch
@@ -1,0 +1,112 @@
+From 5152008081454cefbacc1d57e082f56f47459df2 Mon Sep 17 00:00:00 2001
+From: Drew Moseley <drew.moseley@northern.tech>
+Date: Tue, 13 Feb 2018 14:43:19 -0500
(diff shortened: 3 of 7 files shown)


meta-mender-overo-sdcard/conf/layer.conf
```diff
@@ -0,0 +1,13 @@
...
+# We have a conf and classes directory, add to BBPATH
+BBPATH .= ":${LAYERDIR}"
+
+# We have recipes-* directories, add to BBFILES
+BBFILES += "${LAYERDIR}/recipes-*/*/*.bb \
+	${LAYERDIR}/recipes-*/*/*.bbappend"
+
+BBFILE_COLLECTIONS += "mender-overo"
+BBFILE_PATTERN_mender-overo = "^${LAYERDIR}/"
+BBFILE_PRIORITY_mender-overo = "9"
```
mirzak: Depends on meta-overo?

meta-mender-overo-sdcard/recipes-bsp/u-boot/u-boot-overo.inc
```diff
@@ -0,0 +1,16 @@
+FILESEXTRAPATHS_prepend_overo := "${THISDIR}/patches:"
+
+# Overo board is using 2016.11 u-boot, need to apply a patch that matches our version
+SRC_URI_remove_overo += " file://0003-Integration-of-Mender-boot-code-into-U-Boot.patch"
+SRC_URI_append_overo += " file://0003-U-Boot-pre-2017.03-integration-of-Mender-boot-code-i.patch"
+
+# This will likely not work at the moment.  The default assumes the UBoot environment
+# is in NAND.  For proper SDCard based Mender updates, that should probably be moved
+# to MMC so we can easily update it.
+# UBI is currently tested and working; this is being kept for archival purposes.
```
mirzak: This comment confused me :). The commit message says that it is SD card integration? And you seem to have moved env to MMC based on the u-boot patch?

meta-mender-overo-sdcard/recipes-bsp/u-boot/files/fw_env.config.default
```diff
@@ -0,0 +1,8 @@
+# Configuration file for fw_(printenv/saveenv) utility.
+# Up to two entries are valid, in this case the redundant
+# environment sector is assumed present.
+# Notice, that the "Number of sectors" is ignored on NOR.
+
+# MTD device name       Device offset   Env. size       Flash sector size       Number of sectors
+/dev/mtd2               0x00000          0x20000         0x20000                 1
+/dev/mtd2               0x20000          0x20000         0x20000                 1
```
mirzak: You set `CONFIG_ENV_IS_IN_MMC` in the U-boot patch, does this mean that this file is not used?

meta-mender-overo-sdcard/recipes-bsp/u-boot/u-boot-overo.inc
kacf: +1, also confused! Maybe it can be removed, since you are including the UBI based integration as well?

drewmoseley: It's both.   I started the SDCard integration but didn't finish it but figured it may be worth archiving here.

drewmoseley: I'll cleanup the commit logs a bit to make it more evident.

meta-mender-overo-sdcard/conf/layer.conf
drewmoseley: I can do that but we don't have any LAYERDEPENDS in the morty branch at all.  @kacf preference?

meta-mender-overo-sdcard/recipes-bsp/u-boot/files/fw_env.config.default
drewmoseley: It's probably just something that didn't get fully fleshed out when I switched to UBI based integration.  I can drop but if you guys prefer but I thought it would be good to keep it for future.  Thoughts?

drewmoseley:
OK.  For now I removed the SD Card stuff to see if I can get past the CI issues.  Let me know how you guys think we should proceed with the SD Card stuff.

meta-mender-overo-sdcard/conf/layer.conf
kacf: No need here then, but something to keep in mind if this is forward ported.

PR #543, opened 2018-07-06
FIX: Inconsistent rootfs_size calculation

oleorhagen:
Fix as per Alberto Spin's email on the mailing list.

Changelog: Title


oleorhagen:
Already fixed

PR #548, opened 2018-07-10
Wip pyro

drewmoseley:
This pulls in all my latest patches that have been merged into master. I still need to get the latest bits from others in rocko that are relevant but I figure I'll do that in a second pass.


drewmoseley:
I assume the qemux86-64 failure is ok here since it is not available on this branch.  Anything obviously wrong with this catch-up PR?

PR #550, opened 2018-07-12
Add canary value to U-Boot env to catch bootloader/user-space mismatch.

kacf:
The way it works is that we store the 'mender_check_saveenv_canary=1'
variable in the default U-Boot environment. We do not store the
corresponding 'mender_saveenv_canary=1' value in the default
environment, but instead we write this during the boot process. When
the Mender client finds the former variable, it will look for the
second value as well, and produce an error if it is not present. If
this happens it is an indication that the U-Boot boot loader and the
user space tools do not agree on the location of the environment. It
will also catch cases where mender_setup is not run.

Changelog: Title


kacf:
https://github.com/mendersoftware/mender/pull/294.

PR #553, opened 2018-07-13
Support for U-Boot/UEFI/GRUB combination

kacf:
The main commit is this one

```
commit 286c49baef204277d400ea3fef3e5deb71a31954
Date:   Wed Jul 11 10:14:15 2018

    Implement support for generic ARM UEFI booting using U-Boot and GRUB.
    
    Changelog: Introduce support for a standardized boot method on ARM
    using UEFI and GRUB. The UEFI boot standard is fully supported on ARM
    in theory, but few board manufacturers implement it in practice.
    Therefore U-Boot is still utilized, but acts only as a UEFI loader,
    and hence no U-Boot integration is needed. All boot scripts are then
    handled by GRUB, which needs no patching.
    
    MEN-1595, MEN-1659
    
```

with various supporting commits to complete it.


kacf:
@drewmoseley: FYI, the U-Boot/UEFI/GRUB combo has been merged now, so that's another platform to ignore for rocko and older.

PR #556, opened 2018-07-19
test-update.py now passes with a read-only rootfs 

oleorhagen:


tests/meta-mender-ci/recipes-testing/read-only-rootfs/mender_%.bbappend
```diff
@@ -0,0 +1,9 @@
+FILES_${PN} += "/data/etc/mender"
+# Run after all mender recipes are finished.
+do_install_append() {
+    mkdir -p ${D}/data/etc/mender
```
mirzak: Is the intention to move this to core or demo layer as well? If that is the case I have two suggestions :).

There is a `IMAGE_FEATURE` called `read-only-rootfs` which we could use to do this conditionally, meaning that this could actually reside in meta-mender-core and applied based on if that feature is set. Are you using the `read-only-rootfs` feature when building a read-only rootfs image?

Secondly, we have `/data/mender` directory where all mender related stuff is stored. I would suggest that these files also go there to make it easy to separate what files are Mender related and what is user data.

kacf: No, this is meant to be used only for tests. The reason it's needed is that the tests need to make modifications to `/etc/mender/mender.conf`, but users are expected to do this via an update. So making this generic is not of great concern here, since we will be the only users of it.

oleorhagen:
@kacf 
the x86 fails are both in TestBuild. Not entirely sure what this is about...
The vexpress fail is on me though, I only ran it on x86 locally

kacf:
@oleorhagen: It's almost certainly because we try to look in `/etc/mender/mender.conf` for some values, and `debugfs` cannot follow symlinks (how would it know about mounts?). See [here](https://github.com/mendersoftware/meta-mender/blob/master/tests/acceptance/test_build.py#L123). Should be easy to fix, just point it to the new location.

tests/acceptance/fixtures.py
```diff
@@ -336,6 +336,11 @@ def prepared_test_build_base(request, bitbake_variables):
         build_dir = tempfile.mkdtemp(prefix="test-build-", dir=os.environ['BUILDDIR'])
 
     local_conf = os.path.join(build_dir, "conf", "local.conf")
+
+    # All acceptance tests must handle a read-only filesystem.
+    add_to_local_conf({'local_conf': os.path.join(build_dir, "conf", "local.conf")},
```
oleorhagen: @kacf will this be the appropriate way of adding the read-only-rootfs functionality to all the acceptance tests?

tests/meta-mender-ci/recipes-testing/read-only-rootfs/mender_%.bbappend
```diff
@@ -0,0 +1,12 @@
+
+IMAGE_FEATURES_append = " read-only-rootfs"
+
+FILES_${PN} += "/data/etc/mender"
+# Run after all mender recipes are finished.
+do_install_append() {
+    mkdir -p ${D}/data/etc/mender
+    mv ${D}/etc/mender/mender.conf ${D}/data/etc/mender
+    mv ${D}/etc/mender/server.crt ${D}/data/etc/mender
+    ln -s /data/etc/mender/mender.conf ${D}/etc/mender/mender.conf
```
kacf: @oleorhagen: I have hard proof now. This link exists in the Docker client image, which means there is definitely something wrong in the build, and the link somehow survives the acceptance test stage and gets into the integration tests. I still have not found out why though.

kacf: Btw, even stranger is that the link exists, but its `/data` counterpart does not, so only half of the changes survived. Could it be that the `.mender` artifact survives with the link change (what is on the rootfs), and the `.sdimg` does not (what is in `/data`)? Mysterious...

kacf:
Merged in #763.

PR #557, opened 2018-07-20
Misc

drewmoseley:


meta-mender-core/recipes-core/systemd/systemd_%.bbappend
```diff
@@ -4,5 +4,4 @@
 # in the future and an fsck will be done.  Setting this to 0 results
 # in an epoch of January 1, 1970 which is detected as an invalid time
 # and the fsck will be skipped.
-EXTRA_OECONF += "--with-time-epoch=0"
-EXTRA_OEMESON += " -Dtime-epoch=0"
+PACKAGECONFIG_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-install', ' time-epoch', '', d)}"
```
kacf: Check for `mender-systemd` instead, perhaps?

meta-mender-core/recipes-bsp/u-boot/patches/0002-Generic-boot-code-for-Mender.patch
```diff
@@ -163,12 +163,18 @@ index 0000000..a1996ab
 +    "mender_mtd_ubi_dev_name=" MENDER_MTD_UBI_DEVICE_NAME "\0"
 +#else
 +# define MENDER_UBI_SETTINGS
++# ifdef CONFIG_LOADADDR
++#  define MENDER_DEFAULT_LOADADDR "kernel_addr_r=" __stringify(CONFIG_LOADADDR) "\0"
++# else
++#  define MENDER_DEFAULT_LOADADDR
++# endif
```
kacf: I would still like to try to push [this patch](https://lists.denx.de/pipermail/u-boot/2018-July/334155.html) (and possibly related patches) a bit more on the U-Boot mailing list before resorting to this (a reply on the list from you might help get the activity going btw). I'm a bit uneasy with this patch because many boards define `kernel_addr_r` on their own, and in this case we'll have two `kernel_addr_r` entries in the environment. It's not clear what will happen then. What do you think?

Is there a specific board that prompted this? Why not just use `loadaddr` directly, which `CONFIG_LOADADDR` eventually resolves to in U-Boot itself? At least the auto patcher will use this if it's available.

drewmoseley: There definitely _was_ a board that necessitated this change but at the moment I don't recall specifically.  I _think_ it was the i.MX7D SaberSD board which I was playing with.  I'll take another look when I get some time.  In the meantime, I'll at least try and figure out what is wrong with the tests here.

drewmoseley:
I updated the epoch setting per your suggestion.

drewmoseley:
Rebased and removed kernel_addr_r commit.

kacf:
Thanks!

PR #558, opened 2018-07-21
Morty backports

drewmoseley:


drewmoseley:
@kacf any concerns with anything here? test_ssl_only is a known failure right?

PR #559, opened 2018-07-21
Pyro backports

drewmoseley:


drewmoseley:
@kacf any concerns?  qemux86-64 is not available on pyro so this failure here is unsurprising.

PR #560, opened 2018-07-23
Rocko backports

drewmoseley:


drewmoseley:
@kacf any concern here?

drewmoseley:
@kacf I'm not able to reproduce this build failure locally and the failure log doesn't look related to anything I've changed.  I tried to kick a new build here but it doesn't seem to be running.  Can you help?

> 17:20:03 | ln: /home/jenkins/workspace/yoctobuild/build-vexpress-qemu-flash/tmp/work/x86_64-linux/mender-artifact-native/master-git999-r0/mender-artifact-master-git999/src: cannot overwrite directory
> 17:20:03 | WARNING: exit code 1 from a shell command.
> 17:20:03 | ERROR: Function failed: do_configure (log file is located at /home/jenkins/workspace/yoctobuild/build-vexpress-qemu-flash/tmp/work/x86_64-linux/mender-artifact-native/master-git999-r0/temp/log.do_configure.8001)
> 17:20:03 NOTE: recipe mender-artifact-native-master-git999-r0: task do_configure: Failed

drewmoseley:
@kacf is Jenkins still having issues? I pushed a new commit here and it has not run the new tests yet.

drewmoseley:
I was able to get the testing to run using release_tool.py

drewmoseley:
Not sure if I enabled too many tests when I ran release_tool.  Most of this succeeded but it looks like there were issues with the hardware-based acceptance tests.

kacf:
> is Jenkins still having issues? I pushed a new commit here and it has not run the new tests yet.

Sorry about that. In fact they did run, but a bug was introduced yesterday which made the jobs fail very early, before the call to the Github hook. So the status was not visible here. The release tool worked since it used an older, working version. It has been fixed now, so all jobs should work again.

> Not sure if I enabled too many tests when I ran release_tool. Most of this succeeded but it looks like there were issues with the hardware-based acceptance tests.

Hardware acceptance tests have not worked for a long time. It's why we started the SDMUX project. Basically we don't have the capacity to maintain our previous setup and it was dismantled when we moved office. I suppose the parameters could be removed...

Since everything else worked, I would consider this passed.

PR #563, opened 2018-07-28
cherry-picks (aarch64 and rpi kernel image) 

mirzak:


mirzak:
@kacf , the rename trick did not re-build or it failed again?

PR #565, opened 2018-07-31
Re add old mender recipes

drewmoseley:


drewmoseley:
These are the same known test failures discussed in https://github.com/mendersoftware/meta-mender/pull/558.

Merging

PR #567, opened 2018-08-02
Fixed do_mender_tar_src function to work with external sources

alfrunes:
This function fails when working on external sources (local on your
machine; e.g. using `devtool`).

Changelog: none


kacf:
It's a bit more complicated than that, due to the complex interaction between `u-boot` and `u-boot-fw-utils-mender-auto-provided`. Using `${S}` directly will generate a tar file with absolute paths in it, which won't work correctly.

But this has been bugging me too, so I took a stab at it. Here is my attempt at it: #569.

PR #568, opened 2018-08-02
Switch GRUB/BIOS combination to new machine name to avoid conflict.

kacf:
Machine names that are equal should reflect images that are
compatible, but this is not true for GRUB/UEFI vs GRUB/BIOS, due to
the mount point of the environment being different (/boot/efi vs
/boot/grub). So switch to a new machine name to distinguish between
the two configurations.

Changelog: Switch GRUB/BIOS combination to new machine name to avoid
conflict: mender-qemux86-64-bios. This only affects the QEMU builds.


kacf:
Should be merged together with https://github.com/mendersoftware/mender-qa/pull/204.

PR #569, opened 2018-08-03
Enable EXTERNALSRC to be used with u-boot-fw-utils-mender-auto-provided.

kacf:
Changelog: Title


kacf:
@alfrunes: Does it work for you?

PR #576, opened 2018-08-13
Several fixes to make partition selection better

kacf:

PR #577, opened 2018-08-13
Change default storage on x86 to memory card (mmcblk0).

kacf:
It appears that `/dev/hda` is rather uncommon, and quite specific to
QEMU since its IDE driver (which corresponds to `/dev/hda`) is
obsolete in modern hardware, and hence should only be used for
testing.

This is building on #576, only the top commit is new.

### DIFF
--- a/meta-mender-core/classes/mender-helpers.bbclass
+++ b/meta-mender-core/classes/mender-helpers.bbclass
@@ -32,7 +32,7 @@ get_uboot_device_from_device() {
 
 get_grub_device_from_device_base() {
     case "$1" in
-        /dev/[sh]da)
+        /dev/[sh]d[a-z])
             dev_number=${1#/dev/[sh]d}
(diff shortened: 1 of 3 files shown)


kacf:
@mirzak, @drewmoseley: What do you think about [this](https://github.com/mendersoftware/meta-mender/pull/577/commits/d4a9bb0880efde202ea6d6770967a9f7a1025b3a)? This is inspired by several people, including @estenberg and Arnstein Kleven from the mailing list, having trouble integrating because the storage device is wrong. `/dev/mmcblk0` is not guaranteed to be correct either, but the chances are higher than with `/dev/hda`.

I'm even considering cherry-picking this to older branches, but I think we may need to announce this on the mailing list, because this will break any builds that rely on the current default.

estenberg:
This makes sense to me at least, although the minnowboard x86 presented it as mmcblk2.

PR #580, opened 2018-08-17
Add some debug functionality to GRUB booting process.

kacf:
To use it, enable either or both of `debug-log` and `debug-pause` in
`PACKAGECONFIG` for `grub-mender-grubenv`. The former enables debug
logging in GRUB, which can be tweaked further by setting the
`DEBUG_LOG_CATEGORY` variable. The latter pauses the boot process at
strategic points during the boot, so that screen output can be
captured before it is cleared or scrolls by.

Changelog: Commit

Also some minor cleanup fixes.


kacf:
Thanks!

PR #581, opened 2018-08-17
Switch image names from containing MACHINE to containing DEVICE_TYPE.

kacf:
Changelog: Switch image names from containing `MACHINE` to containing
`MENDER_DEVICE_TYPE`. So for example, if you have a build for the
raspberrypi3 machine type, with device type of "prod_rpi3", the image
will now be called `core-image-minimal-prod_rpi3.sdimg` instead of
`core-image-minimal-raspberrypi3.sdimg`.

The main reason for doing this change is that MACHINE is not enough to
separate images from each other when multiple boot loader
configurations are involved. For example vexpress-qemu can be booted
both with U-Boot and with GRUB, and these images are not compatible
with each other. To make sure that images have distinct names, this
change is necessary.

### DIFF
--- a/meta-mender-core/classes/mender-setup.bbclass
+++ b/meta-mender-core/classes/mender-setup.bbclass
@@ -59,6 +59,9 @@ MENDER_BOOT_PART_FSTYPE_DEFAULT = "auto"
 MENDER_DEVICE_TYPE ??= "${MENDER_DEVICE_TYPE_DEFAULT}"
 MENDER_DEVICE_TYPE_DEFAULT = "${MACHINE}"
 
+# To tell the difference from a beaglebone-yocto image with only U-Boot.
+MENDER_DEVICE_TYPE_DEFAULT_beaglebone-yocto_mender-grub = "${MACHINE}-grub"
+
 # Space separated list of device types compatible with the built update.
 MENDER_DEVICE_TYPES_COMPATIBLE ??= "${MENDER_DEVICE_TYPES_COMPATIBLE_DEFAULT}"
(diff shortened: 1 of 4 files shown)


meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -152,6 +155,12 @@ MENDER_UBOOT_POST_SETUP_COMMANDS_DEFAULT = ""
 IMAGE_INSTALL_append = " mender"
 IMAGE_CLASSES += "mender-part-images mender-ubimg mender-artifactimg mender-dataimg"
 
+# Originally defined in bitbake.conf. We define them here so that images with
+# the same MACHINE name, but different MENDER_DEVICE_TYPE, will not result in
+# the same image file name.
+IMAGE_NAME = "${IMAGE_BASENAME}-${MENDER_DEVICE_TYPE}-${DATETIME}"
+IMAGE_LINK_NAME = "${IMAGE_BASENAME}-${MENDER_DEVICE_TYPE}"
```
kacf: @mirzak: What we talked about! :-)

mirzak: Yeah, that was easy :)

drewmoseley:
Will this need to be backported to any older branches?

kacf:
> Will this need to be backported to any older branches?

It needs to be backported to sumo because [it affects our image uploads](https://tracker.mender.io/browse/MEN-2047), however, I think for older branches it's not necessary. And in fact, since it changes behavior I would recommend not to.

For sumo I will try to backport this in a way that only enables it specifically for our builds and no one else's. This way we solve our problem, but don't change behavior behind the users' backs. I will get back to this.

PR #586, opened 2018-08-22
boot-script recipe for easily demoing OTA updates

estenberg:
can be used to blink lights, show picture, video, etc.

### DIFF
--- a/meta-mender-demo/recipes-core/images/core-image-full-cmdline.bbappend
+++ b/meta-mender-demo/recipes-core/images/core-image-full-cmdline.bbappend
@@ -1,1 +1,0 @@
-IMAGE_INSTALL_append = " hello-mender"
--- a/meta-mender-demo/recipes-mender/boot-script/boot-script.bb
+++ b/meta-mender-demo/recipes-mender/boot-script/boot-script.bb
@@ -1,0 +1,21 @@
+LICENSE = "Apache-2.0"
+
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+SRC_URI = "file://boot-script.service file://boot-script.sh file://LICENSE"
+
+LIC_FILES_CHKSUM = "file://LICENSE;md5=0ea4e253cc22ddc22117b9796e5ce5b7"
+FILES_${PN} += "${sbindir}/boot-script.sh ${systemd_unitdir}/system/boot-script.service"
+
+inherit systemd
+
+SYSTEMD_SERVICE_${PN} = "boot-script.service"
+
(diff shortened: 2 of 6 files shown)


estenberg:
Tested with blinking lights demo on minnowboard, but can be useful for most demo cases.

Is there a better way to "include" it than in the file core-image-full-cmdline.bbappend?
Would be nice if it was not specific to bitbake build target.

meta-mender-demo/recipes-mender/boot-script/boot-script.bb
```diff
@@ -0,0 +1,19 @@
+LICENSE = "Apache-2.0"
+
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+SRC_URI = "file://boot-script.service file://boot-script.sh file://LICENSE"
+
```
kacf: Try to insert

```
S = "${WORKDIR}"
```

I think that will get rid of the `LICENSE` error you're getting in the Jenkins build.

meta-mender-demo/recipes-core/images/core-image-full-cmdline.bbappend
```diff
@@ -1 +1 @@
-IMAGE_INSTALL_append = " hello-mender"
+IMAGE_INSTALL_append = " hello-mender boot-script"
```
kacf: > Is there a better way to "include" it than in the file core-image-full-cmdline.bbappend?
> Would be nice if it was not specific to bitbake build target.

Yes, I believe so. In fact, I think we should move both, so open the `meta-mender-demo/recipes-core/packagegroups/packagegroup-core-boot.bbappend`, and add it like this:

```
RDEPENDS_${PN}_append = " hello-mender boot-script"
```

and then just delete the `core-image-full-cmdline.bbappend` file.

estenberg:
Thanks @kacf ! Updated.

meta-mender-demo/recipes-mender/boot-script/boot-script.bb
estenberg: The secret way to get Jenkins to be happy? I like it ;)

kacf: Hehe, yes, it's a long and complicated reason, so you just have to trust me that it *does* make sense! :-)

kacf:
The usual spurious failures. Those have become really bad lately...

kacf:
I will cherry-pick this later as part of a bigger sumo update.

PR #587, opened 2018-08-23
Forward porting some changes from the recent Toradex work

drewmoseley:


kacf:
Another spurious failure, however it is happening semi-frequently, so I reported [MEN-2061](https://tracker.mender.io/browse/MEN-2061). Don't know if we will ever get to the bottom of it though.

kacf:
Thanks for forward porting this!

PR #589, opened 2018-08-27
MEN-2059: Add inventory tests for bootld. integration and rootfs type.

kacf:
+ some minor fixes.


kacf:
Merged as part of #590.

PR #598, opened 2018-09-12
Morty backports

drewmoseley:


drewmoseley:
@kacf any reason to expect the "Failed: failed to connect to production env. using SSL" issue here? Is that something that can be ignored in the older branches?

kacf:
I don't think I've seen that before... but I can't find that message in any of the two failed jobs (I guess the qemux86-64 is just because of missing support). Maybe this just needs a rerun?

drewmoseley:
Yeah, I did a force-push to this branch to have it run again and this time it didn't get the error.  I'll assume it's an artifact of the older branch and just merge.

PR #600, opened 2018-09-12
Rocko backports

drewmoseley:


kacf:
Hmm, it is relevant, in the sense that the bug it fixes is present on rocko as well. However, I don't think it exhibits on rocko for any of the tests that we run (it was found during the GRUB on ARM work).

If it's causing problems, I *think* it's ok to leave it out. What happens without it is that we produce a slightly corrupted UEFI image, which has its backup header placed in the wrong sector. But I think it's only U-Boot that cares about this, and since there is no UEFI on ARM on rocko, it should not affect that branch.

As for your current build error, it might be because the duplicated grub-efi recipe was taken from sumo, which might be incompatible with rocko. What you can try is:

```
cd poky/meta-mender
git checkout sumo
diff -u ../meta/recipes-bsp/grub/grub-efi_2.02.bb meta-mender-core/recipes-bsp/grub/grub-efi_2.02.bb > grub-efi.patch
git checkout rocko
cp ../meta/recipes-bsp/grub/grub-efi_2.02.bb meta-mender-core/recipes-bsp/grub/grub-efi_2.02.bb
patch -p1 < grub-efi.patch
```

so that you get the basic recipe from rocko, but with the changes that I made.

drewmoseley:
OK.  For now I'll leave out the padding patch.

I did the diff and patch but the resultant file was unchanged.  It looks like the rocko and sumo versions are identical.

drewmoseley:
I added https://github.com/mendersoftware/meta-mender/pull/600/commits/6029b3921728ac17a5ae062e8ce5f0303ef46d61 which simply removed the RDEPEND. The grub-bootconf packages don't exist on Rocko.

PR #603, opened 2018-09-18
mender: Enable systemd in mender-systemd FEATURE handling.

drewmoseley:
Changelog: Title

### DIFF
--- a/meta-mender-core/classes/mender-setup-systemd.inc
+++ b/meta-mender-core/classes/mender-setup-systemd.inc
@@ -2,9 +2,7 @@
 # boot process while looking for devices. This is almost always because the
 # kernel feature CONFIG_FHANDLE is not enabled.
 
-python() {
-    if bb.data.inherits_class('nativesdk', d) or bb.data.inherits_class('native', d):
-        return
-    if bb.utils.contains('DISTRO_FEATURES', 'mender-systemd', True, False, d) and not bb.utils.contains('DISTRO_FEATURES', 'systemd', True, False, d):
-        raise Exception("systemd is required in DISTRO_FEATURES when using mender-systemd feature. See vexpress-qemu.conf for an example of how to enable.")
-}
+DISTRO_FEATURES_append = " ${@bb.utils.contains('MENDER_FEATURES_ENABLE', 'mender-systemd', 'systemd', '', d)}"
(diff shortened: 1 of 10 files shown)


kacf:
That depends. Is the intention to enable the systemd bits whenever our `mender-systemd` feature is set? I'd very much like it, but I don't think this will work correctly, unfortunately. The thing is that the `classes/mender-setup-systemd.inc` file is included unconditionally, so this will always run. I don't think we can make it dependent on our `mender-systemd` feature, because that is itself a `DISTRO_FEATURE`, so using it to decide would be a circular dependency. Is that what you meant?

mirzak:
Me and @drewmoseley, discussed this a bit and yeah the latter mentioned functionality is the desired one.

drewmoseley:
If I make the variable settings here conditional on mender-systemd would that work?

drewmoseley:
I added https://github.com/mendersoftware/meta-mender/pull/603/commits/f0dc1c7e76dcf69203286efc4345268d8f1abdab which is similar to how I do it in my local layer.  Does that make it any more acceptable?

kacf:
That's the problem: There is no `MENDER_FEATURES` variable. We're just reusing `DISTRO_FEATURES` with `MENDER_FEATURES_ENABLE/DISABLE` as syntactic sugar because `DISTRO_FEATURES_BACKFILL_CONSIDERED` is such a horrible and unintuitive name for a variable. So all the features are simply listed in `DISTRO_FEATURES`, which I think means that you cannot test for the presence of something in `DISTRO_FEATURES` and also use that to modify `DISTRO_FEATURES`. Last time I tried that I got infinite recursion.

But who knows, maybe they have improved on this and they "converge" toward the stable variable assignment using multiple passes. You can try replacing `${@bb.utils.contains('MENDER_FEATURES', ...` with `${@bb.utils.contains('DISTRO_FEATURES', ...` in your commit and see if the build still works. If so, maybe we can use this.

drewmoseley:
@kacf locally this seems to work for me.  Making it conditional on mender-systemd being in MENDER_FEATURES_ENABLE seems to do the right thing.

drewmoseley:
Sounds good.  Closing in deference to https://github.com/mendersoftware/meta-mender/pull/608

PR #605, opened 2018-09-19
MEN-1643: removed mender-specific machine type

alfrunes:
Also refactored metadata recipe to be an append of already existing recipe
`tezi-metadata`.

As an aside I also fixed do_mender_src_tar for external sources

changelog: title


drewmoseley:
Some of this is relevant for master right?  I think this needs to be split up into separate, atomic commits and where appropriate, get submitted to master before merging here.

meta-mender-core/recipes-bsp/u-boot/u-boot-mender.inc
```diff
@@ -16,7 +16,14 @@ do_mender_tar_src() {
...
     # there are a lot of bitbake data directories.
-    tar czf ${TMPDIR}/mender-u-boot-src.tar.gz `find -maxdepth 1 -type f` git
+    if [ -n "${EXTERNALSRC}" ]; then
+        # tar u-boot sources, but transform ${S} as if it was git subdir in ${WORKDIR}
+        local tmp=${S}
+        tmp=${tmp#?}
+        tar --transform="s|${tmp}|git|" -czf ${TMPDIR}/mender-u-boot-src.tar.gz `find -maxdepth 1 -type f` ${S}
+    else
+        tar -czf ${TMPDIR}/mender-u-boot-src.tar.gz `find -maxdepth 1 -type f` git
+    fi
```
kacf: Please use [the fix that was submitted to master](https://github.com/mendersoftware/meta-mender/commit/58b19eac9a48c44b875185573f4d8af86e23db01) so we don't end up with two different approaches. This should also be in a separate commit, not mixed with the machine type fixes.

meta-mender-toradex-nxp/classes/mender_tezimg.bbclass
```diff
@@ -1,24 +1,31 @@
...
+def mender_get_image_name(image_type, d):
+    import re
+    image_name_base = d.getVar("IMAGE_BASENAME", True) + "-" + \
+                      d.getVar("MACHINE", True)
+    match = re.search("%s-[0-9]{14}\.%s" % \
+                      (image_name_base, image_type), \
+                      " ".join(os.listdir(d.getVar("IMGDEPLOYDIR", True))))
+    if match != None:
+        return match.group(0)
+    return ""
```
kacf: I think we should drop this. Using `IMAGE_NAME` should be just fine, as long as there is a dependency on the `ubimg` (which there already is below).

There is one bug in Yocto that if you build the image, and then build it again with a change that *only* affects the `tezimg`, then it won't be able to find the `ubimg`, since indeed `IMAGE_NAME` is wrong (it will refer to the previous build, which has now been moved out of the build directory). But this doesn't happen very often, and the workaround is easy (delete `tmp` or call `bitbake -fc rootfs <image_name>` first). I think it's better to solve this upstream in Yocto than to have this (quite hacky) workaround.

meta-mender-toradex-nxp/classes/mender_tezimg.bbclass
```diff
@@ -94,53 +99,51 @@ python mender_toradex_json() {
...
+                    ${IMGDEPLOYDIR}/${IMAGE_BASENAME}-${MACHINE}.ubimg
+        bbnote "Mender Tezi metadata successfully written and packed."
+        ;;
+    *)
+        bbwarn "\"ubifs\" not in IMAGE_FSTYPES. Skipping tezimg."
+    esac
+}
+
+IMAGE_FSTYPES += "mender_tezimg"
+IMAGE_FSTYPES_remove = "teziimg"
```
kacf: Since the original is called `teziimg`, maybe we should stick with two i's in our image too?

Also, do the images conflict? I would not automatically remove the `teziimg`, but rather leave this to the user.

alfrunes:
I split the PR into two commits, and cherry-picked your commit @kacf on top.

mirzak:
A couple of things.

You should probably change your target branch to "master" instead of "rocko" as you are proposing breaking changes. Unless some of your changes are bug fixes.

Would like to add that with the upcoming [meta-mender-community](https://github.com/mendersoftware/meta-mender-community/blob/rocko/meta-mender-toradex-nxp) and hub.mender.io solve the "where to put variables" problem in my opinion and that is in "templates".

Putting it in a class is nice but I would not enforce the inherit in `layer.conf`.

Note https://hub.mender.io/t/toradex-colibri-imx7-dual-512mb-colibri-imx7-solo-256mb/81, does not use `colibri-imx7-mender` and all the required changes are in a [local.conf.append](https://github.com/mendersoftware/meta-mender-community/blob/rocko/meta-mender-toradex-nxp/templates/local.conf.append) template.

alfrunes:
I added this PR to Rocko as Sumo is not going to be supported by Toradex as they only provide support even releases of Yocto, thereby I can't build and confirm that this works myself. But I created a new pull request to master with effectively the same changes:
https://github.com/mendersoftware/meta-mender/pull/630
I also excluded the layer.conf `INHERIT` and renamed the class file for setting the variables.

drewmoseley:
Regarding the comment in #643 about closing this without merging; since we have the Easy Installer images available from the Toradex repos, I'm concerned that we should have that code checked in somewhere. It's unclear if this PR is necessary to ensure we can reproduce those builds.

kacf:
I'm pretty sure the images were produced without this PR, since this was a cleanup we attempted afterwards. I haven't personally tested though.

kacf:
Superseded by #643.

PR #607, opened 2018-09-20
Issues found while doing Colibri imx7 testing

kacf:


kacf:
Not rocko, I think, because these are all related to the Flash auto-patching, which rocko doesn't have. But sumo, most definitely!

mirzak:
Unrelated to pr, 

I am curious what your build conf looked like when testing this (local.conf,layers,revisions), tried setting something up but got some parse errors. Did not spend much time on it though.

kacf:
Yes, this was most likely because `MENDER_MTDIDS` was unset, which I also ran into, and which has a horribly complicated and unintuitive error message, including parse errors and variables being empty. I'm actually about to fix that right now, in addition to much improved docs in https://github.com/mendersoftware/mender-docs/pull/508.

PR #608, opened 2018-09-20
Automatic systemd inclusion

kacf:


drewmoseley:
Nice.  I like this.  How far should we backport?  I'm all for getting this in all Yocto versions.

kacf:
Great!

My recommendation would be to keep this in master only. This isn't a bug fix, just a convenience feature, and I generally prefer stability in release branches. Sorry, I'm a bit conservative here! :-)

kacf:
You can create a ticket for it and schedule it for 1.7.0 ("Fix for" field), which coincides more or less with the next Yocto release. We scan for such tickets when we make a new release. A PR is also welcomed with you can just clearly mark as "do not merge yet" and with links between it and the ticket.

drewmoseley:
@kacf with the change I made in https://github.com/mendersoftware/mender-docs/pull/509, is there anything more you think necessary here?

kacf:
No, LGTM!

PR #610, opened 2018-09-20
Pyro backports

drewmoseley:

PR #611, opened 2018-09-20
Morty backports

drewmoseley:


kacf:
This one looks good too, but it failed in the same place as #610. Maybe there really is something here. My most likely candidate would be the `IOError` commit. Maybe you can try without that one?

drewmoseley:
@kacf does https://github.com/mendersoftware/meta-mender/pull/611/commits/de39d3b583803472d264f1ec185152d92997538c look reasonable here? The previous test run failed due to the missing fixture.

drewmoseley:
@kacf  This seems to be running tests from https://github.com/mendersoftware/mender-image-tests which are not working on the older branches.  Should we update that repo to support the older branches or is there a way to simply skip running them here?

kacf:
No, the build script should be updated to not get the tests from there if they already exist. I made a PR [here](https://github.com/mendersoftware/mender-qa/pull/216) and am rerunning in [a new job](https://mender-jenkins.mender.io/job/mender-builder/1551/).

kacf:
There is still some problem... I'll take a look, since the client team broke this.

kacf:
I'm looking at alternative ways of doing this, since our current approach is quite fragile. I hope it's ok if this is left for a day or two?

drewmoseley:
@kacf sure.  No hurry on this.

kacf:
This should be good now, @drewmoseley, the only failure is `test_ssl_only` which I believe always fails on morty, right?

drewmoseley:
@kacf in addition to test_ssl_only, in the vexpress-qemu-flash tests I get:

> 
> msg = "local() encountered an error (return code 1) while executing 'scp -C -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -P8822 ./image.mender root@localhost:.'"    def abort(msg):        """        Abort execution, print ``msg`` to stderr and exit with error status (1.)            This function currently makes use of `SystemExit`_ in a manner that is        similar to `sys.exit`_ (but which skips the automatic printing to stderr,        allowing us to more tightly control it via settings).            Therefore, it's possible to detect and recover from inner calls to `abort`        by using ``except SystemExit`` or similar.            .. _sys.exit: http://docs.python.org/library/sys.html#sys.exit        .. _SystemExit: http://docs.python.org/library/exceptions.html#exceptions.SystemExit        """        from fabric.state import output, env        if not env.colorize_errors:            red  = lambda x: x        else:            from colors import red            if output.aborts:            sys.stderr.write(red("\nFatal error: %s\n" % _encode(msg, sys.stderr)))            sys.stderr.write(red("\nAborting.\n"))            if env.abort_exception:            raise env.abort_exception(msg)        else:            # See issue #1318 for details on the below; it lets us construct a            # valid, useful SystemExit while sidestepping the automatic stderr            # print (which would otherwise duplicate with the above in a            # non-controllable fashion).            e = SystemExit(1)            e.message = msg>           raise eE           SystemExit: 1/usr/local/lib/python2.7/dist-packages/fabric/utils.py:62: SystemExit
> --
> 
> 
>

kacf:
That is very strange, I didn't get that when I ran it. It looks a bit like this was a problem with the build slave itself (scp missing??). I started a new run of just that platform [here](https://mender-jenkins.mender.io/job/mender-builder/1629/). Hopefully it will recover by itself.

kacf:
Oh, I know why it's failing now: You need [this commit](https://github.com/mendersoftware/meta-mender/commit/9b7036a48b80e9a33f7120d44e94829fc2b4c6f0).

drewmoseley:
All green except for test_ssl_only as expected.  Merging

PR #612, opened 2018-09-24
More fixes after colibri imx7 testing

kacf:
I think this should be the last fixes. After this, I only have device specific settings left in my `local.conf`:

* `MENDER_MTDIDS = "nand0=gpmi-nand"`
* `MENDER_MTDPARTS = "gpmi-nand:512k(mx7-bcb),1536k(u-boot1)ro,1536k(u-boot2)ro,512k(u-boot-env),-(ubi)"`
* `MENDER_STORAGE_PEB_SIZE = "131072"`
* `MENDER_STORAGE_TOTAL_SIZE_MB = "512"`

I am not sure how strongly these are tied to the `colibri-imx7` machine type, especially the last one. We could potentially put these also into the `machine.conf`, but is it appropriate, or should the user set them?

No other changes are required to build for `colibri-imx7`.

### DIFF
--- a/meta-mender-core/classes/mender-setup-image.inc
+++ b/meta-mender-core/classes/mender-setup-image.inc
@@ -4,8 +4,10 @@ IMAGE_FSTYPES_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-image-ub
 IMAGE_FSTYPES_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-image-uefi', ' uefiimg uefiimg.bmap', '', d)}"
 IMAGE_FSTYPES_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-image-bios', ' biosimg biosimg.bmap', '', d)}"
 
-MACHINE_ESSENTIAL_EXTRA_RDEPENDS_append_x86 = " kernel-image"
-MACHINE_ESSENTIAL_EXTRA_RDEPENDS_append_x86-64 = " kernel-image"
+MACHINE_ESSENTIAL_EXTRA_RDEPENDS_append_mender-image_x86 =     " kernel-image"
+MACHINE_ESSENTIAL_EXTRA_RDEPENDS_append_mender-image_x86-64 =  " kernel-image"
+MACHINE_ESSENTIAL_EXTRA_RDEPENDS_append_mender-image_arm =     " kernel-image kernel-devicetree"
+MACHINE_ESSENTIAL_EXTRA_RDEPENDS_append_mender-image_aarch64 = " kernel-image kernel-devicetree"
 
 python() {
     features = d.getVar('DISTRO_FEATURES').split()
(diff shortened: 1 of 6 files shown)


meta-mender-core/classes/mender-setup-image.inc
```diff
@@ -4,8 +4,9 @@ IMAGE_FSTYPES_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-image-ub
 IMAGE_FSTYPES_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-image-uefi', ' uefiimg uefiimg.bmap', '', d)}"
 IMAGE_FSTYPES_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-image-bios', ' biosimg biosimg.bmap', '', d)}"
 
-MACHINE_ESSENTIAL_EXTRA_RDEPENDS_append_x86 = " kernel-image"
-MACHINE_ESSENTIAL_EXTRA_RDEPENDS_append_x86-64 = " kernel-image"
+MACHINE_ESSENTIAL_EXTRA_RDEPENDS_append_mender-image_x86 =    " kernel-image"
+MACHINE_ESSENTIAL_EXTRA_RDEPENDS_append_mender-image_x86-64 = " kernel-image"
+MACHINE_ESSENTIAL_EXTRA_RDEPENDS_append_mender-image_arm =    " kernel-image kernel-devicetree"
```
mirzak: Probably need a `append_mender-image_aarch64` as well as that is used for 64-bit arm boards.

meta-mender-toradex-nxp/recipes-bsp/u-boot/patches/0005-fw_env_main.c-Fix-incorrect-size-for-malloc-ed-strin.patch
```diff
@@ -0,0 +1,28 @@
+From 17236152bf80436567bf3f1f27af3915364ec0b6 Mon Sep 17 00:00:00 2001
+From: Kristian Amlie <kristian.amlie@northern.tech>
+Date: Wed, 4 Apr 2018 10:01:04 +0200
+Subject: [PATCH 5/5] fw_env_main.c: Fix incorrect size for malloc'ed string.
+
+Using sizeof gives the size of the pointer only, not the string.
+
+Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>
```
mirzak: I know we have not really enforced this, but what do you think about adding `Upstream-status:` tags on these patches.

Based on https://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines, Section "Patch Header Recommendations: Upstream-Status".

meta-mender-core/classes/mender-setup-image.inc
kacf: Good point. I'm always thinking that `arm` covers it, but I believe you are right.

meta-mender-toradex-nxp/recipes-bsp/u-boot/patches/0005-fw_env_main.c-Fix-incorrect-size-for-malloc-ed-strin.patch
kacf: Sure, I can add it for convenience. Our patch flow is too low that I would bother about enforcing this now though.

kacf:
> We are in the process of removing the `colibri-mx7-mender` target right?

Yes, and I didn't use that target for my build, I only used `colibri-imx7`, and master branches from all sources. However, because of the EasyInstaller, the work of removing `colibri-imx7-mender` is happening on rocko, which doesn't have the auto-patcher for Flash, so I'm unsure how much overlap there is. I can ask Alf when I see him.

kacf:
Updated!

mirzak:
LGTM

PR #616, opened 2018-09-25
Acceptance test files, i.e. py.test configuration and fixture modifie…

apodogrocki:
…d to work with non-Yocto build images

conftest.py/fixtures.py/common.py modified to work
with non-Yocto build images. Tests stored in the
mender-image-tests repository can be used to validate
whether images built with mender-convert tool are
correct.

Changelog: None

Issues: MEN-2098

Signed-off-by: Adam Podogrocki <a.podogrocki@gmail.com>

### DIFF
--- a/tests/acceptance/common.py
+++ b/tests/acceptance/common.py
@@ -325,32 +325,45 @@ def common_boot_from_internal():
 
 
 def latest_build_artifact(builddir, extension):
-    output = subprocess.check_output(["sh", "-c", "ls -t %s/tmp*/deploy/images/*/*%s | grep -v data*%s| head -n 1" % (builddir, extension, extension)])
+    if pytest.config.getoption('--test-conversion'):
+        sdimg_location = pytest.config.getoption('--sdimg-location')
+        output = subprocess.check_output(["sh", "-c", "ls -t %s/%s/*%s | grep -v data*%s| head -n 1" % (builddir, sdimg_location, extension, extension)])
+    else:
+        output = subprocess.check_output(["sh", "-c", "ls -t %s/tmp*/deploy/images/*/*%s | grep -v data*%s| head -n 1" % (builddir, extension, extension)])
     output = output.rstrip('\r\n')
     print("Found latest image of type '%s' to be: %s" % (extension, output))
     return output
 
-def get_bitbake_variables(target, env_setup="true", export_only=False):
+def get_bitbake_variables(target, env_setup="true", export_only=False, test_conversion=False):
     current_dir = os.open(".", os.O_RDONLY)
     os.chdir(os.environ['BUILDDIR'])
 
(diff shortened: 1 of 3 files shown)


drewmoseley:
Looks reasonable to me but I would definitely wait for @kacf to review as I don't know the test infrastructure very well.

tests/acceptance/common.py
```diff
@@ -325,32 +325,43 @@ def common_boot_from_internal():
...
-    for line in output.stdout:
+    for line in output:
         line = line.strip()
         match = matcher.match(line)
         if match is not None:
             ret[match.group(1)] = match.group(2)
 
-    output.wait()
+    if not test_conversion:
+        output.wait()
```
kacf: `output.wait()` is no longer valid after it was switched to `stdout`. You need to keep the original process handle, let's say `ps`, and do `output = ps.stdout` and then `ps.wait()`.

This is likely the reason for the massive test failures, since almost all the tests use this. :-)

tests/acceptance/common.py
```diff
@@ -325,32 +325,43 @@ def common_boot_from_internal():
...
+def get_bitbake_variables(target, env_setup="true", export_only=False, test_conversion=False):
     current_dir = os.open(".", os.O_RDONLY)
     os.chdir(os.environ['BUILDDIR'])
 
-    output = subprocess.Popen("%s && bitbake -e %s" % (env_setup, target),
-                              stdout=subprocess.PIPE,
-                              shell=True,
-                              executable="/bin/bash")
+    if test_conversion == True:
+        config_file_path = os.path.join(os.getcwd(), pytest.config.getoption('--test-variables'))
```
kacf: Suggest using this instead:

```
config_file_path = os.path.abspath(pytest.config.getoption('--test-variables'))
```

which will work even if `--test-variables` is already an absolute path.

tests/acceptance/common.py
```diff
@@ -325,32 +325,43 @@ def common_boot_from_internal():
...
-def get_bitbake_variables(target, env_setup="true", export_only=False):
+def get_bitbake_variables(target, env_setup="true", export_only=False, test_conversion=False):
     current_dir = os.open(".", os.O_RDONLY)
     os.chdir(os.environ['BUILDDIR'])
 
-    output = subprocess.Popen("%s && bitbake -e %s" % (env_setup, target),
-                              stdout=subprocess.PIPE,
-                              shell=True,
-                              executable="/bin/bash")
+    if test_conversion == True:
```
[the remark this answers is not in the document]
apodogrocki: OK

tests/acceptance/common.py
apodogrocki: OK

tests/acceptance/common.py
apodogrocki: OK

PR #618, opened 2018-09-26
Basehash mismatch and textrel error message fixes

ghost:
```
commit 06f0f6c273e7e67e34fbeefe86b7b2c392d71fc0
Date:   Wed Sep 26 15:20:02 2018

    Fix basehash mismatch error message on recent master.
    
    Switch to using IMAGE_VERSION_SUFFIX instead of DATETIME, which is
    what poky has also done in a65c1216d7c4c. As part of that they also
    removed the `IMAGE_NAME[vardepsexclude] += "DATETIME"` that existed
    before, which is why the error message started appearing. It was
    reintroduced for the IMAGE_VERSION_SUFFIX variable, so that one is
    safe to use.
    
    Changelog: None
    
commit c2d848630d6be198c15e48674f41604abf01fd33
Date:   Wed Sep 26 14:57:54 2018

    Skip Yocto QA textrel check for Mender.
    
    This gets rid of this message:
    
      QA Issue: ELF binary '<path>/mender.test' has relocations in .text
    
    This has already been submitted upstream for other Go components, so
    it appears to be a common Go issue.
    
    Changelog: None
    
```

### DIFF
--- a/meta-mender-qemu/recipes-connectivity/openssh/openssh_7.8p1+git.bbappend
+++ b/meta-mender-qemu/recipes-connectivity/openssh/openssh_7.8p1+git.bbappend
@@ -1,0 +1,2 @@
+DEPENDS_remove_vexpress-qemu-flash = "openssl"
+DEPENDS_append_vexpress-qemu-flash = " openssl10"


drewmoseley:
@kacf yes this fixes it.  Thanks.

PR #619, opened 2018-09-28
grub: Remove restriction that the core.img be at sector 1

mdavis777:


mdavis777:
I wish I could find a better definition of the location of core.img in diskboot.img  However they just

PR #620, opened 2018-10-01
grub: Remove restriction that the core.img be at sector 1

mdavis777:


mirzak:
Thanks, I started the test runners. Lets see what they think :)

kacf:
Thanks @mdavis777!

PR #622, opened 2018-10-03
tests: Add mender-image-tests as a submodule.

kacf:
This is attempting to deal with the fact that image tests now live in
a separate repository. The existing method where we copy over the
tests from that repository is very fragile, and doesn't allow
specifying alternate branches, which will be important as newer
versions of mender-image-tests drift out of sync with older branches
of meta-mender. We will then need to branch mender-image-tests and
switch the submodule over to that.

One slight complication with our use of this submodule is that we want
to track branches, not explicit commits. We solve this by having an
automatic pull before testing starts, which can be disabled if you're
debugging something locally, or in the Jenkins case where we will
check out the exact commit that was specified in the build.

The "repo" tool was considered as an alternative to submodules, but I
didn't see enough benefit, given that repo is very workflow oriented,
and this submodule is not really meant to be used by users very much
(hence the "ignore" attribute in the .gitmodules file). This also has
the benefit of not requiring another tool.

Changelog: None

Goes together with https://github.com/mendersoftware/mender-qa/pull/217.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,5 @@
+[submodule "tests/acceptance/image-tests"]
+	path = tests/acceptance/image-tests
+	url = https://github.com/mendersoftware/mender-image-tests
+	branch = master
+	ignore = all
--- a/tests/acceptance/conftest.py
+++ b/tests/acceptance/conftest.py
@@ -15,6 +15,7 @@
 
 import os
 import os.path
+import subprocess
 
 from fabric.api import *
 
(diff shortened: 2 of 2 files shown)


kacf:
Feel free to comment, but I'm still testing this.

kacf:
Alright, looks promising. I still need the final results, but the tricky part of the cloning is over.

kacf:
@drewmoseley: This is related to #611.

.gitmodules
```diff
@@ -0,0 +1,5 @@
+[submodule "tests/acceptance/image-tests"]
+	path = tests/acceptance/image-tests
```
apodogrocki: Setting such path implies changes in _mender-qa/scripts/jenkins-yoctobuild-build.sh_. Cannot we use simply _tests/acceptance_ path here? Having tests in two paths would cause the py.test must be run twice, wouldn't it?

apodogrocki: Ok, I saw [mender-qa/pull/217](https://github.com/mendersoftware/mender-qa/pull/217) So only question is whether py.test searches tests recursively, as mutual tests for Yocto and non-Yocto builds are stored now in _meta-mender/tests/acceptance/image-tests/tests_?

kacf: Yes, it searches recursively, I checked the log to be sure, and it runs all tests including those. The directory nesting is a bit cumbersome, but a submodule must be in a subdirectory, so we have no choice.

PR #623, opened 2018-10-04
QEMU: Always try to run with KVM enabled, only fall back to emulation.

kacf:
Changelog: Title


drewmoseley:
This looks reasonable to me.

PR #625, opened 2018-10-08
meta-mender-toradex-nxp: fix fw_env.config override

mirzak:
This is only valid for rocko as auto-patching works on master and sumo and no custom fw_env.config file is provided.


drewmoseley:
In principle I have no problem with this but do we have any idea why this was needed now?  @alfrunes ping.

drewmoseley:
@kacf can you take a look at the CI failure here? It looks related to the recent repository changes.

mirzak:
@drewmoseley , you refresh patches are included in this PR and required as "rocko" and "rocko-next" are now identical in upstream Toradex repos.

mirzak:
I started the jobs manually and I probably failed at that

mirzak:
"vexpress-qemu_master_rocko_acceptance_tests" seems to fail with the same error it initially did when I opened the PR.

I guess we can not get rid of the "master_master" fails that I triggered which I should not have :)

mirzak:
"vexpress-qemu_master_rocko_acceptance_tests" still failing. We can ignore the "maste_master" builds/test as that was unintentionally triggered by me.

drewmoseley:
@mirzak I submitted https://github.com/mendersoftware/meta-mender/pull/631 which is identical to this but with a different commit hash.  The tests there succeeded so I think we can call this done.  The failures here are apparently CI infra related and not due to these changes.

mirzak:
Thanks @drewmoseley

PR #626, opened 2018-10-08
Added test cases for mender-artifact-V3 

oleorhagen:
Changelog: None


oleorhagen:
In short:
1. This mirrors the tests for version-2 in test_update.
2.  Modifies slightly the testing of the artifact-structure in `test_mender-artifact.py`. Here version 2 is now no longer treated as the default case. Also I did not add version 3 as the new default, as I did not think this necessarily will be true for a little while.

oleorhagen:
Also, I have not tested the `test-mender-artifact.py` locally yet, but `test-update` passes.

tests/acceptance/test_update.py
```diff
@@ -397,25 +397,145 @@ def test_network_based_image_update(self, successful_image_update_mender, bitbak
...
+                                            success=False),
+                              SignatureCase(label="EC, Correctly signed, but header does not match checksum, key present, version 3",
+                                            signature=True,
+                                            signature_ok=True,
+                                            key=True,
+                                            key_type="EC",
+                                            checksum_ok=True,
+                                            header_checksum_ok=False,
+                                            update_written=False,
+                                            artifact_version=3,
```
kacf: I don't see you using `artifact_version` anywhere. Is that coming?

oleorhagen: https://github.com/mendersoftware/meta-mender/pull/626/files/47db16c6eceecf22bd305b3a811bc0cf035da76d#diff-379340f30fc917e8f0c80a32964a52b9R575

This will generate an artifact-version 3, if I'm not mistaken?

oleorhagen:
```
       # Generate artifact with specific version. None means default.
         if sig_case.artifact_version is not None:
             artifact_args += " -v %d" % sig_case.artifact_version
```

PR #630, opened 2018-10-10
MEN-1643 Remove mender-specific device for Toradex layer

alfrunes:
Also refactored teziimg class to the format used in the Toradex layers, and included the correct "slides" in the metadata files for easyinstaller.


kacf:
Decided not to bring this into master. The whole patching step is not needed anymore due to auto-patching, and the `tezimg` stuff doesn't really belong in meta-mender. We can fetch it from rocko if we ever need it.

drewmoseley:
@kacf I'm confused; we already have the teziimg stuff and Mender-specific machine in master. Does this need to be replaced with something that simply removes that? Will we have teziimg support in future branches? That's pretty important as the Toradex Easy Installer is part of our approach to supporting them.

drewmoseley:
OK. I see #643 lower in my inbox. Taking the discussion over there. ;-)

PR #632, opened 2018-10-11
[rocko] QEMU: Always try to run with KVM enabled, only fall back to emulation.

kacf:
Changelog: Title

(cherry picked from commit 8bb9c4bb1ddde7e2b00216b490854ade23393a42)


kacf:
I believe this is necessary after https://github.com/mendersoftware/integration/commit/462552e27f6c577caea7dc492e138c0283094a81, because otherwise it will try to use `-enable-kvm` on vexpress-qemu, which is ARM, and will therefore fail.

kacf:
Passed everything with just a couple of spurious errors.

PR #635, opened 2018-10-15
Add PARTUUID generation and integration

mdavis777:


mdavis777:
This isn't 100% yet.  At the very least it needs a Changelog and a lot of error checking built in.  Just thought I would get the review of it started so if there are any major changes needed.

mdavis777:
Added some error checking.  Might need to move to a different place, or be split up but it seems to work.
Also fixed some incorrect comments and cleaned up some debug messages.

meta-mender-core/classes/mender-helpers.bbclass
```diff
@@ -82,6 +82,30 @@ get_part_number_hex_from_device() {
...
+
+def test_disk_identifier(d, varName):
+    import re
+    pattern = re.compile("^[0-9a-f]{8}$")
+    uuid = d.getVar(varName)
+    if len(uuid) == 0:
+       bb.fatal("%s Must be defined" % varName )
+    elif not pattern.match(uuid):
+       bb.fatal("%s Is not a valid Disk Identifier.  It most follow the format ########  Allow lower case hex values" % varName )
+    return
```
kacf: Please prefix all three of these functions with `mender_`, since this is in a global class.

meta-mender-core/recipes-bsp/grub-mender-grubenv/grub-mender-grubenv_1.1.0.bb
```diff
@@ -41,6 +41,25 @@ python() {
...
 }
 
+python do_partuuid_error_check() {
+        if bb.utils.contains('DISTRO_FEATURES', 'mender-image-gpt', True, False, d) or bb.utils.contains('DISTRO_FEATURES', 'mender-image-uefi', True, False, d):
+            test_part_uuid(d, 'MENDER_ROOTFS_PART_A_UUID')
+            test_part_uuid(d, 'MENDER_ROOTFS_PART_B_UUID')
+            test_part_uuid(d, 'MENDER_DATA_PART_UUID')
+            test_part_uuid(d, 'MENDER_BOOT_PART_UUID')
+        elif bb.utils.contains('DISTRO_FEATURES', 'mender-image-sd', True, False, d) or bb.utils.contains('DISTRO_FEATURES', 'mender-image-bios', True, False, d):
+            test_disk_identifier(d, 'MENDER_DISK_IDENTIFIER')
```
kacf: * I would make it a new `if` instead of `elif`. Although we are not testing it, theoretically `DISTRO_FEATURES` could contain both cases. Might as well catch both errors then.
* Indentation should be four spaces.

meta-mender-core/recipes-bsp/grub-mender-grubenv/grub-mender-grubenv_1.1.0.bb
```diff
@@ -60,13 +79,37 @@ mender_grub_storage_device=$mender_grub_storage_device
...
 }
+do_configure[vardeps] = " \
+    MENDER_ROOTFS_PART_A \
+    MENDER_ROOTFS_PART_B \
+    MENDER_STORAGE_DEVICE_BASE \
+    MENDER_ROOTFS_PART_A_UUID \
+    MENDER_ROOTFS_PART_B_UUID \
+    MENDER_DISK_IDENTIFIER \
+    KERNEL_DEVICETREE \
+"
```
kacf: This should not be needed when you are making direct references to variables using the `${VAR}` notation. It will discover the references automatically.

meta-mender-core/classes/mender-helpers.bbclass
```diff
@@ -82,6 +82,30 @@ get_part_number_hex_from_device() {
     fi
 }
 
+def get_part_nummber_padded_from_device(device):
```
kacf: `s/nummber/number/`

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -23,16 +23,32 @@ def mender_linux_partition_base(dev):
...
 MENDER_ROOTFS_PART_A_DEFAULT = "${MENDER_STORAGE_DEVICE_BASE}${@bb.utils.contains('MENDER_BOOT_PART_SIZE_MB', '0', '1', '2', d)}"
 MENDER_ROOTFS_PART_B ??= "${MENDER_ROOTFS_PART_B_DEFAULT}"
 MENDER_ROOTFS_PART_B_DEFAULT = "${MENDER_STORAGE_DEVICE_BASE}${@bb.utils.contains('MENDER_BOOT_PART_SIZE_MB', '0', '2', '3', d)}"
 
+# The partition uuid of the two rootfs partitions in the A/B partition layout.
+# Follows the Format of XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXX
+MENDER_ROOTFS_PART_A_UUID ??= "${MENDER_ROOTFS_PART_A_UUID_DEFAULT}"
+MENDER_ROOTFS_PART_A_UUID_DEFAULT = ""
+MENDER_ROOTFS_PART_B_UUID ??= "${MENDER_ROOTFS_PART_B_UUID_DEFAULT}"
+MENDER_ROOTFS_PART_B_UUID_DEFAULT = ""
```
kacf: I think that these should be reflected in `MENDER_ROOTFS_PART_A` and `MENDER_ROOTFS_PART_B`. Shouldn't those be set to values like `/dev/disk/by-uuid/12345678-1234-5678-9abc-56789abcdef`? I think this is better than relying on the existing indexed numbers.

meta-mender-core/recipes-bsp/grub-mender-grubenv/grub-mender-grubenv_1.1.0.bb
```diff
@@ -60,13 +79,37 @@ mender_grub_storage_device=$mender_grub_storage_device
...
+            mender_rootfsb_uuid=${MENDER_ROOTFS_PART_B_UUID}
+        else
+            mender_rootfsa_uuid="${MENDER_DISK_IDENTIFIER}-${@get_part_nummber_padded_from_device(MENDER_ROOTFS_PART_A)}"
+            mender_rootfsb_uuid="${MENDER_DISK_IDENTIFIER}-${@get_part_nummber_padded_from_device(MENDER_ROOTFS_PART_B)}"
+        fi
+
+        cat >> ${B}/mender_grubenv_defines <<EOF
+mender_rootfsa_uuid=$mender_rootfsa_uuid
+mender_rootfsb_uuid=$mender_rootfsb_uuid
+EOF
```
kacf: Following my earlier comment about setting

```
MENDER_ROOTFS_PART_A="/dev/disk/by-uuid/12345678-1234-5678-9abc-56789abcdef"
```

I believe the need for `mender_rootfsa_uuid` and `mender_rootfsb_uuid` falls away, and we can use

```
mender_rootfsa_part="PARTUUID=12345678-1234-5678-9abc-56789abcdef"
```

directly instead.

meta-mender-core/recipes-mender/mender/mender.inc
```diff
@@ -113,8 +113,21 @@ python do_prepare_mender_conf() {
...
+            tmpRootAUUID=d.getVar('MENDER_ROOTFS_PART_A_UUID')
+            tmpRootBUUID=d.getVar('MENDER_ROOTFS_PART_B_UUID')
+        else:
+            tmpRootAUUID=d.getVar('MENDER_DISK_IDENTIFIER') + "-" + get_part_nummber_padded_from_device(d.getVar('MENDER_ROOTFS_PART_A'))
+            tmpRootBUUID=d.getVar('MENDER_DISK_IDENTIFIER') + "-" + get_part_nummber_padded_from_device(d.getVar('MENDER_ROOTFS_PART_B'))
+        conf_maybe_add("RootfsPartA", "/dev/disk/by-partuuid/" + tmpRootAUUID, getvar=False, integer=False)
+        conf_maybe_add("RootfsPartB", "/dev/disk/by-partuuid/" + tmpRootBUUID, getvar=False, integer=False)
+    else:
+        conf_maybe_add("RootfsPartA", "MENDER_ROOTFS_PART_A", getvar=True, integer=False)
+        conf_maybe_add("RootfsPartB", "MENDER_ROOTFS_PART_B", getvar=True, integer=False)
```
kacf: This if block might also be unneeded, ref my earlier comment.

meta-mender-core/classes/mender-setup.bbclass
mdavis777: We would still need the number of the partition for order and if someone wants to have custom partitions. I was going for least change possible.

mdavis777:
I am out for a couple days. I will get started on the changes once I am back.

meta-mender-core/recipes-bsp/grub-mender-grubenv/grub-mender-grubenv_1.1.0.bb
kacf: TBH, I'm not sure either. I'm just thinking there is no reason why we couldn't have both, just in case.

mdavis777:
Ok thinking about it I can see 2 real possible solutions.

1.  Put the path (/dev/disk/by-partuuid/X) in MENDER_ROOTFS_PART_X

For msdos we can parse the partition numbers so that isn't a big deal easy enough to fix.
For gpt we would still need a part number in a separate variable
This would of course be defaulted with the normal structure

2.  Separate the partition number to a new variable for everything
Generate the MENDER_ROOTFS_PART_X variables for everything
This would use MENDER_STORAGE_DEVICE  + PARTITION NUMBER 
For example (/dev/sda+3) = MENDER_ROOTFS_PART_A
This would get rid of the need for the get_part_number_from_device type functions

I don't know that either path is really better or worse so I will leave it up to you to choose.

mdavis777:
Contrary to what I thought originally going down path 2 actually results in much cleaner code.  Obviously it is a very fundamental low level change, so I would understand if that isn't the path you decide to go with.

kacf:
> Put the path (/dev/disk/by-partuuid/X) in MENDER_ROOTFS_PART_X

Yes, this what I was aiming at.

> Separate the partition number to a new variable for everything
> This would get rid of the need for the get_part_number_from_device type functions

I'm not sure I follow. The idea was to try to avoid partition numbers altogether when using the `mender-partuuid` feature. Something like this:

```diff
 MENDER_ROOTFS_PART_A ??= "${MENDER_ROOTFS_PART_A_DEFAULT}"
 MENDER_ROOTFS_PART_A_DEFAULT = "${MENDER_STORAGE_DEVICE_BASE}${@bb.utils.contains('MENDER_BOOT_PART_SIZE_MB', '0', '1', '2', d)}"
+MENDER_ROOTFS_PART_A_DEFAULT_mender-partuuid = "/dev/disk/by-partuuid/${MENDER_ROOTFS_PART_A_UUID}"
 MENDER_ROOTFS_PART_B ??= "${MENDER_ROOTFS_PART_B_DEFAULT}"
```

Unfortunately, the `/dev/disk/by-partuuid/<ID>` notation doesn't work as the kernel argument, so some tweaking is necessary to get GRUB to pass this correctly. I assume this would something like `mender_get_kernel_root_from_dev_reference()` where you replace `/dev/disk/by-partuuid/` with `PARTUUID=`.

mdavis777:
> I'm not sure I follow. The idea was to try to avoid partition numbers altogether when using the `mender-partuuid` feature. Something like this:

It really isn't so much about getting rid of partition numbers are fixing boots on devices that don't have a fixed drive order.  I don't think there is a way to get rid of partitions all together with PARTUUID.  They are still needed for. 
> ${mender_grub_storage_device},${ptable_type}${mender_boot_part})/boot/${kernel_imagetype}
and device tree

> devicetree (${mender_grub_storage_device},${ptable_type}${mender_boot_part})/boot/${kernel_devicetree}

mdavis777:
We could require that the first / last digit of the gpt partuuid is the partition number.  It is a hack, but it would reduce it down to 1 variable.

mdavis777:
I have updated with a full working PATH 2 implementation.
MENDER_ROOTFS_PART_A = "/dev/disk/by-partuuid/12345678-1234-1234-1234-123456789abc"
is fully supported.
This is the more break change as users would need to now set MENDER_DEVICE_BASE and/or MENDER_ROOTFS_PART_A_NUMBER(and the others) if they want to move from the defaults.

I now I will work on a path 1 version and make a separate pull request so that they can be compared.

mdavis777:
I created pull request #636 as an alternative to this one.  It does not break existing non-default setups like this pull request would.  Let me know which you prefer.  This one would still need some error checking and maybe expanded defaults, but the general principle is there.

meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
```diff
@@ -54,14 +54,11 @@ do_provide_mender_defines() {
...
-        MENDER_BOOT_PART_NUMBER_HEX=`get_part_number_hex_from_device ${MENDER_BOOT_PART}`
+        MENDER_BOOT_PART_NUMBER_HEX=`number_to_hex ${MENDER_BOOT_PART_NUMBER}`
     fi
 
-    MENDER_ROOTFS_PART_A_NUMBER=`get_part_number_from_device ${MENDER_ROOTFS_PART_A}`
-    MENDER_ROOTFS_PART_A_NUMBER_HEX=`get_part_number_hex_from_device ${MENDER_ROOTFS_PART_A}`
-    MENDER_ROOTFS_PART_B_NUMBER=`get_part_number_from_device ${MENDER_ROOTFS_PART_B}`
-    MENDER_ROOTFS_PART_B_NUMBER_HEX=`get_part_number_hex_from_device ${MENDER_ROOTFS_PART_B}`
+    MENDER_ROOTFS_PART_A_NUMBER_HEX=`number_to_hex ${MENDER_ROOTFS_PART_A_NUMBER}`
+    MENDER_ROOTFS_PART_B_NUMBER_HEX=`number_to_hex ${MENDER_ROOTFS_PART_B_NUMBER}`
```
kacf: `MENDER_ROOTFS_PART_A_NUMBER` is referred to below, so we need to keep it around. I suggest logic similar to this:

```
if ${@bb.utils.contains('DISTRO_FEATURES', 'mender-partuuid', 'true', 'false', d)}; then
    MENDER_ROOTFS_PART_A_NUMBER_HEX=`number_to_hex ${MENDER_ROOTFS_PART_A_NUMBER}`
    MENDER_ROOTFS_PART_B_NUMBER_HEX=`number_to_hex ${MENDER_ROOTFS_PART_B_NUMBER}`
else
    MENDER_ROOTFS_PART_A_NUMBER=`get_part_number_from_device ${MENDER_ROOTFS_PART_A}`
    MENDER_ROOTFS_PART_A_NUMBER_HEX=`get_part_number_hex_from_device ${MENDER_ROOTFS_PART_A}`
    MENDER_ROOTFS_PART_B_NUMBER=`get_part_number_from_device ${MENDER_ROOTFS_PART_B}`
    MENDER_ROOTFS_PART_B_NUMBER_HEX=`get_part_number_hex_from_device ${MENDER_ROOTFS_PART_B}`
fi
```

And then `get_part_number_from_device` can contain some detection code for UUIDs, and suggest enabling the `mender-partuuid` and setting the `MENDER_ROOTFS_PART_A_NUMBER` variables, in the error message.

meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
```diff
@@ -54,14 +54,11 @@ do_provide_mender_defines() {
     fi
 
     if [ -n "${MENDER_BOOT_PART}" ]; then
-        MENDER_BOOT_PART_NUMBER=`get_part_number_from_device ${MENDER_BOOT_PART}`
-        MENDER_BOOT_PART_NUMBER_HEX=`get_part_number_hex_from_device ${MENDER_BOOT_PART}`
+        MENDER_BOOT_PART_NUMBER_HEX=`number_to_hex ${MENDER_BOOT_PART_NUMBER}`
```
kacf: Same here as below.

meta-mender-core/recipes-bsp/grub-mender-grubenv/grub-mender-grubenv_1.1.0.bb
```diff
@@ -44,8 +44,8 @@ python() {
 do_configure() {
     set -x
 
-    mender_rootfsa_part=$(get_part_number_from_device ${MENDER_ROOTFS_PART_A})
-    mender_rootfsb_part=$(get_part_number_from_device ${MENDER_ROOTFS_PART_B})
+    mender_rootfsa_part=${MENDER_ROOTFS_PART_A_NUMBER}
+    mender_rootfsb_part=${MENDER_ROOTFS_PART_B_NUMBER}
```
kacf: Should use detection for `mender-partuuid` here as well, like my other comment.

mdavis777:
Ya I thought Grub supported PARTUUID to, but sadly it appears to be a battle between grub and the kernel.  The kernel only supports PARTUUID and Grub refuses to add it saying the kernel should support labels and fs-uuid.  
I think a good next step would be to support labels in grub and use partuuid for the kernel.  That would get rid of the need for partition numbers, strings like /dev/sdX, and grub device (hd0) stuff all together.

I think the main hurdle is to have the client reinstate the label after update.

Anyways I will start the work on getting this pull request through.

mdavis777:
New request pushed.

meta-mender-core/recipes-bsp/grub-mender-grubenv/grub-mender-grubenv_1.1.0.bb
```diff
@@ -44,10 +44,18 @@ python() {
...
+        mender_rootfsb_part=${MENDER_ROOTFS_PART_B_NUMBER}
+    else
+        mender_rootfsa_part=$(get_part_number_from_device ${MENDER_ROOTFS_PART_A})
+        mender_rootfsb_part=$(get_part_number_from_device ${MENDER_ROOTFS_PART_B})
+    fi
+
     if [ -n "${MENDER_GRUB_STORAGE_DEVICE}" ]; then
         mender_grub_storage_device=${MENDER_GRUB_STORAGE_DEVICE}
+    elif ${@bb.utils.contains('DISTRO_FEATURES', 'mender-partuuid', 'true', 'false', d)}; then
+        bberror "MENDER_GRUB_STORAGE_DEVICE is required for meta-partuuid"
```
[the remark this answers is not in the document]
mdavis777: I don't know why I can't type mender.  Always ends up meta.  Too much messing around with meta-layers I guess.

kacf:
Excellent, thanks!

And I have one more request: Add a `Signed-off-by:` line with the same content as the author. Sorry, I didn't notice it was missing before, and our test farm *will* complain about it! :-)

I think we can do a test run of this soon, when both of the other two PRs are ready.

mdavis777:
It used to be there.  Most of lost it in the rework somewhere.  It is back now.

meta-mender-core/classes/mender-helpers.bbclass
```diff
@@ -82,6 +85,24 @@ get_part_number_hex_from_device() {
...
+    if deviceName:
+        uuid=deviceName.replace('/dev/disk/by-partuuid/', '')
+        gptMatch = re.search('^[0-9a-f]{8}\-[0-9a-f]{4}\-[0-9a-f]{4}\-[0-9a-f]{4}\-[0-9a-f]{12}$', uuid)
+        if gptMatch:
+            return gptMatch.group(0)
+
+        msdosMatch = re.search("^[0-9a-f]{8}\-[0-9]{2}$", uuid)
+        if msdosMatch:
+            return msdosMatch.group(0)
+        bb.fatal("%s Does not contain a valid PARTUUID path" % deviceName )
```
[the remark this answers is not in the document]
mdavis777: Sorry I was out most of the week I will look at this and put it through another round of testing with partuuid off to make sure I didn't miss any others.

mdavis777:
Ok I fixed that issue and ran it through another round of testing with and without uuid and cleaned up a couple other small issues.

kacf:
Thanks, retesting.

kacf:
Merged in #641.

kacf:
Thanks for the hard work, @mdavis777, everything is in now!

I'll make sure the mender changes will be backported to older release branches, particularly 1.7. The meta-mender changes will be available in the upcoming thud.

PR #636, opened 2018-10-20
Add PARTUUID generation and integration

mdavis777:
Changelog: Title


mdavis777:
This is an alternate to pull request #635.  It is a little messier, but keeps compatibility with old setups

mdavis777:
Sounds like #635 is the chosen solution

PR #643, opened 2018-11-09
Remove colibri-imx7-mender MACHINE type, replaced by colibri-imx7.

kacf:
Changelog: Title


kacf:
@mirzak, @drewmoseley: This is an alternative to #605, for master. I say we merge this one, for the future, and then simply close the rocko one without any further work. Agree?

mirzak:
Thanks for picking this up. This is good to go in my opinion.

Just an FYI.

There is a plan to remove all community layers from meta-mender, [MEN-2078](https://tracker.mender.io/browse/MEN-2078). My intention was to do this before we create a Thud branch, otherwise we will be stuck with it for another 6 months.

kacf:
Right, also a good initiative. I plan to create thud for December, probably branching in the last week of November some time. So anytime before then. :-)

drewmoseley:
What about the easy installer support? Do we still have that without any of the contents of #605? That's pretty important and we don't want to have to redo that work the next time we decide to have a trade show demo or a webinar.

PR #653, opened 2018-11-19
Add raspberrypi-demo layer to build.

kacf:
Changelog: Fix missing wpa_supplicant in Raspberry Pi demo images.


estenberg:
:+1: 

Only comment is that we should have this in master (and future thud) as well, so we get it in 1.7 final.

Thanks for testing & fixing, @kacf !

kacf:
Indeed, this was mostly for testing purposes!

PR #656, opened 2018-11-22
u-boot: Remove patch which enables kernel load from `loadaddr` variable.

kacf:
This requirement has been discussed with U-Boot developers, and has
already been removed from the auto patching tool in df968dc6ac3, in
favor of the `kernel_addr_r` variable. We don't expect many problems
with boards, since most already support the `kernel_addr_r`, and in
general we want to reduce the amount of patching of U-Boot.

Changelog: Remove support for `loadaddr` variable and
`CONFIG_LOADADDR` config setting in U-Boot. This only affects boards
that use U-Boot as a first stage boot loader in order to use UEFI to
load GRUB as the second stage boot loader. For most boards it should
not be problematic, since most support `kernel_addr_r`. If there is
a problem however, it might be necessary to forward port [this
patch](https://github.com/mendersoftware/meta-mender/blob/b39aa8aeecdf2b8cce3dbcce25ec044073568348/meta-mender-core/recipes-bsp/u-boot/patches/0007-distro_bootcmd-Switch-bootefi-to-use-loadaddr-by-def.patch)
to the U-Boot version in question.


kacf:
Thanks!

PR #658, opened 2018-11-26
MEN-2145: Split mender.conf configuration file into /etc and /data.

oleorhagen:
Changelog: Title


meta-mender-core/recipes-mender/mender/mender.inc
```diff
@@ -171,7 +182,8 @@ do_install() {
 
     #install configuration
     install -d ${D}/${sysconfdir}/mender
-    install -m 0644 ${B}/mender.conf ${D}/${sysconfdir}/mender
+    install -m 0644 ${B}/transient_mender.conf ${D}/${sysconfdir}/mender/mender.conf
+    install -m 0600 ${B}/data/persistent_mender.conf ${D}/${datadir}/mender/mender.conf
```
kacf: The `/data` directory is our invention, and the `${datadir}` variable from Yocto is something completely different (`/usr/share` I believe). I don't think we have a dedicated variable for it, so just use `${D}/data` directly.

Also, you almost certainly will have to create the directory first, aka the `install -d` a couple of lines up.

oleorhagen:
Slight rewrite, but the functionality is now tested, and works like expected. Have a look at the persistent_configs variable to see what you would like in your persistent config.

meta-mender-core/recipes-mender/mender/mender.inc
```diff
@@ -122,9 +122,28 @@ python do_prepare_mender_conf() {
     # Tenant-token is optional, but falls back to a default-value set in config.go
     conf_maybe_add("TenantToken", "MENDER_TENANT_TOKEN", getvar=True, integer=False)
 
-    dst_conf = os.path.join(d.getVar("B"), "mender.conf")
+    # Split the mender.conf file into a transient and a persistent configuration file.
+    persistent_configs = ["RootfsPartA", "RootfsPartB"]
```
oleorhagen: Here!

meta-mender-core/recipes-mender/mender/mender.inc
```diff
@@ -122,9 +122,28 @@ python do_prepare_mender_conf() {
...
+
+    persistent_conf = {}
+
+    # Extract the variables that are destined for the persistent mender-configuration.
+    for config_var in transient_conf:
+        if config_var in persistent_configs:
+          persistent_conf[config_var] = transient_conf[config_var]
```
kacf: It just occurred to me that this is not upgrade-safe. Imagine if you have an existing device that has everything in `/etc/mender/mender.conf`. Its new update will have stuff split into `/etc` and `/data`, but `/data` is not distributed in updates, only in initial deployments. Hence the device will be missing this information once it reboots into the new update. And there is no simple workaround with the current approach.

I think we can solve this by introducing a `PACKAGECONFIG` option, say `split-mender-conf`, which defaults to on, but which can be turned off for those who need it. Essentially it would be along these lines:

```bash
# Just needs to defined. It normally deals with dependencies,
# but we don't need any here.
PACKAGECONFIG[split-mender-conf] = ",,,"

PACKAGECONFIG_append = " split-mender-conf"

# And then in every place where you need to make a decision based on it.
# Python:
bb.utils.contains('PACKAGECONFIG', 'split-mender-conf', True, False, d)
# Shell
${bb.utils.contains('PACKAGECONFIG', 'split-mender-conf', 'true', 'false', d)}
```

So then you simply don't do any splitting if it's not set. What do you think?

kacf: This should also be mentioned in the changelog, what has happened, and how to turn it off for those who need it. And maybe even an entry in our troubleshooting guide.

kacf: Perhaps @mirzak or @drewmoseley have some input also. Basically, how do we ensure that upgrading works if we start splitting `mender.conf` across `/etc` and `/data`.

drewmoseley: Is it possible to create the /data/ contents with a state script of some kind?

kacf: That's also a possibility. One advantage of doing it that way is that we would properly migrate everyone, instead of keeping support for the old way indefinitely. We could make this as an optional script, and document the upgrade procedure.

oleorhagen: I feel like it should be the other way around. You have to enable this functionality, rather than disable it. Thus migrating will be an option, but strongly advised. However, I do not have any strong feelings regarding this. First taker gets the decision.

kacf: From a compatibility perspective, I see your point, but I think this is too detailed that most people will care enough to enable it, so I still think that our recommendation (which is to split) should be the default.

It will of course mean that we cannot backport this, but this is rather expected I would say.

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -160,6 +160,18 @@ MENDER_UBOOT_PRE_SETUP_COMMANDS_DEFAULT = ""
...
+# All the allowed mender configuration variables
+MENDER_CONFIGURATION_VARS ?= "\
+    RootfsPartA \
+    RootfsPartB \
+    InventoryPollIntervalSeconds \
+    RetryPollIntervalSeconds \
+    ArtifactVerifyKey \
+    ServerCertificate \
+    ServerURL \
+    UpdatePollIntervalSeconds"
```
oleorhagen: One unified place for storing this info. Don't know if this is your preference, @kacf.

oleorhagen:
@drewmoseley I have added a state-script for migrating the configuration as well, for those that want to move to the new configuration management (it's [also] what we do). It has an explicit dependency upon jq. Don't know if this is acceptable or not?

meta-mender-core/recipes-mender/mender-migrate-configuration/files/mender-migrate-configuration
```diff
@@ -0,0 +1,30 @@
...
+# file currently on the device - Fail.
+jq --exit-status '%jq-program-marker% | to_entries | .[] | .value == null | not' /etc/mender/mender.conf > /dev/null
+if [ "$?" -ne "0" ]; then
+    NON_EXISTENT_VARIABLE=$(jq '%jq-program-marker% | to_entries | .[] | select(.value == null) | .key' /etc/mender/mender.conf)
+    echo "Variable ${NON_EXISTENT_VARIABLE} was not found in /etc/mender/mender.conf"
+fi
+
+# Create the persistent configuration file from the variables set in
+# MENDER_PERSISTENT_CONFIGURATION_VARS.
+jq '%jq-program-marker%' /etc/mender/mender.conf > /data/mender/mender.conf
```
kacf: For resiliency against spontaneous reboots, it is usually better to do:

```
jq '%jq-program-marker%' /etc/mender/mender.conf > /data/mender/mender.conf.new
sync
mv /data/mender/mender.conf.new /data/mender/mender.conf
sync
```

meta-mender-core/recipes-mender/mender-migrate-configuration/files/mender-migrate-configuration
```diff
@@ -0,0 +1,30 @@
...
+
+# Remove the fields from the transient configuration file (/etc/mender/mender.conf),
+# that has now been moved into the persistent configuration (/data/mender/mender.conf).
+# No need to error-check here, as a missing field will simply behave like an identity transformation.
+jq 'del(%jq-delete-fields-marker%)' /etc/mender/mender.conf > mender-transient.conf
+mv mender-transient.conf /etc/mender/mender.conf
```
kacf: `sync` after each one would be good here as well.

meta-mender-core/recipes-mender/mender-migrate-configuration/mender-migrate-configuration.bb
```diff
@@ -0,0 +1,40 @@
...
+          file://LICENSE;subdir=${PN}-${PV} \
+          "
+
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=e3fc50a88d0a364313df4b21ef20c29e"
+
+inherit mender-state-scripts
+
+ALLOW_EMPTY_${PN} = "1"
+
```
kacf: Suggest adding:

```
RDEPENDS_${PN} += "jq"
```

oleorhagen:
Good eye as usual! @kacf. All your comments have now been implemented, and this should be ready to merge.

kacf:
I [added a test for it](https://github.com/mendersoftware/mender-image-tests/pull/5) as well, since I need to test the cooperation between meta-mender and mender-image-tests a bit anyway.

Tested [here](https://github.com/mendersoftware/meta-mender/pull/679) with the latest checksum changes as well.

PR #659, opened 2018-11-27
Rocko backports

drewmoseley:
This brings rocko inline with tag v2018.11.2 in sumo


drewmoseley:
@kacf anything obviously wrong here?

PR #662, opened 2018-11-28
Remove error message during GRUB booting.

DominikAdamski:
When GRUB is launched it tries to search configuration files in directory ((hd0,gpt1)/EFI/BOOT)/EFI/BOOT/grub.cfg . This directory does not exist. Search command was was setup  redundantly by yocto grub layer. Searching cfg file is not needed, because prefix path already points to the path where the proper grub.cfg exists. Replacing improper yocto cfg file by the correct one causes that there is no strange error message during booting.

Signed-off-by: Dominik Adamski <adamski.dominik@gmail.com>

### DIFF
--- a/meta-mender-core/recipes-bsp/grub/files/cfg
+++ b/meta-mender-core/recipes-bsp/grub/files/cfg
@@ -1,0 +1,1 @@
+set prefix=($root)/EFI/BOOT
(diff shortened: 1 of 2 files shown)


drewmoseley:
Looks ok to me but I'll defer to @kacf

kacf:
Expected failure (problem in master, nothing to do with this PR). Merging.

PR #663, opened 2018-11-29
Several fixes for thud and later

kacf:
```
commit c45085187ec79dc2a2dcbb2a2566fe509f0194f0
Date:   Fri Nov 30 10:20:18 2018

    Fix images not being modifiable by mender-artifact in thud and later.
    
    Changelog: Title
    
    The problem is that the "metadata_csum" feature is very recent, and in
    fact even the version shipped in poky (1.44.3) doesn't support it
    (both fsck.ext4 and tune2fs refuse to handle it). Why they turned it
    on in the first place is a mystery to me. In any case it should be
    safe to turn off for our images, so that debugfs (and hence
    mender-artifact) is able to handle the image.
    
    If anyone requires this feature, it can be flipped on again on the
    built image by a recent version of e2fsprogs/tune2fs.
    
commit 84805fae0be5272a04b5b334399384dcb7232154
Date:   Thu Nov 29 11:08:01 2018

    Fix regression that removed boot flag from boot partition.
    
    This was unintentionally removed in 0b737428d0570de.
    
    Changelog: Title
    
    Changelog: Fix regression that caused Beaglebone to not boot.
    
```

### DIFF
--- a/meta-mender-core/classes/mender-part-images.bbclass
+++ b/meta-mender-core/classes/mender-part-images.bbclass
@@ -148,7 +148,7 @@ EOF
     if [ "${MENDER_BOOT_PART_SIZE_MB}" -ne "0" ]; then
         mender_merge_bootfs_and_image_boot_files
         cat >> "$wks" <<EOF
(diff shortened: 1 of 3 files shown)


kacf:
@mirzak, @drewmoseley: You can't see any reason why removing the `metadata_csum` filesystem feature should cause any problems, do you? It's recently introduced, and the old ones were working well. It's blocking the release, so it's quite important to fix.

drewmoseley:
I've never even heard of it so as long as the tests pass I'd say it's good to go.

PR #665, opened 2018-11-30
Fix boot directory being excluded from thud images

mdavis777:
In thud the entire boot directory is excluded because the variable does not end in a /.
This causes mounts to fail.

Changelog: Title


kacf:
Thanks!

d99roric:
It would be appreciated if this lands in the thud branch soon too :)

PR #669, opened 2018-12-08
part-images: add missing u-boot:deploy dependency for ARM

mirzak:
I am not 100 % sure this is the correct solution, but could not think of a better one without hacking around to much :). See commit message for more information.


meta-mender-core/classes/mender-part-images.bbclass
```diff
@@ -249,8 +249,12 @@ _MENDER_PART_IMAGE_DEPENDS = " \
...
 "
 
-_MENDER_PART_IMAGE_DEPENDS_append_mender-uboot = " u-boot:do_deploy"
-_MENDER_PART_IMAGE_DEPENDS_append_mender-grub = " grub-efi:do_deploy"
+# This assumes that U-boot is used on ARM, this could become problematic
+# if we add support other bootloaders on ARM, e.g Barebox.
+_MENDER_PART_IMAGE_DEPENDS_append_arm =     " u-boot:do_deploy"
+_MENDER_PART_IMAGE_DEPENDS_append_aarch64 = " u-boot:do_deploy"
+
+_MENDER_PART_IMAGE_DEPENDS_append_mender-grub = " grub-efi:do_deploy "
```
kacf: We should keep the `_mender-uboot` variant in addition to the `_arm` and `_aarch64` variants, because it is required for U-Boot on x86. I don't know if anyone is using it, but I can see some potential use cases, such as strict non-GPL3 requirements.

mirzak:
> but thinking about future compatibility, do you think it makes sense to add a mender-uboot-uefi-provider feature?

Yes, I do think that this best describes our intentions.

meta-mender-core/classes/mender-part-images.bbclass
mirzak: If I update to keep the `_mender-uboot` variant, would that be an acceptable quick fix?

I have not really looked in to what the effort would be to add a new feature.

mirzak:
Pushed an update

PR #673, opened 2018-12-17
MEN-2224: Updated the license checksum in mender-artifact

oleorhagen:
Changelog: None

### DIFF
--- a/meta-mender-core/recipes-mender/mender/mender_git.bb
+++ b/meta-mender-core/recipes-mender/mender/mender_git.bb
@@ -59,7 +59,7 @@ def mender_license(branch):
         }
     else:
         return {
-                   "md5": "debbe5e440f2e65465e86b25fc7c9fcc",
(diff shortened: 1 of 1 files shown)


meta-mender-core/recipes-mender/mender/mender_git.bb
```diff
@@ -59,7 +59,7 @@ def mender_license(branch):
         }
     else:
         return {
-                   "md5": "debbe5e440f2e65465e86b25fc7c9fcc",
+                   "md5": "0f993507ab8e2f5afa812bf033a4a24b",
```
kacf: You say mender-artifact, but you changed mender! :smile:

oleorhagen: Well, yeh. Will update it :see_no_evil: . It is mender-artifact vendoring that has changed in mender. However, it is still not correct it seems.

PR #677, opened 2018-12-20
MEN-2290: Fix data directory not being empty on rootfs.

kacf:
Turns out that argparse works a bit strangely, and wants all arguments
after one flag, instead of one argument per flag. For example:

  --exclude-path data/ boot/efi/

instead of:

  --exclude-path data/ --exclude-path boot/efi/

Changelog: Title


kacf:
Should be tested together with https://github.com/mendersoftware/mender-image-tests/pull/4.

kacf:
Will test this a bit later. We need to fix a checksum problem first.

PR #679, opened 2018-12-28
MEN-2145: Split mender.conf configuration file into /etc and /data.

kacf:


kacf:
To be tested with https://github.com/mendersoftware/mender-image-tests/pull/5.

PR #683, opened 2019-01-07
mender-helpers.bbclass: Add NVMe support

bogdanovs:
Add NVMe devices to the list of valid devices for GRUB.

Signed-off-by: Stoyan Bogdanov <stoyan.bogdanov@konsulko.com>
@trini

### DIFF
--- a/meta-mender-core/classes/mender-helpers.bbclass
+++ b/meta-mender-core/classes/mender-helpers.bbclass
@@ -42,6 +42,11 @@ get_grub_device_from_device_base() {
             dev_number=${dev_number%p}
             echo "hd$dev_number"
             ;;
+        /dev/nvme[0-9]n[0-9]p)
+            dev_number=${1#/dev/nvme?n}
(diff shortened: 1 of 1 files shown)


meta-mender-core/classes/mender-helpers.bbclass
```diff
@@ -42,6 +42,11 @@ get_grub_device_from_device_base() {
             dev_number=${dev_number%p}
             echo "hd$dev_number"
             ;;
+        /dev/nvme[0-9]n[0-9]p)
+            dev_number=${1#/dev/nvme?n}
+            dev_number=$(expr $(printf "%d" "'$dev_number") - $(printf "%d" "'1") || true)
```
kacf: This will end up subtracting one from the number after the `n`, which I think is correct, but I just want to double check: That is what GRUB expects, right?

bogdanovs: Yes exactly this is what grub expects.

PR #686, opened 2019-01-18
grub: ensure "test" module is builtin

mirzak:
Mender generates a grub.cfg that utilizes `[` which is an alias for `test` and
we should probably make sure that it is builtin when grub-mender.inc is used.

This has probably gone unnoticed because the upstream recipe includes this by
default. 

I experienced problems when working with a downstream grub-efi recipe,
which did not include the `test` module in GRUB_BUILDIN. 

Changelog: Title

### DIFF
--- a/meta-mender-core/recipes-bsp/grub/grub-mender.inc
+++ b/meta-mender-core/recipes-bsp/grub/grub-mender.inc
@@ -1,7 +1,7 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/patches:"
 
 # Mender needs these.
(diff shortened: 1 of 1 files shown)


kacf:
@mirzak: Just FYI, if you want to fasttrack this PR, can you resubmit it using a branch from your own repository instead of a branch from the official repository?

Long story, there is some Git branch magic going on that barfs on this PR and I haven't had the time to fix it due to the update module demands.

mirzak:
Oh, will create from my own fork. Probably a mistake resulted in this coming from the main repository.

mirzak:
Superseded by #698

PR #689, opened 2019-01-24
add 'rootwait' to bootargs

mirzak:
Description of 'rootwait' in kernel docs,

    "Wait (indefinitely) for root device to show up.
     Useful for devices that are detected asynchronously
     (e.g. USB and MMC devices)."

The large majority of devices that we integrate use eMMC for storage
medium, and it is quite common to use a USB "live disk" for provisioning.

Without the 'rootwait' argument, mount of rootfs failes occasionally,
it is simple a race condition at boot and you are only lucky if it works
without it.

Changelog: Title


mirzak:
Change to master created here https://github.com/mendersoftware/grub-mender-grubenv/pull/7

drewmoseley:
LGTM

PR #690, opened 2019-01-24
remove meta-mender-toradex-nxp layer.conf

mirzak:
This looks like it has been added by mistake. We have removed the toradex layers, as
they should be in meta-mender-community instead.

Changelog: None


mirzak:
Thud only, sumo has the actual layers there

kacf:
Thanks!

kacf:
Will cherry-pick this later as part of normal procedure.

PR #688, opened 2019-01-22
Added bitbake variable  to artifact class to override mender-artifact image

dellgreen:
Useful when wanting to supply a different image to mender-artifact  than the one used to create the sdimg
  - Issue: #MEN-2333 : https://tracker.mender.io/browse/MEN-2333

Signed-off-by: Dell Green <dell.green@ideaworks.co.uk>

PR #695, opened 2019-02-08
Added bitbake variable to  artifact image class to override image pas…

kacf:
…sed to mender-artifact.

  - Issue: #MEN-2333

Changelog: mender artifact bbclass image name override variable added.

Signed-off-by: Dell Green <dell.green@ideaworks.co.uk>


kacf:
Rebased version of #688.

PR #696, opened 2019-02-08
Added bitbake variable to add optional swap partition

dellgreen:
- Issue: #MEN-2361

Changelog: Title

Signed-off-by: Dell Green <dell.green@ideaworks.co.uk>


dellgreen:
https://tracker.mender.io/browse/MEN-2361

dellgreen:
Docs: https://github.com/mendersoftware/mender-docs/pull/601

dellgreen:
On reflection, maybe the swap partition should come after the A/B partitions but before the data partition so that the data partition can be expanded to fill storage if required (see other feature requests)

dellgreen:
will update on monday with swap partition ordered before the data partition so that this makes the data partition available for those people who wish to dynamically resize the last partition to fill the storage device.

dellgreen:
This should be complete now

meta-mender-core/classes/mender-helpers.bbclass
```diff
@@ -227,3 +227,12 @@ mender_get_clean_kernel_devicetree() {
...
 }
+
+
+def mender_get_data_part_num(d):
+    n = 3
+    boot_part_size = d.getVar('MENDER_BOOT_PART_SIZE_MB')
+    swap_part_size = d.getVar('MENDER_SWAP_PART_SIZE_MB')
+    if (boot_part_size and boot_part_size != '0'): n += 1
+    if (swap_part_size and swap_part_size != '0'): n += 1
+    return n
```
kacf: Won't the data partition become partition 6 if the swap partition is added? I think making the swap partition will push the partitions over the four partitions limit, which will trigger an extended partition number 4, swap partition number 5, and data partition number 6.

At least this is the case for MBR partition tables, but I'm not sure if the same thing applies to GPT partition tables. Were you using x86 for your testing?

dellgreen: Yes this problem was experiences and discussed in a unrelated issue yesterday by someone else

https://hub.mender.io/t/meta-mender-add-new-partition/213

Which made me release this would also be a problem for this patch. Is seems to be a general problem in the mender code that doesn't account for extended partition creation. 

I will be looking into this on Friday as it blocks further tickets related to adding extra partitions and is tripping everyone up who is trying to add extra partitions into mender. 

It doesn't apply for GPT as my Friday project is x86_64 uefi project using GPT and mender and it works fine.

kacf: Alright. All the possible images with their partition table types should be neatly readable [here](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/classes/mender-part-images.bbclass#L232-L243), but unfortunately it's evaluated at shell execution time, which is probably too late for your Python function. It might be time to move the partition table type into a variable somewhere.

dellgreen: exaclty what i was thinking too :+1:

dellgreen: Have made the changes, just waiting until monday to test on real hardware before pushing.

dellgreen:
hope this is enough, i have tested it with msdos partition table on ARM and GPT on x86_64 hardware

dellgreen:
all done

dellgreen:
does this need any further work so it get merged into master?

kacf:
No, apologies, I was gone for quite a few days. There's been a checksum update, so I have rebased the change and resubmitted in #700. I'll just leave this open until that is merged.

dellgreen:
Awesome may thanks :)

kacf:
Merged in #700.

PR #698, opened 2019-02-15
grub: ensure "test" module is builtin

mirzak:
Mender generates a grub.cfg that utilizes `[` which is an alias for `test` and
we should probably make sure that it is builtin when grub-mender.inc is used.

This has probably gone unnoticed because the upstream recipe includes this by
default. 

I experienced problems when working with a downstream grub-efi recipe,
which did not include the `test` module in GRUB_BUILDIN. 

Changelog: Title

### DIFF
--- a/meta-mender-core/recipes-bsp/grub/grub-mender.inc
+++ b/meta-mender-core/recipes-bsp/grub/grub-mender.inc
@@ -1,7 +1,7 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/patches:"
 
 # Mender needs these.
-GRUB_BUILDIN_append_mender-grub = " cat echo gcry_sha256 halt hashsum loadenv sleep reboot"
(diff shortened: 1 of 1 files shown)


kacf:
It's red now, but should trigger a new build.

mirzak:
Yeah I see that it has started

kacf:
Looks like you also need to rebase (there was a checksum update recently).

mirzak:
> E           Commit 6584d532e6f6c642db13c757a3a2eacbffbb56ba has less than three words in its changelog tag! Typo?

Really? Is the parser confused by usage of `" "` in the commit message/changelog tag?

kacf:
I have no clue what causes it. Simply running `git commit --amend`, and not changing anything fixes the issue... Maybe you have some weird unicode character in there? Or if this commit was made on Github, I suspect it might have CRLF line endings. After all, it is Microsoft! ;-)

mirzak:
Yes it was initially created the web interface! Damn you Microsoft and CRLF.

Will clean it up.

PR #702, opened 2019-03-12
Update modules support code

kacf:


kacf:
@mirzak: Check the `mender-data-dir.service` file.

mirzak:
> @mirzak: Check the mender-data-dir.service file.

Checked it, nice one!

meta-mender-core/recipes-mender/mender/mender.inc
```diff
@@ -47,6 +44,10 @@ PACKAGECONFIG_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-grub', '
 PACKAGECONFIG[mender-install] = ",,,mender-artifact-info ca-certificates"
 PACKAGECONFIG[u-boot] = ",,,u-boot-fw-utils"
 PACKAGECONFIG[grub] = ",,,grub-editenv grub-mender-grubenv"
+# The docker module depends on bash, and of course on docker. However, docker is
+# a very large requirement, which we will not mandate. Bash however, we require,
+# because otherwise the Yocto QA checks will complain.
+PACKAGECONFIG[install-modules] = ",,,bash"
```
mirzak: This is really a minor/nitpick, but I reacted to the name `install-modules` and found it a bit odd/confusing.

My opinion only but `modules` or `mender-modules` would be more consistent with the options we have.

kacf: When I wrote it I had the exact same feeling, but I said to myself: "Nah, you're just being paranoid!". But since you noticed it too, I will change it! :-)

kacf:
Updated.

PR #703, opened 2019-03-12
bugfix to prevent yocto error when rebuilding image

xeniter:
small fix for yocto rebuilds

### DIFF
--- a/meta-mender-core/classes/mender-ubimg.bbclass
+++ b/meta-mender-core/classes/mender-ubimg.bbclass
@@ -72,7 +72,7 @@ vol_name=rootfsb
 
 [data]
 mode=ubi
-image=${IMGDEPLOYDIR}/${IMAGE_NAME}.dataimg
+image=${IMGDEPLOYDIR}/${IMAGE_BASENAME}-${MACHINE}.dataimg
 vol_id=2
 vol_size=${MENDER_DATA_PART_SIZE_MB}MiB
 vol_type=dynamic


meta-mender-core/classes/mender-ubimg.bbclass
```diff
@@ -72,7 +72,7 @@ vol_name=rootfsb
 
 [data]
 mode=ubi
-image=${IMGDEPLOYDIR}/${IMAGE_NAME}.dataimg
+image=${IMGDEPLOYDIR}/${IMAGE_BASENAME}-${MACHINE}.dataimg
```
kacf: Shouldn't this match what is in `IMAGE_CMD_dataimg_mender-image-ubi`?

xeniter: yes, however IMAGE_NAME has timestamp data in it, this causes an error if you make a rebuild with yocto (it will not find it, because the timestamp changed). The only solution till now is to make a complete clean rebuild, which takes time.

With ${IMAGE_BASENAME}-${MACHINE} it points to the symlink which points to the ${IMAGE_NAME}.dataimg

kacf: :+1:

xeniter:
changed the commit message like requested

xeniter:
any idea why test complains its not signed of?
Commit 91fdbfc4968fa37a8c5edc8981d9a5118df1b77f is not signed off! Use --signoff with your commit.
I used amend -s and see the " Signed-off-by..." in the commit message

mirzak:
> Manuel Dipolt authored and xeniter committed 9 days ago

Probably because of this, the author of the commit is not the same as the sign-off tag.

You can probably fix this with:

     git commit --amend --author="Author Name <email@address.com>"

And make sure that the signoff tag is the same as above

kacf:
Yes, that's it. Let me know when you've changed it. Jenkins builds still need to be started manually.

xeniter:
change now the signoff author to xeniter

xeniter:
Sorry got confused, assumed name should fit to the github user name, should work now

PR #705, opened 2019-03-20
Feat btrfs

timothyfroehlich:
The goal of this PR is to allow the data partition to be written a btrfs partition.

This is controlled by the MENDER_DATA_PART_FSTYPE variable. If unset, it should stay defaulted to auto and allow data partitions that were generated outside the system to be automatically mounted. If set to btrfs, it will generate the data partition as such, rather than ARTIFACT_IMG_TYPE

I've also added FSOPTS and FSTAB_OPTS to set the filesystem options and fstab options.

Example block I have in my config file:

MENDER_DATA_PART_FSTYPE = "btrfs"
MENDER_DATA_PART_SIZE_MB = "512"
MENDER_DATA_PART_FSOPTS = " -d dup -m dup"
MENDER_DATA_PART_FSTAB_OPTS = "defaults,noatime,sync,ssd_spread"


meta-mender-core/classes/mender-dataimg.bbclass
```diff
@@ -2,9 +2,25 @@
 # filesystem.
 
 IMAGE_CMD_dataimg() {
-    dd if=/dev/zero of="${WORKDIR}/data.${ARTIFACTIMG_FSTYPE}" count=0 bs=1M seek=${MENDER_DATA_PART_SIZE_MB}
-    mkfs.${ARTIFACTIMG_FSTYPE} -F "${WORKDIR}/data.${ARTIFACTIMG_FSTYPE}" -d "${IMAGE_ROOTFS}/data" -L data
-    install -m 0644 "${WORKDIR}/data.${ARTIFACTIMG_FSTYPE}" "${IMGDEPLOYDIR}/${IMAGE_NAME}.dataimg"
+    [ ${MENDER_DATA_PART_FSTYPE} = "auto" ] && DATA_FSTYPE=${ARTIFACTIMG_FSTYPE} || DATA_FSTYPE=${MENDER_DATA_PART_FSTYPE}
+    echo $DATA_FSTYPE
```
kacf: Debug print.

kacf:
Also, Github doesn't tell me if you force push, so leave a comment too.

meta-mender-core/classes/mender-dataimg.bbclass
```diff
@@ -2,9 +2,21 @@
 # filesystem.
 
 IMAGE_CMD_dataimg() {
-    dd if=/dev/zero of="${WORKDIR}/data.${MENDER_DATA_PART_FSTYPE}" count=0 bs=1M seek=${MENDER_DATA_PART_SIZE_MB}
-    mkfs.${MENDER_DATA_PART_FSTYPE} -F "${WORKDIR}/data.${MENDER_DATA_PART_FSTYPE}" -d "${IMAGE_ROOTFS}/data" -L data "${MENDER_DATA_PART_FSOPTS}"
-    install -m 0644 "${WORKDIR}/data.${MENDER_DATA_PART_FSTYPE}" "${IMGDEPLOYDIR}/${IMAGE_NAME}.dataimg"
+    [ ${MENDER_DATA_PART_FSTYPE} = "auto" ] && DATA_FSTYPE=${ARTIFACTIMG_FSTYPE} || DATA_FSTYPE=${MENDER_DATA_PART_FSTYPE}
```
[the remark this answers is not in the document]
timothyfroehlich: I was trying to preserve existing functionality, though I just realized what I did isn't going to work and will break the wks file if you don't set DATA_PART_FSTYPE. (I posted this as a draft to make sure I was on the right track). 

MENDER_DATA_PART_FSTYPE is currently defined as "This configuration is only used in fstab. Most filesystems can be auto detected, but some can not and hence this variable exists to override the auto detection.".

I'm assuming the functionality describe above is worth preserving, right? @mirzak?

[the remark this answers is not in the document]
timothyfroehlich: Ok, I've resolved this one by adding this to mender-setup:

`MENDER_DATA_PART_FSTYPE_TO_GEN ??= "${@bb.utils.contains('MENDER_DATA_PART_FSTYPE', 'auto', '${ARTIFACTIMG_FSTYPE}', '${MENDER_DATA_PART_FSTYPE}', d)}"
`

Force pushing my changes. I've tested both setting DATA_PART_FSTYPE to "btrfs" and leaving it unset. The sdimage boots / mounts the data partition and I can mount the .dataimg file on my machine without any issues (mounting as btrfs or ext4, with PART_FSTYPE set and unset respectively)

timothyfroehlich:
Ok, should be ready to go. I think I've done everything properly, but this will be my first open source contribution so please let me know if there's something I need to fix.

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -62,11 +62,27 @@ MENDER_DATA_PART_DEFAULT = "${MENDER_STORAGE_DEVICE_BASE}${MENDER_DATA_PART_NUMB
...
+# Filesystem type of data partition. Used for both FS generation and fstab construction
+# Leave as default (auto) to generate a partition using the same Filesystem as 
+# the rootfs ($ARTIFACTIMG_FSTYPE) and set the fstab to auto-detect the partition type
+# Set to a known filesystem to generate the partition using that type
 MENDER_DATA_PART_FSTYPE ??= "${MENDER_DATA_PART_FSTYPE_DEFAULT}"
 MENDER_DATA_PART_FSTYPE_DEFAULT = "auto"
+
+# Filesystem type of data partition to generate. Used only for FS generation
+# Typically you'll be best off setting MENDER_DATA_PART_FSTYPE instead
+MENDER_DATA_PART_FSTYPE_TO_GEN ??= "${@bb.utils.contains('MENDER_DATA_PART_FSTYPE', 'auto', '${ARTIFACTIMG_FSTYPE}', '${MENDER_DATA_PART_FSTYPE}', d)}"
```
kacf: Use the `DEFAULT` pattern that the other variables use, please.

The reason we do it this way is because:

```
MENDER_DATA_PART_FSTYPE_TO_GEN ??= "wrong"
MENDER_DATA_PART_FSTYPE_TO_GEN_arm ??= "also-wrong"
MENDER_DATA_PART_FSTYPE_TO_GEN = "right"

MENDER_DATA_PART_FSTYPE_TO_GEN evaluates to "also-wrong"
```

However, with the `DEFAULT` pattern:

```
MENDER_DATA_PART_FSTYPE_TO_GEN ??= "${MENDER_DATA_PART_FSTYPE_TO_GEN_DEFAULT}"
MENDER_DATA_PART_FSTYPE_TO_GEN_DEFAULT = "wrong"
MENDER_DATA_PART_FSTYPE_TO_GEN_DEFAULT_arm = "also-wrong"
MENDER_DATA_PART_FSTYPE_TO_GEN = "right"

MENDER_DATA_PART_FSTYPE_TO_GEN evaluates to "right"
```

meta-mender-core/classes/mender-dataimg.bbclass
```diff
@@ -2,9 +2,22 @@
...
+    else #Assume ext3/4
+        force_flag="-F"
+        root_dir_flag="-d"
+    fi
+
+    dd if=/dev/zero of="${WORKDIR}/data.${MENDER_DATA_PART_FSTYPE_TO_GEN}" count=0 bs=1M seek=${MENDER_DATA_PART_SIZE_MB}
+    mkfs.${MENDER_DATA_PART_FSTYPE_TO_GEN} \
+        ${force_flag} \
+        "${WORKDIR}/data.${MENDER_DATA_PART_FSTYPE_TO_GEN}" \
+        ${root_dir_flag} "${IMAGE_ROOTFS}/data" \
```
kacf: Didn't spot this the first time around: Use `$force_flag` notation for shell variables, not `${force_flag}`. Reduces chance that it will clash with Bitbake parsing.

Same goes for `$root_dir_flag`

meta-mender-core/classes/mender-setup.bbclass
timothyfroehlich: I'm not sure I completely grasp why it works like that, (and I'm a bit confused because I don't see any suffixes on the DEFAULTS in mender-setup) but I'll make the change.

timothyfroehlich: Oh, wait, it wasn't that hard, it's just that the override takes precedence and using the second version is just prep for when we'd want to set the default for certain overrides, even if we're not doing so yet.

meta-mender-core/classes/mender-dataimg.bbclass
timothyfroehlich: Got it. I'm guessing this is taking advantage of the fact that bitbake only recognizes variables with the {}?

timothyfroehlich:
Ok, should be good. I believe I've fixed up my name in the config file correctly so that commit -s works (it was just set to Tim before). Had to --reset-author to fix the commit. Is that what you're looking for?

meta-mender-core/classes/mender-dataimg.bbclass
kacf: Correct!

timothyfroehlich:
Sounds good :) When can I expect it to be merged in?

kacf:
Now! :-)

Thanks for the contribution!

timothyfroehlich:
You're welcome! Thanks for the change requests and patience! Contributing to open source has always been an intimidating thing for me but this was a very positive experience :)

kacf:
I'm glad to hear it!

PR #707, opened 2019-03-26
allow IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET to be aligend to 512 bytes

mirzak:
This is a partial back-port of:

     33225086db3 ("MEN-1845: Implement x86 BIOS support together with GRUB")

Which removed the limitation that the IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET must
be 1kB aligned. It is useful to get rid of this limitation as there are some
boards/SoCs that require the bootloader to be at the first block (512 bytes / block)

Changelog: Title

### DIFF
--- a/meta-mender-core/classes/mender-part-images.bbclass
+++ b/meta-mender-core/classes/mender-part-images.bbclass
@@ -79,11 +79,20 @@ mender_part_image() {
         install -m 0644 "${DEPLOY_DIR_IMAGE}/${IMAGE_BOOTLOADER_FILE}" "${WORKDIR}/"
 
         if [ $(expr ${IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET} % 2) -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


kacf:
@mirzak: Tip: If you want to avoid the license errors in Jenkins, you can run this manually with the release tool and a released version. This avoids the unstable md5sums in the Git version:

```
integration/extra/release_tool.py -b 1.7.0 --pr meta-mender/707
```

kacf:
You might have to change some other parameters too, to match rocko, such as `POKY_REV`.

mirzak:
> integration/extra/release_tool.py -b 1.7.0 --pr meta-mender/707

What will this do? Create a commit?

drewmoseley:
Once we retry with the jenkins changes and go all green I'll merge this.

mirzak:
The retry is in progress, only acceptance tests left

mirzak:
The 1.7.0 builds/tests are all green.

PR #708, opened 2019-03-26
Thud update with Mender 2.0.0b1

kacf:


kacf:
The two failures are known and [will be fixed before the final](https://tracker.mender.io/browse/MEN-2438). Everything else green. Merging!

lluiscampos:
:tada:

PR #711, opened 2019-03-29
mender-artifact (3.0.0b1): add missing build dep on "xz"

mirzak:
There is one lite quirk here that I have not handled.

The run-time dependency "liblzma" is not available as a native package, not sure how to handle this.

### DIFF
--- a/meta-mender-core/recipes-mender/mender-artifact/mender-artifact_3.0.0b1.bb
+++ b/meta-mender-core/recipes-mender/mender-artifact/mender-artifact_3.0.0b1.bb
@@ -21,3 +21,5 @@ DEFAULT_PREFERENCE = "-1"
 
 LICENSE = "Apache-2.0 & BSD-2-Clause & BSD-3-Clause & ISC & MIT"
 LIC_FILES_CHKSUM = "file://src/github.com/mendersoftware/mender-artifact/LIC_FILES_CHKSUM.sha256;md5=3c56d1f34c03d02dc3001a58712e0b7c"
+
+DEPENDS += "xz"


meta-mender-core/recipes-mender/mender-artifact/mender-artifact_3.0.0b1.bb
```diff
@@ -21,3 +21,5 @@ DEFAULT_PREFERENCE = "-1"
 
 LICENSE = "Apache-2.0 & BSD-2-Clause & BSD-3-Clause & ISC & MIT"
 LIC_FILES_CHKSUM = "file://src/github.com/mendersoftware/mender-artifact/LIC_FILES_CHKSUM.sha256;md5=3c56d1f34c03d02dc3001a58712e0b7c"
+
+DEPENDS += "xz"
```
kacf: Hmm, I wonder why this worked for me..? Did you get an error message?

In any case, should we also list this here, like the client does:

```
RDEPENDS_${PN} += "liblzma"
```

mirzak: > Hmm, I wonder why this worked for me..? Did you get an error message?

Yes, I got an error message,

```
| ../../../../../../../build/src/github.com/mendersoftware/mender-artifact/vendor/github.com/mendersoftware/go-liblzma/reader.go:9:18: fatal error: lzma.h: No such file or directory
|  #include <lzma.h>
|                   ^
| compilation terminated.
```

> In any case, should we also list this here, like the client does:

Unfortunately `RDEPENDS_${PN} += "liblzma` wont work as it does not provide a native package when we build `mender-artifact-native` This

```
ERROR: Nothing RPROVIDES 'liblzma-native' (but virtual:native:/home/mirza.krak/hub/build/../sources/meta-mender/meta-mender-core/recipes-mender/mender-artifact/mender-artifact_3.0.0b1.bb RDEPENDS on or otherwise requires it)
NOTE: Runtime target 'liblzma-native' is unbuildable, removing...
Missing or unbuildable dependency chain was: ['liblzma-native']
ERROR: Required build target 'core-image-base' has no buildable providers.
Missing or unbuildable dependency chain was: ['core-image-base', 'mender-artifact-native', 'liblzma-native']
```

I am not sure how to handle this (I mentioned this n the PR text as well :D)

kacf: > Yes, I got an error message,
> ```
> | ../../../../../../../build/src/github.com/mendersoftware/mender-artifact/vendor/github.com/mendersoftware/go-liblzma/reader.go:9:18: fatal error: lzma.h: No such file or directory
> |  #include <lzma.h>
> |                   ^
> | compilation terminated.
> ```

I guess my system header leaked into the build somehow. I thought that wasn't supposed to happen.

> I am not sure how to handle this (I mentioned this n the PR text as well :D)

Right, sorry, I skimmed over it too fast! I think we probably don't need to handle it then, since the runtime liblzma is basically everywhere, and it runs fine with the system version.

kacf:
I'll get this into the Git recipe as well!

PR #713, opened 2019-04-15
Sumo backports

drewmoseley:
This brings sumo in-line with the thud-2019.03 tag


kacf:
> tests: Add mender-image-tests as a submodule

I'm not sure if this is a good idea. This new repository has never been tested with older branches, and the idea was to use this only for thud and newer. The same tests already exist as plain files in sumo, so I would consider keeping them instead.

I'm also wondering whether mender-2.0 and mender-artifact-3.0 are things we want to add to sumo. They are new major versions that break compatibility with the older versions (arguments for example). IMHO this is not appropriate for a stable branch. I think we also save quite a bit of effort by omitting them, since it was tricky to get tests to pass for both editions. In thud we will add them, but won't make them the default, as Mirza and I discussed yesterday.

kacf:
Btw, I think that the build failure is due to Yocto trying to build the tests, where we apparently use some functionality that the Go version in sumo doesn't have. There are two ways to deal with this: Turn off the `compile_ptest_base` stage (I think there is a way, or you can simply include a dummy definition of that Bitbake function). Or, if you don't include mender-2.0 and mender-artifact-3.0, and test with 1.7 using the release tool, then the problem should never show up, since it's not present there.

drewmoseley:
I'm fine dropping the the mender-image-tests stuff. I added it here because it applied cleanly but if you think it's unneeded, I'll drop it.

As for Mender versions, we've always kept the old branches up to date with the latest and I'm concerned that users will be left out.  Vendor BSPs are known to be several versions out of date and if we _require_ thud or later to use Mender 2.0, that will be a severely limiting restriction.  For instance, I know that Technexion, Compulab, Variscite and Toradex all are older than thud at present.

drewmoseley:
I added an agenda item for tomorrow's client team meeting to discuss versioning

drewmoseley:
@kacf @lluiscampos I added https://github.com/mendersoftware/meta-mender/pull/713/commits/17f456db78c45b9a42bf01cff12197a601c3b0ae which at least seems to allow the test to pass although I suspect that's not ultimately how we want to handle this.

lluiscampos:
Hi @drewmoseley,

As discussed in Client meeting, I will be taking over the back-porting responsibility so that we can offload you a bit.

I had a look at the failures together with @kacf and we think we found the issue. I pushed into your fork directly, see commit 3aac6ed, and started again integration tests.

lluiscampos:
@kacf @drewmoseley Can you please look carefully to my last commit 40b24f5? I am not very confident in my assessment.

drewmoseley:
@lluiscampos I don't see anything obviously wrong but I'm not the right one to comment on test infrastructure. Definitely defer to @kacf for that.

kacf:
> Can you please look carefully to my last commit 40b24f5? I am not very confident in my assessment.

That is indeed correct. It should not have been included, but probably was due to incorrect conflict resolution. 0b737428d0570de16935e0758db7752de4134d2a is the origin of that test, which is not cherry-picked to sumo.

lluiscampos:
Exactly, that is what I thought.

All acceptance and integration tests have passed now. Both here and in the rocko backport PR. @kacf approve?

lluiscampos:
Cherry-picked the last fix from PR #716. Running tests again.

PR #719, opened 2019-04-25
mender: Do not exclude missing directories.

drewmoseley:
In thud and newer, when processing the exclude_path logic, if the directory
to be excluded does not exist, then wic will error out.  Only list the
entries if they exist.

Changelog: Title

### DIFF
--- a/meta-mender-core/classes/mender-part-images.bbclass
+++ b/meta-mender-core/classes/mender-part-images.bbclass
@@ -158,12 +158,15 @@ EOF
 
     alignment_kb=$(expr ${MENDER_PARTITION_ALIGNMENT} / 1024)
 
+    # remove leading and trailing spaces
+    IMAGE_BOOT_FILES_STRIPPED=$(echo "${IMAGE_BOOT_FILES}" | sed -r 's/(^\s*)|(\s*$)//g')
+
     if [ "${MENDER_BOOT_PART_SIZE_MB}" -ne "0" ]; then
         mender_merge_bootfs_and_image_boot_files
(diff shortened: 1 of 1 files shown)


meta-mender-core/classes/mender-part-images.bbclass
```diff
@@ -158,12 +158,15 @@ EOF
...
+    # remove leading and trailing spaces
+    IMAGE_BOOT_FILES_STRIPPED=$(echo "${IMAGE_BOOT_FILES}" | sed -r 's/(^\s*)|(\s*$)//g')
+    
     if [ "${MENDER_BOOT_PART_SIZE_MB}" -ne "0" ]; then
         mender_merge_bootfs_and_image_boot_files
         cat >> "$wks" <<EOF
 part --source rootfs --rootfs-dir ${WORKDIR}/bootfs.${BB_CURRENTTASK} --ondisk "$ondisk_dev" --fstype=vfat --label boot --align $alignment_kb --fixed-size ${MENDER_BOOT_PART_SIZE_MB} --active
 EOF
-    elif [ -n "${IMAGE_BOOT_FILES}" ]; then
+    elif [ -n "${IMAGE_BOOT_FILES_STRIPPED}" ]; then
```
kacf: Use `"$IMAGE_BOOT_FILES_STRIPPED"` (no brackets) when referring to shell variables, to avoid potential clashes with Bitbake's own variables.

drewmoseley: Fixed and rebased.

PR #720, opened 2019-04-25
Thud version of #719

drewmoseley:
These patches have been tested only against thud.


kacf:
I suspect the failing build may be fixed with a rebase.

drewmoseley:
This is already rebased on the latest thud branch.

kacf:
Ok, then I'm not sure what's missing. I'm going to do cherry-picks for thud very soon anyway, so I'll just close this for now and investigate as part of that.

PR #723, opened 2019-04-26
Update mender and mender-artifact recipes after branch renames

lluiscampos:
They should use 2.0.x and 3.0.x respectively

Changelog: Fix mender 2.0.x and mender-artifact 3.0.x recipes to use the
correct branches when fetching the source.


lluiscampos:
Spurious error. Merging and porting to thud/sumo/rocko

PR #729, opened 2019-05-02
test_uboot_compile: Make sure all cores are used.

kacf:
```
commit 47d604773c1867ed8f144976597909602de33650
Date:   Thu May 2 13:58:07 2019

    test_uboot_compile: Make sure all cores are used.
    
    Right now we may limit by memory since each build directory is very
    memory consuming (RAM drive). But that may leave cores unused, so make
    sure the sub-make invocations try to use those cores.
    
    Changelog: None
    
commit 6b97f98ded0f3e59b691d6f700aeb6fdb63f0142
Date:   Thu May 2 13:56:38 2019

    test_uboot_compile: Make sure the Makefile is counted in time calc.
    
    Otherwise the test might not run even if the file has changed.
    
    Changelog: None
    
```

### DIFF
--- a/tests/acceptance/files/Makefile.test_uboot_automation
+++ b/tests/acceptance/files/Makefile.test_uboot_automation
@@ -37,12 +37,12 @@ $(LOGS)/%_defconfig:
 		|| ( echo AutoPatchFailed >> $@ )
 
 	# Setup compile.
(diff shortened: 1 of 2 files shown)


kacf:
Exactly!

PR #731, opened 2019-05-02
Break mender_git recipe into pre- and post-2.0 recipes.

kacf:
This is required in order build 1.x recipes using the git recipe,
since it uses the old makefile style and many other conditionals as
well, such as module inclusion.

Changelog: None


kacf:
Spurious failure.

PR #734, opened 2019-05-07
Fix build failing in do_image_ubimg task

jonasn-ge:
Fix/workaround for a build problem. The problem can be reproduced by inserting a long sleep at the start of meta-mender-core/recipes-bsp/u-boot/u-boot-mender.inc do_deploy_append_mender-uboot().


kacf:
Thanks! Cherry-picked to master as well, but please submit to master in the future, since we usually cherry-pick the other way.

jonasn-ge:
> Thanks! Cherry-picked to master as well, but please submit to master in the future, since we usually cherry-pick the other way.

All right, I was wondering how you worked with master vs release branches.

PR #735, opened 2019-05-07
Add recipes for Mender 2.0.0

kacf:


kacf:
Github went down while the build was running, so the status didn't get updated, but this [is passing](https://mender-jenkins.mender.io/job/mender-builder/3344/).

PR #739, opened 2019-05-09
rename boot.efi to bootia32.efi in mender-setup-grub.inc

mirzak:
The correct name is "bootia32.efi" on x86, otherwise it will fail
to boot.

This is an backport of:

5fbb3074913a9 ("FIX: Renamed boot.efi to bootia32.efi in mender-setup-grub.inc")

Changelog: Fix GRUB start-up on x86 by renaming "boot.efi" to "bootia32.efi"


lluiscampos:
Superseded by #744

PR #740, opened 2019-05-09
mender: Setup rootfs partitions in automount blacklist.

drewmoseley:
Changelog: None

### DIFF
--- a/meta-mender-core/recipes-mender/mender/mender-old-makefile.inc
+++ b/meta-mender-core/recipes-mender/mender/mender-old-makefile.inc
@@ -96,4 +96,9 @@ do_install() {
         install -m 755 -d ${D}/data/mender
         install -m 444 ${B}/device_type ${D}/data/mender/
     fi
+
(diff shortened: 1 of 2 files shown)


drewmoseley:
I thought about that but since the udev-extraconf recipe adds the /etc/udev/mount-blacklist.d directory it seems ok to add those files in from extra repos. And if udev-extraconf is not present, then it doesn't hurt anything.

kacf:
Yes, I think it's fairly common for packages to add files to such {{.d}} directories, even if they are not used. That's at least how many deb packages do it.

kacf:
Btw, @drewmoseley don't worry about the test error. It's a long standing issue that I'm finally going to fix now. Not related to your PR.

kacf:
FYI: https://github.com/mendersoftware/meta-mender/pull/741

kacf:
Rebased.

@mirzak: Ok with the `mount-blacklist.d` files, or do you still have concerns?

mirzak:
> @mirzak: Ok with the mount-blacklist.d files, or do you still have concerns?

No concerns

PR #741, opened 2019-05-10
Fix failure when submitting a PR with a branch in the main repo.

kacf:
The problem is that the version detection mistakes the custom branch
for a Yocto branch, so be explicit that only the listed branches are
candidates for Yocto branches.

Changelog: None


kacf:
I explicitly submitted this from a branch in this repo, to test it.

PR #749, opened 2019-05-24
Thud update

kacf:


lluiscampos:
Cherry-pick also the second license update, commit 35842ee, else it will not work for master recipe.

kacf:
Oh right, race condition! :-) Thanks!

lluiscampos:
> Oh right, race condition! :-) Thanks!

Yes, sorry. I did it in two updates by mistake

PR #750, opened 2019-05-24
mender: Add support for a bootimg FSTYPE.

drewmoseley:
This contains the contents of /boot for systems that need it separate.

### DIFF
--- a/meta-mender-core/classes/mender-bootimg.bbclass
+++ b/meta-mender-core/classes/mender-bootimg.bbclass
@@ -1,0 +1,25 @@
+# Class to create the "bootimg" type, which contains the boot partition as a raw
+# filesystem.
+
+inherit mender-helpers
+
+IMAGE_CMD_bootimg() {
+    if [ ${MENDER_BOOT_PART_SIZE_MB} -ne 0 ]; then
+        mender_merge_bootfs_and_image_boot_files
+        rm -f "${WORKDIR}/boot.vfat"
+        dd if=/dev/zero of="${WORKDIR}/boot.vfat" count=0 bs=1M seek=${MENDER_BOOT_PART_SIZE_MB}
+        mkfs.vfat -n "BOOT" "${WORKDIR}/boot.vfat"
+        for i in $(find ${WORKDIR}/bootfs.${BB_CURRENTTASK}/ -mindepth 1 -maxdepth 1); do
+           mcopy -i "${WORKDIR}/boot.vfat" -s "$i" ::/
+        done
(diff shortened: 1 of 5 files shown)


tests/acceptance/test_bootimg.py
```diff
@@ -0,0 +1,33 @@
...
+    @pytest.mark.min_mender_version('1.0.0')
+    def test_bootimg_creation(self, bitbake_variables, prepared_test_build):
+        """Test that we can build a bootimg successfully."""
+
+        add_to_local_conf(prepared_test_build, 'IMAGE_FSTYPES = "bootimg"')
+        run_bitbake(prepared_test_build)
+
+        built_img = latest_build_artifact(prepared_test_build['build_dir'], "core-image*.bootimg")
+
+        # What can we check here to validate this?
```
kacf: I suggest something along the lines of:

```
distro_features = bitbake_variables['DISTRO_FEATURES'].split()
if "mender-grub" in distro_features and "mender-image-uefi" in distro_features:
    output = subprocess.check_output(["mdir", "-i", built_img, "-b", "/EFI/BOOT"])
    assert "mender_grubenv1" in output.split('/')
```

I haven't tested that :-)

meta-mender-core/classes/mender-bootimg.bbclass
```diff
@@ -0,0 +1,11 @@
+# Class to create the "bootimg" type, which contains the boot partition as a raw
+# filesystem.
+
+IMAGE_CMD_bootimg() {
+    dd if=/dev/zero of="${WORKDIR}/boot.${ARTIFACTIMG_FSTYPE}" count=0 bs=1M seek=${MENDER_BOOT_PART_SIZE_MB}
+    mkfs.${ARTIFACTIMG_FSTYPE} -F "${WORKDIR}/boot.${ARTIFACTIMG_FSTYPE}" -d "${IMAGE_ROOTFS}/boot" -L boot
+    install -m 0644 "${WORKDIR}/boot.${ARTIFACTIMG_FSTYPE}" "${IMGDEPLOYDIR}/${IMAGE_NAME}.bootimg"
```
kacf: Should we take `${MENDER_BOOT_PART_SIZE_MB} = 0` into account here, and make nothing?

kacf:
Perfect! Thanks!

> I finally got around to updating this. I had to move the helper function. We call that twice now which is a bit of extra effort but I figure for the time being that's probably ok.

Indeed. Bonus points if you can make wic use the output from this image as its input!

PR #755, opened 2019-06-03
u-boot-fw-utils: Manually build autogenerated files.

drewmoseley:
For certain versions of U-Boot (notably u-boot-firefly 2017.09) these
autogenerated files are not created when building the env utilities.

Changelog: None


meta-mender-core/recipes-bsp/u-boot/u-boot-fw-utils-mender-auto-provided_1.0.bb
```diff
@@ -63,6 +63,7 @@ do_compile () {
         exit 1
     fi
 
+    oe_runmake include/generated/timestamp_autogenerated.h include/generated/version_autogenerated.h
```
kacf: What do you think about adding a `|| true` at the end of this? My thinking is that this recipe is used for a wide array of U-Boot versions, and might break if these exact files are moved or removed. If we add `|| true`, it will still succeed even if the files don't exist, and if they do exist, then real build errors should still be caught in the next `oe_runmake $ENV_TARGET` command.

drewmoseley: Done

kacf:
Great, thanks!

PR #757, opened 2019-06-06
Add MENDER_DTB_NAME_FORCE 

reachBAM:
This new define is to allow builds when no dtb is produced or the dtb is included inside another file, fit image for example.

### DIFF
--- a/meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
+++ b/meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
@@ -95,12 +95,16 @@ do_provide_mender_defines() {
             ;;
     esac
 
-    # Strip leading and trailing whitespace, then newline divide, and remove dtbo's.
-    MENDER_DTB_NAME="$(echo "${KERNEL_DEVICETREE}" | sed -r 's/(^\s*)|(\s*$)//g; s/\s+/\n/g' | sed -ne '/\.dtbo$/b; p')"
+    if [ -n "${MENDER_DTB_NAME_FORCE}" ]; then
+        MENDER_DTB_NAME="${MENDER_DTB_NAME_FORCE}"
+    else
+        # Strip leading and trailing whitespace, then newline divide, and remove dtbo's.
+        MENDER_DTB_NAME="$(echo "${KERNEL_DEVICETREE}" | sed -r 's/(^\s*)|(\s*$)//g; s/\s+/\n/g' | sed -ne '/\.dtbo$/b; p')"
 
-    if [ -z "$MENDER_DTB_NAME" ]; then
-        bbfatal "Did not find a dtb specified in KERNEL_DEVICETREE"
-        exit 1
+        if [ -z "$MENDER_DTB_NAME" ]; then
+            bbfatal "Did not find a dtb specified in KERNEL_DEVICETREE"
+            exit 1
+        fi
     fi
 
     DTB_COUNT=$(echo "$MENDER_DTB_NAME" | wc -l)


meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
```diff
@@ -95,8 +95,12 @@ do_provide_mender_defines() {
             ;;
     esac
 
-    # Strip leading and trailing whitespace, then newline divide, and remove dtbo's.
-    MENDER_DTB_NAME="$(echo "${KERNEL_DEVICETREE}" | sed -r 's/(^\s*)|(\s*$)//g; s/\s+/\n/g' | sed -ne '/\.dtbo$/b; p')"
+    if [ -n "${MENDER_MENDER_DTB_NAME_FORCE}" ]; then
```
lluiscampos: I think we can use the same variable name, MENDER_DBT_NAME.

In any case, let's call it MENDER_DBT_NAME_FORCE and avoid the MENDER_MENDER in the name.

meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
```diff
@@ -95,8 +95,12 @@ do_provide_mender_defines() {
...
+        MENDER_DTB_NAME="${MENDER_MENDER_DTB_NAME_FORCE}"
+    else
+	    # Strip leading and trailing whitespace, then newline divide, and remove dtbo's.
+	    MENDER_DTB_NAME="$(echo "${KERNEL_DEVICETREE}" | sed -r 's/(^\s*)|(\s*$)//g; s/\s+/\n/g' | sed -ne '/\.dtbo$/b; p')"
```
lluiscampos: Please fix the indentation here, we are mixing spaces and tabs

meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc
```diff
@@ -95,8 +95,12 @@ do_provide_mender_defines() {
...
+    fi
 
     if [ -z "$MENDER_DTB_NAME" ]; then
         bbfatal "Did not find a dtb specified in KERNEL_DEVICETREE"
```
lluiscampos: I think this error should be inside the else block above

reachBAM:
Thanks for the feedback everyone.  I have used Git quite a bit but this my first pull request to an Open Source repository.

I have pushed up a new commit to my fork.  Is it best to squash the two commits together?

lluiscampos:
@reachBAM Yes, you have to sqash them for two reasons: because your first commit is still unsigned and have no Changelog tag (so it won't pass the quality tests) and because it will be much cleaner and portable to have your full change in a single commit. :)

lluiscampos:
Hello,

Sorry I missed it in my first round, but this should too be submitted to master and not to rocko. Please close and re-sumbit there.

reachBAM:
Closing this pull request and moving changes to master branch.  New pull request will be submitted soon.

reachBAM:
@lluiscampos 

After closing this pull request and reviewing the master branch I am less sure of the value of this approach.  This section of code is done differently in the master branch and I don't have a means of testing the master branch at the moment.  I anticipate that backporting the changes from master to rocko would be a manually process not a merge or cherry-pick.  

So, is this pull request still valid?

kacf:
It is still valid, but since this is not a clean cherry-pick, the best would be if you could provide both (so this PR can remain targeted to rocko). In general we require all additions to go into the master branch since otherwise upgrading will "remove" features.

Fundamentally it does not look very different in master, you just have to apply the same fix in the new `mender_get_clean_kernel_devicetree` function.

kacf:
Closing due to age.

PR #759, opened 2019-06-11
Allow the machine-id to persist

drewmoseley:
This patchset  passes the machine-id on the kernel command line to systemd. This ensures that it is the same regardless of which partition is /. It also works properly for read-only root.

I have only lightly tested this on Master but I have a version in Thud that has been tested more thoroughly.

### DIFF
--- a/meta-mender-core/classes/mender-setup.bbclass
+++ b/meta-mender-core/classes/mender-setup.bbclass
@@ -259,6 +259,9 @@ python() {
 
         # Use PARTUUID to set fixed drive locations.
         'mender-partuuid',
+
+        # Setup the systemd machine ID to be persistent across OTA updates.
+        'mender-persist-systemd-machine-id',
     }
 
     mfe = d.getVar('MENDER_FEATURES_ENABLE')
--- a/meta-mender-core/recipes-bsp/u-boot/patches/0002-Generic-boot-code-for-Mender.patch
+++ b/meta-mender-core/recipes-bsp/u-boot/patches/0002-Generic-boot-code-for-Mender.patch
@@ -1,4 +1,4 @@
-From c07ad477cd5715079a3bd086293e0adf91f15b88 Mon Sep 17 00:00:00 2001
+From 8ecd3470f58cf8022e3188da3c2eb77448c244cb Mon Sep 17 00:00:00 2001
 From: Drew Moseley <drew.moseley@northern.tech>
 Date: Sun, 28 Jan 2018 17:44:32 -0500
 Subject: [PATCH] Generic boot code for Mender.
@@ -7,15 +7,15 @@ Signed-off-by: Maciej Borzecki <maciej.borzecki@rndity.com>
 Signed-off-by: Drew Moseley <drew.moseley@northern.tech>
 Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>
 ---
- include/config_mender.h |  98 +++++++++++++++++++++++++++++++
- include/env_mender.h    | 152 ++++++++++++++++++++++++++++++++++++++++++++++++
- 2 files changed, 250 insertions(+)
+ include/config_mender.h |  98 +++++++++++++++++++++++++
+ include/env_mender.h    | 157 ++++++++++++++++++++++++++++++++++++++++
+ 2 files changed, 255 insertions(+)
  create mode 100644 include/config_mender.h
  create mode 100644 include/env_mender.h
 
 diff --git a/include/config_mender.h b/include/config_mender.h
 new file mode 100644
-index 0000000..3836a6d
+index 0000000000..3836a6ddd4
 --- /dev/null
 +++ b/include/config_mender.h
 @@ -0,0 +1,98 @@
@@ -119,10 +119,10 @@ index 0000000..3836a6d
 +#endif /* HEADER_CONFIG_MENDER_H */
 diff --git a/include/env_mender.h b/include/env_mender.h
 new file mode 100644
-index 0000000..a1996ab
+index 0000000000..0605f6a9fa
 --- /dev/null
 +++ b/include/env_mender.h
-@@ -0,0 +1,152 @@
+@@ -0,0 +1,157 @@
 +/*
 +  Copyright 2017 Northern.tech AS
 +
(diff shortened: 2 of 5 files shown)


mirzak:
What is the motivation of including this in the "Mender core"?

drewmoseley:
@mirzak according to https://www.freedesktop.org/software/systemd/man/machine-id.html this value should "stays constant for all subsequent boots". It seems to me a bug that we have different ids depending on which rootfs is active.

drewmoseley:
Dang. I just realized I didn't push this branch into my fork but rather into the main fork.  Oh well. I guess we'll just leave it for now.

drewmoseley:
@kacf I squashed the two patches together. I've still got the grub-mender stuff here as a patch so I can test.  I'll submit separately to that repo and when that is merged I can update this with the new commit hash.

I need this fix for thud; should we include the grub-mender-grub-env fix as a patch there or in an appropriate branch?

mirzak:
I think there is a difference between "subsequent boots" and image update and I do not believe that we are violating any rules here.

The `machine-id` is something that is generated on first boot, and imo if a user wants to preserve this id across firmware updates they can either provide this id in the data partition or use a state-script to migrate it, or generate it a build time and provide it as a kernel argument. I do not think that we should include this in the Mender core functionality.

There are other "ids" that we are not migrating such as, ssh identity, dbus machine id (which is referred to by systemd docs as well), and probably other things that I am not aware of. And if we are not preserving them all, then I do not think that should treat the "machine-id" any differently.

drewmoseley:
Well, the dbus uuid is _usually_ a symlink to the systemd machine-id so that one is being handled with this patchset as well.  You are correct that we don't do anything with the ssh key but maybe we should? Since Mender is an updater and not an installer, I don't think of the two images as fundamentally different. I certainly would not expect my IDs to change on my laptop when I ran an "apt-get update".

kacf:
I see Mirza's point, but I think I'm leaning towards the principle of "works out of the box". It's not easy to configure this sort of thing yourself, and Drew's implementation is relatively "soft" in the sense that it does nothing if either the file is missing of you're not using systemd. FWIW, I tried to replace `/etc/machine-id` with a symlink to `/data/machine-id` and all hell broke loose, so it's not easy to go that route.

Machine-id is a standard component of systemd, so I think it's reasonable to take it into account if `mender-systemd` is enabled.

kacf:
@drewmoseley: I think your patch got messed up somehow. There are several unrelated changes in there now...

drewmoseley:
@kacf sorry, I forgot to mention that. I had to make some other updates just to get it to build. Of course I have other locally that are required to build in master now that are not submitted; notably the LAYERSERIES_COMPAT stuff.  How should we handle this since the current HEAD of master won't build for testing? Is the Jenkins stuff using an older commit hash?

kacf:
Our master branch is entirely based on the latest stable poky branch, IOW thud (there is a ticket for warrior, not there yet), which was updated just last week, so it's pretty recent. The difference in our thud and master branch ATM is only our own features, plus the fact that master will eventually become warrior, whereas our thud branch will stay as is.

You mentioned that you had a thoroughly tested thud variant. You can probably submit that to meta-mender/master, since it builds with thud.

drewmoseley:
OK. I force pushed my thud version here. I'll keep the other one locally since there were some minor changes needed to get it to apply but the warrior work may supersede that.

drewmoseley:
Oh, and I created https://github.com/mendersoftware/grub-mender-grubenv/pull/8.  Once that is merged, I can update this to use the new commit hash rather than a local patch.

mirzak:
> FWIW, I tried to replace /etc/machine-id with a symlink to /data/machine-id and all hell broke loos

Yeah would assume this to be problem, but there is a somewhat simple solution to make sure `/data` is mounted before systemd runs.

/sbin/init
```
#!/bin/sh
mount /dev/mmcblk0p4 /data
exec /lib/systemd/systemd "$@"
```

> I see Mirza's point, but I think I'm leaning towards the principle of "works out of the box"

But has it not worked out of the box so far? I understand that the `machine-id` might change between updates, but is this a problem as in breaking functionality? I would not expect it to break anything as each update is considered a "first time boot".

I am not gonna oppose strongly against this, just looking to understand

drewmoseley:
Application code may depend on this. In fact, this came up due to a user application who was using the machine id and it broke their code when it changed.

drewmoseley:
I'm wondering if we _should_ setup things like ssh keys, wifi creds, etc as persistent when Mender is enabled?  Maybe it can be an optional config but I think it would be useful to have since many users need that.

kacf:
> I'm wondering if we *should* setup things like ssh keys, wifi creds, etc as persistent when Mender is enabled? Maybe it can be an optional config but I think it would be useful to have since many users need that.

This is definitely a gray area, and as we can see there are different opinions. I think I would draw the line at things that change the rootfs. And indeed this PR does *not* change the rootfs, we only store one value in the bootloader area, and systemd already has the necessary mechanics to pick this up via the kernel cmdline. For SSH keys we would have to write those keys to the filesystem, which is a bit more intrusive, and possibly in direct conflict with other key arrangements.

OTOH I'm not strongly leaning one way or the other either. I can see pros and cons in either direction. @drewmoseley, have you run into the original issue more than once? One option is to just put this in the "wait and give it some time" pile, and see if this is a recurring issue.

drewmoseley:
I've discussed it with a number of users but in most cases we would handle the changes in their layer. If there were a configurable option in the mender layer it would certainly make it easier for them and more consistent.

For this particular fix, I have a current user who needs this on Thud.

kacf:
Ok, still only one user though. Mirza's primary concern was adding this to the core functionality, so how about providing this under a `MENDER_FEATURE` like `mender-preserve-machine-id` or something like that? This would mean going back to the patch approach that Drew first proposed, instead of adding it directly to grub-mender-grubenv. @mirzak, @drewmoseley, is that an ok solution?

drewmoseley:
I'm fine with that; plus we can then add things like _mender-preserve-ssh-keys_ and _mender-preserve-wifi-credentials_ which I'm sure users would find useful.

Although, I think using _mender-persist-*_ would probably be more self-explanatory.

kacf:
:+1:

mirzak:
> This would mean going back to the patch approach that Drew first proposed, instead of adding it directly to grub-mender-grubenv. @mirzak, @drewmoseley, is that an ok solution?

Makes sense to me.

drewmoseley:
@kacf @mirzak I finally got around to conditionalizing this based on a MENDER_FEATURE. Note that the patch to grub-mender-grubenv is still valid since within the patch it checks if the env variable is set before adding it to the command line. I think we should still consider merging  https://github.com/mendersoftware/grub-mender-grubenv/pull/8 so that it works for non-Yocto builds.

drewmoseley:
What do you think about adding other mender-persist FEATURES? ie mender-persist-wifi-credentials which would modify wpa_supplicant, NetworkManager and connman recipes?

drewmoseley:
Done, although now I guess we need a commit hash update for grub-mender-grubenv.

kacf:
Yes, I have arranged the branches correctly, and am testing in #791.

PR #765, opened 2019-06-19
Ensure that artifact and partition images' checksums are equal.

kacf:
We also switch the data partition to use the same mechanism, mostly as
a free testing mechanism for our dataimg image type.

A test is added for ubimg and the corresponding part image test is in
the mender-image-tests repository.

MEN-2597

Changelog: Title

### DIFF
--- a/meta-mender-core/classes/mender-part-images.bbclass
+++ b/meta-mender-core/classes/mender-part-images.bbclass
@@ -42,19 +42,6 @@ IMAGE_NAME_SUFFIX = ""
 # Block storage
 ################################################################################
 
-def mender_wic_exclude_path_options(rootfs, exclude_path):
-    exclude_path = exclude_path.strip()
-
-    existing_exclude_paths = ""
-    for path in exclude_path.split():
-        if os.path.lexists(os.path.realpath(os.path.join(rootfs, path))):
-           existing_exclude_paths += path + " "
-
-    if existing_exclude_paths:
-        return "--exclude-path %s" % existing_exclude_paths
-    else:
-        return ""
-
 mender_part_image() {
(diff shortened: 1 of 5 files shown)


kacf:
Goes together with https://github.com/mendersoftware/mender-image-tests/pull/7

kacf:
Rerunning after [x86](https://s3-eu-west-1.amazonaws.com/mender-testing-reports/integration-reports-qemux86-64-uefi-grub/3701/report.html) and [arm](https://s3-eu-west-1.amazonaws.com/mender-testing-reports/integration-reports-vexpress-qemu/3701/report.html) failed.

meta-mender-core/classes/mender-part-images.bbclass
```diff
@@ -42,19 +42,6 @@ IMAGE_NAME_SUFFIX = ""
...
-    existing_exclude_paths = ""
-    for path in exclude_path.split():
-        if os.path.lexists(os.path.realpath(os.path.join(rootfs, path))):
-           existing_exclude_paths += path + " "
-
-    if existing_exclude_paths:
-        return "--exclude-path %s" % existing_exclude_paths
-    else:
-        return ""
-
```
mirzak: Does this mean that we are no longer excluding `/boot` and `/data` from the rootfs?

kacf: `rawcopy` uses the `ext4` file, where they should already be removed. The difference is that `wic` is not doing that job anymore.

mirzak: That is even better :)

kacf:
Thanks!

PR #776, opened 2019-07-04
Thud update

kacf:


kacf:
@lluiscampos: FYI, d5a7353 is apparently needed too.

PR #777, opened 2019-07-10
Set LAYERDEPENDS and do not force changes to MENDER_FEATURES_ENABLE/DISABLE

pbrkr:
* Set LAYERDEPENDS for all layers.
* Allow default bootloader to be overridden.
* Do not force changes to MENDER_FEATURES_ENABLE/DISABLE.


meta-mender-core/conf/layer.conf
```diff
@@ -18,3 +18,4 @@ BBFILE_PRIORITY_mender = "6"
 INHERIT += "mender-setup"
 
 LAYERSERIES_COMPAT_mender = "thud"
+LAYERDEPENDS_mender = "core openembedded-layer"
```
kacf: We don't depend on `jq` in general do we? AFAICS it's only for the `mender-migrate-configuration` script.

pbrkr: I used `yocto-check-layer` which threw the error here - that script checks every recipe rather than just what you're trying to build at the time.

I can update this to not depend on `openembeded-layer`, people will just get a nasty error about no providers for `jq` if they try to build `mender-migrate-configuration` without the meta-oe layer included.

kacf: Yes, please do. `mender-migrate-configuration` is a quite uncommon package, so we should not add a hard dependency for it.

pbrkr:
@kacf: I've updated this merge request to address your feedback and pull in the other changes discussed on Mender Hub.

kacf:
Thanks, @pbrkr!

PR #779, opened 2019-07-16
mender: Setup Live installer.

drewmoseley:
The current HDDIMG (and ISO for that matter) support in Yocto sets up multiple
bootloader options. One is standard live boot but the other is an install mechanism
so you can install your build onto a target without removable media. This patch
provides a version of this scripting that works with Mender. Basically, the SDIMG
UEFIIMG or MTDIMG file are copied into the HDDIMG and the initrd runs a script
which uses 'dd' to write the image.

This has been tested on an X86 platform with an internal MSATA drive and an external
USB drive.

### DIFF
--- a/meta-mender-core/classes/mender-part-images.bbclass
+++ b/meta-mender-core/classes/mender-part-images.bbclass
@@ -261,6 +261,33 @@ IMAGE_TYPEDEP_biosimg_append = "${@bb.utils.contains('IMAGE_FSTYPES', 'sdimg', '
 IMAGE_TYPEDEP_gptimg_append = "${@bb.utils.contains('IMAGE_FSTYPES', 'sdimg', ' sdimg', '', d)} \
                                ${@bb.utils.contains('IMAGE_FSTYPES', 'uefiimg', ' uefiimg', '', d)} \
                                ${@bb.utils.contains('IMAGE_FSTYPES', 'biosimg', ' biosimg', '', d)}"
+# Make sure the Mender part image is available in the live installer
+IMAGE_TYPEDEP_hddimg_append = "${@bb.utils.contains('IMAGE_FSTYPES', 'sdimg', ' sdimg', '', d)} \
+                               ${@bb.utils.contains('IMAGE_FSTYPES', 'gptimg', ' gptimg', '', d)} \
+                               ${@bb.utils.contains('IMAGE_FSTYPES', 'uefiimg', ' uefiimg', '', d)} \
+                               ${@bb.utils.contains('IMAGE_FSTYPES', 'biosimg', ' biosimg', '', d)}"
+
+# Use the Mender part image as the Live image
+python() {
+    if bb.utils.contains('IMAGE_FSTYPES', 'sdimg', True, False, d):
+        type='sdimg'
+    elif bb.utils.contains('IMAGE_FSTYPES', 'uefiimg', True, False, d):
+        type='uefiimg'
+    elif bb.utils.contains('IMAGE_FSTYPES', 'biosimg', True, False, d):
+        type='biosimg'
+    elif bb.utils.contains('IMAGE_FSTYPES', 'gptimg', True, False, d):
+        type='gptimg'
+    else:
+        return
+
+    d.setVar('LIVE_ROOTFS_TYPE', type)
+    d.setVar('ROOTFS', "${IMGDEPLOYDIR}/${IMAGE_LINK_NAME}.%s.bz2" % type)
+    d.setVar('IMAGE_FSTYPES_append', ' %s.bz2 ' % type)
+
+    # Remove the boot option on the Live installer; it won't work since Mender hard codes
+    # the device nodes
+    d.setVar('LABELS_LIVE_remove', 'boot')
+}
 
 # So that we can use the files from excluded paths in the full images.
 do_image_sdimg[respect_exclude_path] = "0"
--- a/meta-mender-core/recipes-core/initrdscripts/files/init-install-efi-mender.sh
+++ b/meta-mender-core/recipes-core/initrdscripts/files/init-install-efi-mender.sh
@@ -1,0 +1,40 @@
+#!/bin/sh
+
+set -ue
+PATH=/sbin:/bin:/usr/sbin:/usr/bin
+
+devnode=@MENDER_STORAGE_DEVICE@
+
+if [ -b ${devnode} ]; then
+    echo "Installing image on ${devnode} ..."
+else
+    echo "No device available matching ${devnode}. Installation aborted."
+    exit 1
+fi
+
+if [ -e /run/media/$1/$2 ]; then
+    echo "Writing Mender-enabled Yocto image /run/media/$1/$2 to ${devnode}"
+    echo "bzcat /run/media/$1/$2 | /bin/dd of=$devnode bs=8M"
+    echo -n "OK to proceed? <y/n> "
+    read yn
+    if [ "$yn" = "y" ] || [ "$yn" = "Y" ]; then
+        rc=0
+        bzcat /run/media/$1/$2 | /bin/dd of=$devnode bs=8M || rc=$?
+        sync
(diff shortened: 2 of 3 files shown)


meta-mender-core/recipes-core/initrdscripts/files/init-install-efi-mender.sh
```diff
@@ -0,0 +1,133 @@
+#!/bin/sh -e
+#
+# Copyright (c) 2012, Intel Corporation.
+# All rights reserved.
```
kacf: Where does this file come from? This isn't a file we can redistribute according to this license text. Though if it comes from Yocto, it's pretty likely they've made a mistake.

meta-mender-core/classes/mender-part-images.bbclass
```diff
@@ -261,6 +261,28 @@ IMAGE_TYPEDEP_biosimg_append = "${@bb.utils.contains('IMAGE_FSTYPES', 'sdimg', '
...
+     type='biosimg'
+
+  d.setVar('LIVE_ROOTFS_TYPE', type)
+  d.setVar('ROOTFS', "${IMGDEPLOYDIR}/${IMAGE_LINK_NAME}.%s.bz2" % type)
+  d.setVar('IMAGE_FSTYPES_append', ' %s.bz2 ' % type)
+
+  # Remove the boot option on the Live installer; it won't work since Mender hard codes
+  # the device nodes
+  d.setVar('LABELS_LIVE_remove', 'boot')
+}
```
kacf: Inconsistent indentation, both using 2 and 3 spaces. Should stick to 4, like in other Python sections.

meta-mender-core/recipes-core/initrdscripts/files/init-install-efi-mender.sh
```diff
@@ -0,0 +1,133 @@
...
+
+if [ -e /run/media/$1/$2 ]; then
+    echo "Writing Mender-enabled Yocto image /run/media/$1/$2 to ${device}"
+    echo "bzcat /run/media/$1/$2 | /bin/dd of=$device bs=8M"
+    bzcat /run/media/$1/$2 | /bin/dd of=$device bs=8M
+    rc=$?
+    if [ $rc = 0 ]; then
+        echo "Installation successful."
+    else
+        echo "/bin/dd returned error $rc. Review above log. Installation aborted"
```
kacf: You are running with `-e` shell argument. This block will never be run, because it will terminate before you have a chance to get here. A better construct for this scenario is:

```
rc=0
bzcat /run/media/$1/$2 | /bin/dd of=$device bs=8M || rc=$?
if [ $rc = 0 ]; then
    ...
```

The `||` operator causes the error termination to be suppressed, and you can get the error code instead.

meta-mender-core/recipes-core/initrdscripts/files/init-install-efi-mender.sh
```diff
@@ -0,0 +1,133 @@
...
+    echo "No hard drive selected. Installation aborted."
+    exit 1
+fi
+
+device=/dev/$TARGET_DEVICE_NAME
+
+if [ -e /run/media/$1/$2 ]; then
+    echo "Writing Mender-enabled Yocto image /run/media/$1/$2 to ${device}"
+    echo "bzcat /run/media/$1/$2 | /bin/dd of=$device bs=8M"
+    bzcat /run/media/$1/$2 | /bin/dd of=$device bs=8M
```
kacf: I would put a `sync` after this, since `reboot -f` is used later on, which doesn't wait for disk syncing. Since there is a pause in between, it's pretty unlikely to be a problem, but doesn't hurt to be safe.

meta-mender-core/classes/mender-part-images.bbclass
```diff
@@ -261,6 +261,28 @@ IMAGE_TYPEDEP_biosimg_append = "${@bb.utils.contains('IMAGE_FSTYPES', 'sdimg', '
 IMAGE_TYPEDEP_gptimg_append = "${@bb.utils.contains('IMAGE_FSTYPES', 'sdimg', ' sdimg', '', d)} \
                                ${@bb.utils.contains('IMAGE_FSTYPES', 'uefiimg', ' uefiimg', '', d)} \
                                ${@bb.utils.contains('IMAGE_FSTYPES', 'biosimg', ' biosimg', '', d)}"
+# Make sure the Mender part image is available in the live installer
+IMAGE_TYPEDEP_hddimg_append = "${@bb.utils.contains('IMAGE_FSTYPES', 'sdimg', ' sdimg', '', d)} \
+                               ${@bb.utils.contains('IMAGE_FSTYPES', 'uefiimg', ' uefiimg', '', d)} \
+                               ${@bb.utils.contains('IMAGE_FSTYPES', 'biosimg', ' biosimg', '', d)}"
```
kacf: Is `gptimg` excluded for any particular reason?

drewmoseley: Just an oversight.  Fixed.

meta-mender-core/recipes-core/initrdscripts/files/init-install-efi-mender.sh
drewmoseley: Added

meta-mender-core/recipes-core/initrdscripts/files/init-install-efi-mender.sh
drewmoseley: Done.

meta-mender-core/classes/mender-part-images.bbclass
drewmoseley: Fixed.

meta-mender-core/recipes-core/initrdscripts/files/init-install-efi-mender.sh
drewmoseley: I just copied the original one. But it has way too much logic in it so I have a new version coming that just starts from scratch and uses the MENDER_STORAGE_DEVICE variable instead of all the scanning that is done in the original.

drewmoseley:
@kacf I've pushed updated commits here.  As far as I know, this is ready to go.

meta-mender-core/recipes-core/initrdscripts/files/init-install-efi-mender.sh
kacf: :+1:

meta-mender-core/classes/mender-part-images.bbclass
```diff
@@ -261,6 +261,29 @@ IMAGE_TYPEDEP_biosimg_append = "${@bb.utils.contains('IMAGE_FSTYPES', 'sdimg', '
...
+
+# Use the Mender part image as the Live image
+python() {
+    if bb.utils.contains('IMAGE_FSTYPES', 'sdimg', True, False, d):
+        type='sdimg'
+    elif bb.utils.contains('IMAGE_FSTYPES', 'uefiimg', True, False, d):
+        type='uefiimg'
+    elif bb.utils.contains('IMAGE_FSTYPES', 'biosimg', True, False, d):
+        type='biosimg'
```
kacf: Jenkins revealed that you will need to add `gptimg` here too. Maybe it's a good idea to put this at the end though, in case the user isn't building an image, but only an artifact:

```
else:
    return
```

otherwise he may get a `UnboundLocalError` for the `type` variable, as in the currently failing build. I would perhaps try a local build and remove all the `mender-image-*` using `MENDER_FEATURES_DISABLE_append`, just to be sure.

drewmoseley: Confirmed and fixed.

kacf:
Thanks!

I think the [current failures](https://s3-eu-west-1.amazonaws.com/mender-testing-reports/integration-reports-qemux86-64-uefi-grub/4047/report.html) are spurious, but I'll rerun once to be sure.

PR #780, opened 2019-07-24
Thud backports

drewmoseley:
These are backports of a few things that have made it into master. I need these for customer projects.


drewmoseley:
Wrong base branch.

PR #781, opened 2019-07-24
Thud backports

drewmoseley:
These are backports of a few things that have made it into master. I need these for customer projects.


kacf:
I think the failures are unrelated. For now I'll just cherry-pick these to #776, where I will continue to work to get it in. Unfortunately due our severe CI instability, it may be some time before I can merge it.

PR #783, opened 2019-07-29
rpi: Fix several assignment bugs regarding `MENDER_BOOTLOADER_DEFAULT`.

kacf:
* `rpi` is not a valid machine name, and there are anyway too many
  machine names in that layer to specify all of them. Instead set
  unconditionally for the meta-mender-raspberrypi layer.

* We need weak assignment in the meta-mender-core layer, since the
  `mender-full.bbclass` is evaluated after the `layer.conf` file, even
  if the former layer is listed first.

Changelog: title

### DIFF
--- a/meta-mender-core/classes/mender-full.bbclass
+++ b/meta-mender-core/classes/mender-full.bbclass
@@ -8,8 +8,8 @@ MENDER_FEATURES_ENABLE_append = " \
     mender-systemd \
 "
 
-_MENDER_IMAGE_TYPE_DEFAULT = "mender-image-uefi"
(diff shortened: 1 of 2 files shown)


mirzak:
FWIW , `rpi` should be a valid machine override which is defined here,

https://github.com/agherzan/meta-raspberrypi/blob/master/conf/machine/include/rpi-base.inc#L5-L6

And it should be a common override for all raspberry pi machines.

kacf:
I'm pretty sure I tested this, but maybe it got mixed up with the weak assignment. Let me retry it.

kacf:
#788.

PR #784, opened 2019-07-30
Fix vexpress qemu not starting

lluiscampos:
```
commit 41091d8dfde32a0bfad205d9d7f769a610c46123 (HEAD -> fix-vexpress-qemu-not-starting, lluiscampos/fix-vexpress-qemu-not-starting)
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Tue Jul 30 12:42:39 2019 +0200

    Switch to upstream qemu for ARM architecture
    
    Building it from source does not work for Alpine Linux 3.7. The
    resulting Docker image is unable to start-up qemu, as many shared
    libraries ar not found.
    
    It is unknown if the reset race bug is fixed upstream for this Linux
    version.
    
    Changelog: None
    
commit 93e733b67a312f1e05d7266891c70ec72efa926a
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Tue Jul 30 12:40:54 2019 +0200

    Exit with failure when qemu-system-xxx cannot start
    
    This was the case when building it from source on Alpine Linux 3.7, but
    the error went unnoticed
    
    Changelog: None
    
```

### DIFF
--- a/meta-mender-qemu/docker/vexpress-qemu/Dockerfile
+++ b/meta-mender-qemu/docker/vexpress-qemu/Dockerfile
@@ -21,90 +21,11 @@ FROM alpine:3.7
 # Install packages
 RUN apk update && apk upgrade && \
     apk add util-linux multipath-tools \
-            bash e2fsprogs-extra python3 && \
-    rm -rf /var/cache/apk/*
+            bash e2fsprogs-extra python3
 
-# Install qemu from source
-RUN apk update && apk add --virtual build-dependencies \
-    alsa-lib-dev \
-    bison \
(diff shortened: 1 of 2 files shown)


lluiscampos:
@kacf I am unsure about the resolution as I cannot trace the original bug to the upstream version in which it was fixed.

But for now I am launching the PR to see how the testing goes.

lluiscampos:
The vexpress integration tests have passed but, ironically, the x86-64 integration tests have aborted with a java Jenkins exception...

Merging, as such failure is not related with this change set.

PR #788, opened 2019-08-05
Restore `rpi` based variable selection for Raspberry Pi.

kacf:
Apparently it is defined through `SOC_FAMILY` in the
github.com/agherzan/meta-raspberrypi layer.

Changelog: None

### DIFF
--- a/meta-mender-raspberrypi/conf/layer.conf
+++ b/meta-mender-raspberrypi/conf/layer.conf
@@ -17,5 +17,5 @@ LAYERDEPENDS_mender-raspberrypi = "mender raspberrypi"
 LAYERSERIES_COMPAT_mender-raspberrypi = "thud"
 
 # Raspberry Pi doesn't work with GRUB currently.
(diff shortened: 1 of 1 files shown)


mirzak:
Thanks for testing this, I can see that it fails.

Then I suspect it has something to do with parse order and `_rpi` is not provided when `layer.conf` is parsed because the `_rpi`is  frequently used in meta-raspberrypi recipes and is for sure a valid override.

mirzak:
I wonder if the "correct" way of doing this would be,

```
# Raspberry Pi doesn't work with GRUB currently.
_MENDER_IMAGE_TYPE_DEFAULT_mender-raspberrypi = "mender-image-sd"
_MENDER_BOOTLOADER_DEFAULT_mender-raspberrypi = "mender-uboot"
```

No idea if it works.

kacf:
Yes, there are some special rules for `layer.conf` I think. Closing this then.

PR #791, opened 2019-08-07
Persistent machine id

kacf:
Just testing #759.


kacf:
Original #759 merged.

PR #792, opened 2019-08-07
DEBUG: Extend the valid time frame for test_uboot_compile to 14 days

lluiscampos:
For debug purposes.

Changelog: None


lluiscampos:
The experiments have shown that 1) we can use /dev/shm in GitLab if we just mount the volume for the containers and 2) using disk instead of ramfs made the test_uboot_compile 1h slower.

PR #794, opened 2019-08-08
Live installer device scan

drewmoseley:
Also added a warning to ensure users know that MENDER_STORAGE_DEVICE is not relevant when using mender-partuuid.

### DIFF
--- a/meta-mender-core/classes/mender-setup.bbclass
+++ b/meta-mender-core/classes/mender-setup.bbclass
@@ -316,6 +316,9 @@ python() {
         bb.fatal("IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET is deprecated. Please define MENDER_IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET instead.")
     if d.getVar('MENDER_DATA_PART_DIR'):
         bb.fatal("MENDER_DATA_PART_DIR is deprecated. Please use recipes to add files directly to /data instead.")
+
+    if bb.utils.contains('DISTRO_FEATURES', 'mender-partuuid', True, False, d) and d.getVar('MENDER_STORAGE_DEVICE', True) != "":
+        bb.warn("MENDER_STORAGE_DEVICE is ignored when mender-partuuid is enabled. Clear MENDER_STORAGE_DEVICE to remove this warning.")
 }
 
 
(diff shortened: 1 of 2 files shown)


meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -316,6 +316,9 @@ python() {
         bb.fatal("IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET is deprecated. Please define MENDER_IMAGE_BOOTLOADER_BOOTSECTOR_OFFSET instead.")
     if d.getVar('MENDER_DATA_PART_DIR'):
         bb.fatal("MENDER_DATA_PART_DIR is deprecated. Please use recipes to add files directly to /data instead.")
+
+    if bb.utils.contains('DISTRO_FEATURES', 'mender-partuuid', True, False, d) and d.getVar('MENDER_STORAGE_DEVICE', True) != "":
+        bb.warn("MENDER_STORAGE_DEVICE is ignored when mender-partuuid is enabled. Clear MENDER_STORAGE_DEVICE to remove this warning.")
```
kacf: Are you sure this is ok? The `grub-mender-grubenv` recipe has references to `MENDER_STORAGE_DEVICE_BASE`, which is deduced from `MENDER_STORAGE_DEVICE`.

drewmoseley: @kacf I'm pretty sure this is correct.  That recipe will not use MENDER_STORAGE_DEVICE_BASE in the case that mender-partuuid is enabled; It will use MENDER_GRUB_STORAGE_DEVICE instead. And if _that_ is not set it will error out.

drewmoseley: No. Maybe I'm wrong.  Looking some more.

drewmoseley: OK.  I am now ok with this.  In the grub-mender-grubenv.inc file, the variable _mender_kernel_root_base_ is set to ${MENDER_STORAGE_DEVICE_BASE} which in this case will be blank.  In _90_mender_boot_grub.cfg_, the _mender_kernel_root_base_ variable is bypassed when UUIDs are set.

kacf: Ok, thanks for checking.

PR #800, opened 2019-08-15
add psplash in meta-mender-demo

texierp:
This PR add `psplash` support with the Mender logo (Mender.io (color) -> https://mender.io/resources/digital-assets).
This is way to demonstrate that it is actually the Mender demo image running.

But also to use psplash for demonstrations
with Mender updates modules and state scripts:

```
$: psplash &
$: psplash-write "MSG Install in progress ..."
$: psplash-write "PROGRESS 50"
$: psplash-write "MSG Done"
$: psplash-write "PROGRESS 100"
$: psplash-write "QUIT"
```


drewmoseley:
Does this work with systemd? I thought psplash didn't work with it.

texierp:
> Does this work with systemd? I thought psplash didn't work with it.

You're right @drewmoseley, in fact the main idea at first, is not necessarily to use psplash during the boot phase, but rather manually with update modules and state scripts. For example:

```
$: psplash &
$: psplash-write "MSG Install in progress ..."
$: psplash-write "PROGRESS 50"
$: psplash-write "MSG Done"
$: psplash-write "PROGRESS 100"
$: psplash-write "QUIT"
```

drewmoseley:
Interesting.  I wonder if this is appropriate to add as a MENDER_FEATURE in meta-mender-core?  @kacf?

@texierp  have you tried dietsplash? It's supposed to work with systemd but I could only get the initial splash screen displayed when I tried it. It didn't seem to update with the cli utility, although that utility _does_ work to stop the splash screen.

texierp:
> @texierp have you tried dietsplash? It's supposed to work with systemd but I could only get the initial splash screen displayed when I tried it. It didn't seem to update with the cli utility, although that utility _does_ work to stop the splash screen.

Never tested :/. I will try.
However, it is possible to run psplash with systemd during the boot phase with a custom script as done here in `meta-mentor` and `meta-agl` : 
- https://github.com/MentorEmbedded/meta-mentor/blob/master/meta-mel/recipes-core/psplash/psplash_git.bbappend

- https://git.automotivelinux.org/AGL/meta-agl/tree/meta-agl-profile-core/recipes-core/psplash/psplash_git.bbappend

texierp:
Ping :)

mirzak:
@kacf , is out of office so I would be patient :).

I like the `psplash-start.service` approach, this way it will display by default without the need of running `psplash` manually.

> Interesting. I wonder if this is appropriate to add as a MENDER_FEATURE in meta-mender-core?

texierp:
> @kacf , is out of office, patience :)

Thanks @mirzak, I'll be patient :)

> I like the `psplash-start.service` approach, this way it will display by default without the need of running `psplash` manually.

Sure, this PR is now updated with both scripts :)

> FWIW, I think this is to intrusive to be in core. Can not really imagine a use-case where someone would want that if they are not building demo images.

Agree with that.

mirzak:
One remark on the commit messages, and specifically these three:

    psplash: add simple systemd scripts
    layer.conf: add splash in IMAGE_FEATURES
    psplash: use Mender logo by default

You have a Changelog entry for all three of them, which is unnecessary. You can have `None` on two of them and on the third set something like:

    Changelog: Add support for psplash in meta-mender-demo, which will display the Mender logo during boot

Changelog entries are supposed to be user-facing/friendly and not necessarily high technical detail.

texierp:
Ok, thanks @mirzak for the review, I will update ;)

meta-mender-demo/recipes-core/psplash/psplash_git.bbappend
```diff
@@ -2,6 +2,17 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
...
+	file://psplash-start.service \
+	file://psplash-quit.service \
 "
 
+inherit systemd
+
+SYSTEMD_SERVICE_${PN} = "psplash-start.service psplash-quit.service"
+
 SPLASH_IMAGES = "file://mender.io.png;outsuffix=default"
+
```
mirzak: Did a quick test on my Pi4, I had to add the following to "win" over meta-raspberrypi who set priority to 200.

```
ALTERNATIVE_PRIORITY = "300"
```

meta-mender-demo/recipes-core/systemd/systemd_%.bbappend
```diff
@@ -20,4 +20,7 @@ do_install_append() {
         install -m 0755 ${WORKDIR}/eth.network ${D}${sysconfdir}/systemd/network
         install -m 0755 ${WORKDIR}/enp.network ${D}${sysconfdir}/systemd/network
   fi
+
+  # Remove getty@tty1.service
+  rm ${D}${sysconfdir}/systemd/system/getty.target.wants/getty@tty1.service
```
mirzak: This does not seem to work, at least not on RPi4. I do get the console on the HDMI output and I have to manually start psplash for it to show.

meta-mender-demo/recipes-core/psplash/psplash_git.bbappend
texierp: Ah indeed, thanks for testing. I will update the recipe.

meta-mender-demo/recipes-core/systemd/systemd_%.bbappend
texierp: It seems we have to set `SPLASH = "psplash"` ? (https://github.com/agherzan/meta-raspberrypi/blob/warrior/recipes-core/images/rpi-basic-image.bb#L9)

meta-mender-demo/conf/layer.conf
```diff
@@ -16,5 +16,8 @@ MENDER_STORAGE_TOTAL_SIZE_MB_DEFAULT ?= "608"
 IMAGE_OVERHEAD_FACTOR = "1.0"
 IMAGE_FEATURES += "splash"
 
+# Special for RPi
+ENABLE_UART = "1"
```
kacf: This is already present in [meta-mender-raspberrypi-demo](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-raspberrypi-demo/classes/rpi-demo-wifi.bbclass#L6), no need to add it here.

meta-mender-demo/recipes-core/systemd/systemd_%.bbappend
kacf: Yes, and is it necessary? I thought the idea was to run psplash until the login prompt is ready and then switch to the prompt? Certainly I don't think we want to leave the login disabled, since the user may prefer to use the HDMI output to log in.

meta-mender-demo/conf/layer.conf
texierp: Indeed :), I will remove

meta-mender-demo/recipes-core/systemd/systemd_%.bbappend
texierp: > I don't think we want to leave the login disabled, since the user may prefer to use the HDMI output to log in.

You're indeed right, I will update accordingly.

meta-mender-demo/conf/layer.conf
texierp: > Nice piece of work @texierp! I have a few change requests, but nothing major. We'll have to verify that RaspberryPi indeed works though.

With pleasure @kacf ! PR updated :)

mirzak:
Did a test run on my Raspberry Pi 3, looks good. Once concern though about including:

```
psplash: add simple systemd scripts
```
(I liked the idea initially but after testing I am not convinced),

The problem with above is that it relies on that you do not have `console=tty1` (or no `console=` definition in which case it will default to the first mentioned) in kernel arguments, for it to be "pretty".

If you have `console=tty1` the output on the screen is not pretty, and typically what happens is:

- Kernel boot log is displayed
- psplash is started and Mender logo is displayed (for a couple of milliseconds)
- the kernel boot partly overwrites the Mender boot logo in the continued boot process and finally removing it completely once it reaches the login prompt.

I believe it is a fairly common default behavior to have console output on HDMI, and for this reason we should opt-out of this, as we will not be modifying kernel arguments on all boards we support to make this pretty.

But the procedure you initially proposed: 

```
$: psplash &
$: psplash-write "MSG Install in progress ..."
$: psplash-write "PROGRESS 50"
$: psplash-write "MSG Done"
$: psplash-write "PROGRESS 100"
$: psplash-write "QUIT"
``` 

works great, and I think this is the best we can do in a layer such as `meta-mender-demo`.

A nice follow up would be to update the [example state scripts](https://github.com/mendersoftware/meta-mender/tree/master/meta-mender-demo/recipes-mender/example-state-scripts) to write to `psplash` as well.

texierp:
You are indeed right @mirzak , and in fact, I am also convinced by my initial idea, which is less intrusive.

After your feedback, I think the second approach (with systemd) is platform dependent. So, I will drop this patch.

Thanks !

texierp:
> A nice follow up would be to update the [example state scripts](https://github.com/mendersoftware/meta-mender/tree/master/meta-mender-demo/recipes-mender/example-state-scripts) to write to `psplash` as well.

I could do that in another PR.

mirzak:
> I could do that in another PR.

Yeah, did not mean to include it here.

kacf:
So with the newly proposed method, does this mean that the splash screen will only be active during an update?

mirzak:
> So with the newly proposed method, does this mean that the splash screen will only be active during an update?

With only merging this PR, it will not be active at all by default, unless you run `psplash` command.

If we follow up with adding `psplash` commands to the state script example then it would be active during upgrade.

kacf:
Alright then, it will be dormant, but sounds reasonable to provide in the demo image.

Will run this through auto-testing.

texierp:
> Alright then, it will be dormant, but sounds reasonable to provide in the demo image.
> 
> Will run this through auto-testing.

Thanks @kacf !

PR #804, opened 2019-08-23
u-boot-fw-utils: set PACKAGE_ARCH as MACHINE_ARCH

texierp:
Add PACKAGE_ARCH to force package to be machine-dependent

Changelog: Title

Signed-off-by: Pierre-Jean Texier <pjtexier@koncepto.io>

### DIFF
--- a/meta-mender-core/recipes-bsp/u-boot/u-boot-fw-utils-mender-auto-provided_1.0.bb
+++ b/meta-mender-core/recipes-bsp/u-boot/u-boot-fw-utils-mender-auto-provided_1.0.bb
@@ -74,3 +74,5 @@ do_install () {
     install -m 755 ${B}/tools/env/fw_printenv ${D}${base_sbindir}/fw_setenv
     install -m 0644 ${B}/tools/env/fw_env.config ${D}${sysconfdir}/fw_env.config
 }
+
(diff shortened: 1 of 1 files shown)


kacf:
This package is not machine dependent though, only configuration dependent. It only adds regular binaries and a configuration text file. Is there a specific problem this is fixing?

mirzak:
I would agree that it is machine specific, since the `fw_printenv/fw_setenv` binaries will depend on the `defconfig` used by a specific machine which impacts built in default environment etc.

I guess what problem this solves (correct me if I am wrong @texierp  :)), is if you are generating package feeds, you would want this package only to be available to the machine it was built for and not end up in one of the generic feeds e.g `armhf`.

texierp:
> I guess what problem this solves (correct me if I am wrong @texierp :)), is if you are generating package feeds, you would want this package only to be available to the machine it was built for and not end up in one of the generic feeds e.g `armhf`.

Exactly, this is the first reason :)

The second one is that it seems to me interesting to follow the [openembedded-core](http://cgit.openembedded.org/openembedded-core/tree/meta/recipes-bsp/u-boot/u-boot-fw-utils_2019.07.bb#n33).

kacf:
Thanks!

PR #806, opened 2019-08-28
example-state-scripts: use show-artifact instead of parsing file

texierp:
The /etc/mender/artifact_info file is deprecated from Mender 2.0,
as such info is now saved in the database.

Changelog: Title

Signed-off-by: Pierre-Jean Texier <pjtexier@koncepto.io>

### DIFF
--- a/meta-mender-demo/recipes-mender/example-state-scripts/files/example-script
+++ b/meta-mender-demo/recipes-mender/example-state-scripts/files/example-script
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-echo "$(cat /etc/mender/artifact_info): $(basename "$0") was called!" >&2
+echo "$(mender -show-artifact): $(basename "$0") was called!" >&2
 exit 0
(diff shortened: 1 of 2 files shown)


meta-mender-demo/recipes-mender/example-state-scripts/files/example-script
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-echo "$(cat /etc/mender/artifact_info): $(basename "$0") was called!" >&2
+echo "$(mender -show-artifact): $(basename "$0") was called!" >&2
```
kacf: Thanks, good catch!

However, there is a subtle difference in behavior in `-show-artifact`, namely that it will not be updated until the artifact is committed. The old `artifact_info` file would be updated as soon as the device was rebooted. Some of the point of the example script is to show this difference, therefore I suggest that you add one more line to the recipe file, so that we also see the new name after it has been committed:

```
cp example-script ${MENDER_STATE_SCRIPTS_DIR}/ArtifactCommit_Leave_50
```

texierp: Ah indeed, thanks for this feedback, I will do so.

texierp: Updated ;)

texierp:
@mirzak I just realized that you closed the PR. 
But to be honest, I'm not sure I understand why :/

mirzak:
Apologize @texierp, this was just me clicking the wrong button at some point. Thank you for bringing it to my intention

texierp:
> Apologizes @texierp, this was just me clicking the wrong button at some point. Thank you for bringing it to my intention

;)

kacf:
Merged in #811 with additional fixes.

PR #807, opened 2019-08-29
Add README for meta-mender-commercial layer.

kacf:
Changelog: None


kacf:
This is a stub for now, I have updated https://tracker.mender.io/browse/MEN-2735 to include an update this README.

PR #809, opened 2019-08-29
fix MENDER_DATA_PART default value when creating image without bootpart

mirzak:
The index is "3" when no bootpart is present, because we will get rid of
the "extended" partition if we remove boot part.

Changelog: Fix boot when MENDER_DATA_PART_DEFAULT = "0"

### DIFF
--- a/meta-mender-core/classes/mender-setup.bbclass
+++ b/meta-mender-core/classes/mender-setup.bbclass
@@ -34,7 +34,7 @@ MENDER_ROOTFS_PART_B_NAME_DEFAULT = "${MENDER_ROOTFS_PART_B}"
 
 # The partition number holding the data partition.
 MENDER_DATA_PART ??= "${MENDER_DATA_PART_DEFAULT}"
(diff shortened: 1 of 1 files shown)


kacf:
This is one of the really old branches, so Gitlab builds can't be trusted. I'll let you guys handle this one.

mirzak:
> Do we have any other platforms that do not use the ext partition?

Not sure I follow this question, can you clarify?

> What about other branches?

This seems ok to me on other branches. I do not really know why but on morty we seem to have `data` on a extended partition when boot part is included, this is essentially what is causing the problem. This is not the case on other branches

drewmoseley:
I seem to recall that in pyro the default partitioning was changed so it's not entirely unexpected. I just wonder if there might be other machines/image types in morty that will be affected by this.  But I suppose we can just handle those if they happen.

From my POV this is good for merge.

oleorhagen:
What's going on with this one? @mirzak

mirzak:
I think I just forgot to click the button

PR #812, opened 2019-09-04
Support passing global options to mender-artifact

changyuheng:
Starting from mender-artifact v3, the command supports compressing the
image with LZMA. However, there's no way to specify it from the Yocto
building system. With this commit, one's able to do so by inserting the
following variable into `local.conf`.

```
MENDER_ARTIFACT_EXTRA_OPTS_append = " --compression lzma"
```

### DIFF
--- a/meta-mender-core/classes/mender-artifactimg.bbclass
+++ b/meta-mender-core/classes/mender-artifactimg.bbclass
@@ -2,6 +2,9 @@ inherit mender-helpers
 
 # ------------------------------ CONFIGURATION ---------------------------------
 
+# Extra global options that should be passed to mender-artifact.
+MENDER_ARTIFACT_EXTRA_OPTS ?= ""
+
 # Extra arguments that should be passed to mender-artifact.
 MENDER_ARTIFACT_EXTRA_ARGS ?= ""
 
@@ -55,7 +58,8 @@ IMAGE_CMD_mender () {
         image_flag=-f
     fi
 
-    mender-artifact write rootfs-image \
+    mender-artifact ${MENDER_ARTIFACT_EXTRA_OPTS} \
+        write rootfs-image \
         -n ${MENDER_ARTIFACT_NAME} \
         $extra_args \
         $image_flag ${IMGDEPLOYDIR}/${ARTIFACTIMG_NAME}.${ARTIFACTIMG_FSTYPE} \
--- a/meta-mender-core/conf/mender-vars.json
+++ b/meta-mender-core/conf/mender-vars.json
@@ -19,6 +19,7 @@
      "'MENDER_FEATURES_ENABLE': ['mender-uboot', 'mender-install']",
      "(using single quote (') for double quote in the above string)"],
     "MENDER_ARTIFACT_EXTRA_ARGS": "",
+    "MENDER_ARTIFACT_EXTRA_OPTS": "",
     "MENDER_ARTIFACT_NAME": "",
     "MENDER_ARTIFACT_SIGNING_KEY": "",
     "MENDER_ARTIFACT_VERIFY_KEY": "",


kacf:
While your change is perfectly fine, I would ideally like us to solve this in a different way: What we should do is change mender-artifact to accept the option after the `rootfs-image` argument as well. This would have several advantages beyond what this PR provides:

* We would get the fix for free in meta-mender, because `MENDER_ARTIFACT_EXTRA_ARGS` already exists and can be used to pass such parameters. This is also less confusing than having two different, but similar variables to pass arguments

* It would also solve the problem that it is impossible to add compression options at the end of a mender-artifact command line, which is needed heavily by Update Module generators that typically can pass raw mender-artifact options by specifying `--` (see [this generator](https://github.com/mendersoftware/mender/blob/master/support/modules-artifact-gen/single-file-artifact-gen#L21) for example).

So in short, that would be a more global fix for the problem. For this reason I don't think we should accept this PR, since we would have to maintain it even after fixing mender-artifact.

Do you agree with my reasoning, @mirzak?

changyuheng:
Hi @kacf, I agree with the path you're going. However, if we want to make the interface of `mender-artifact` right, I think just nest `--compression` to `rootfs-image` isn't perfect.

Below is the interface of `mender-artifact` version 3.0.0:

`mender-artifact -h`
```
NAME:
   mender-artifact - interface for manipulating Mender artifacts

USAGE:
   mender-artifact [--version][--help] <command> [<args>]

VERSION:
   3.0.0

AUTHOR:
   Northern.tech AS <contact@northern.tech>

COMMANDS:
     write     Writes artifact file.
     read      Reads artifact file.
     validate  Validates artifact file.
     sign      Signs existing artifact file.
     modify    Modifies image or artifact file.
     cp        cp <src> <dst>
     cat       cat [artifact|sdimg|uefiimg]:<filepath>
     install   install -m <permissions> <hostfile> [artifact|sdimg|uefiimg]:<filepath>
     rm        rm [artifact|sdimg|uefiimg]:<filepath>
     help, h   Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --compression value  Compression to use for data and header inside the artifact, currently supports: none, gzip, lzma. (default: "gzip")
   --help, -h           show help
   --version, -v        print the version
```

As you can see, users should only specify the compression method when they generate artifacts, which is the `write` sub-command, other sub-commands that read or modify existing artifacts, the compression method should be automatically obtained from the existing artifacts as we shouldn't need to have multiple compression methods within a single artifact.

--

In addition, I also have a suggestion to the current sub-command `write`. `write` here doesn't make too much sense to me as `cp` and `install` are also kinds of writing commands. And look into `write`'s manual, it actually packs/build an artifact with a block image.

`mender-artifact write -h`
```
NAME:
   mender-artifact write - Writes artifact file.

USAGE:
   mender-artifact write command [command options] [arguments...]

COMMANDS:
     rootfs-image  Writes Mender artifact containing rootfs image
     module-image  Writes Mender artifact for an update module

OPTIONS:
   --help, -h  show help
```

Look further into `write` with `rootfs-image`, its usage is:

`mender-artifact write rootfs-image -h`
```
NAME:
   mender-artifact write rootfs-image - Writes Mender artifact containing rootfs image

USAGE:
   mender-artifact write rootfs-image [command options] <image path>

OPTIONS:
   --file FILE, -f FILE             Payload FILE.
   --device-type value, -t value    Type of device(s) supported by the Artifact. You can specify multiple compatible devices providing this parameter multiple times.
   --artifact-name value, -n value  Name of the artifact
   --output-path value, -o value    Full path to output artifact file.
   --version value, -v value        Version of the artifact. (default: 3)
   --key value, -k value            Full path to the private key that will be used to verify the artifact signature.
   --script value, -s value         Full path to the state script(s). You can specify multiple scripts providing this parameter multiple times.
```

My suggestions are:

Option 1: Rename `write` sub-command to `pack`, `build`, `compose` or `compile`, and also rename `<image path>` to `<source image path>`. - Or so. Just do some renaming because the current names are confusing.
Option 2: Lift `rootfs-image` and `module-image` with one level up, and rename them to `pack-rootfs-image` and `pack-module-image`, respectively.

mirzak:
> So in short, that would be a more global fix for the problem. For this reason I don't think we should accept this PR, since we would have to maintain it even after fixing mender-artifact.

> Do you agree with my reasoning, @mirzak?

Makes sense to me.

kacf:
@changyuheng: We are restricted to adding arguments only, to maintain backward compatibility, so we cannot rename sub commands. Otherwise I agree with your option 2 example, that the `--compression` argument needs to come after the `rootfs-image` sub command.

changyuheng:
Makes sense. Please feel free to close the PR if it's not relevant.

kacf:
Alright, thanks for the effort anyway!

PR #813, opened 2019-09-06
Several partition type fixes

kacf:
```
commit b66beb2945ebb4ceb1465914c53118a48921a6b2
Date:   Fri Sep 6 07:36:14 2019

    Fix incorrect partition types after we started using `rawcopy`.
    
    This was introduced in 30a27a9f5e9ba8edb1e879380a64fc8cc994af43. When
    we switched away from the `rootfs` source, we lost the identity of the
    filesystem, and all partitions became the generic "Microsoft basic
    data" type. We fix GPT by using `--part-type` argument, but the MBR
    code in wic doesn't support using it, so we need to fix those entries
    manually with fdisk.
    
    Changelog: None
    
commit 4eb39076050b08c3a9ce5a335914fb1bd1f45459
Date:   Fri Sep 6 08:38:31 2019

    Fix incorrect boot partition type for EFI boot partitions.
    
    This was broken in 0b737428d0570de16935e0758db7752de4134d2a, where we
    lost the `boot_part_params` along the way.
    
    Changelog: Title
    
```

### DIFF
--- a/meta-mender-core/classes/mender-part-images.bbclass
+++ b/meta-mender-core/classes/mender-part-images.bbclass
@@ -44,7 +44,7 @@ IMAGE_NAME_SUFFIX = ""
 
 mender_part_image() {
     suffix="$1"
-    part_type="$2"
+    ptable_type="$2"
     boot_part_params="$3"
 
     set -ex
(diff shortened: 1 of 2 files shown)


kacf:
Actually, this should really have an acceptance test too. But the code can be reviewed anyway.

mirzak:
FWIW I took at this and looks good.

Though I do not fully understand why we would set the same partition type on all partitions in the GPT case, but I am sure there is a reason. Is it because in the GPT case, you are actually using "Partition type GUIDs" to differentiate between the partitions?

kacf:
> Though I do not fully understand why we would set the same partition type on all partitions in the GPT case, but I am sure there is a reason. Is it because in the GPT case, you are actually using "Partition type GUIDs" to differentiate between the partitions?

Not completely sure I understand the question. The Partition Type GUIDs are just a different way to express the types, it is still essentially the same as in MBR partition tables. `wic` will auto set the boot partition to Microsoft Basic Data because of `--fstype vfat`, and the rest should be Linux Filesystem all of them, don't you think?

kacf:
Added an acceptance test too.

kacf:
Still [some error](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/292940299). Might be spurious, not sure. [Rerunning](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/81994775).

PR #816, opened 2019-09-13
sumo: Add meta-mender-commercial layer

lluiscampos:


lluiscampos:
Starting build only for 2.1.0b1 release, as we are only interested in the commercial test.

ref: https://mender-jenkins.mender.io/job/mender-builder/4331/

kacf:
I guess it's missing [this one](https://github.com/mendersoftware/meta-mender/commit/6252cb5f73f71adb4ff501c6faa9b39c19a4bc46).

lluiscampos:
@kacf Yes, I know. But that will still be a workaround. I don't understand why is it building mender master in the first place.

kacf:
I'm not 100% sure, but I *think* it might be because tags are always listed in reverse chronological order, so it always grabs the build1 tag. Normally we never have a build tag on the same commit as another tag (we would just release the original instead), *except* for betas. So I assume we simply have never been in the situation before where a beta was promoted to a final without any fixes in between.

Now, how to solve that....

lluiscampos:
Latest build: https://mender-jenkins.mender.io/job/mender-builder/4345/

lluiscampos:
All tests successfully run and passed in the previous build: https://mender-jenkins.mender.io/job/mender-builder/4349/

The last push of changes are to port https://github.com/mendersoftware/meta-mender/pull/819 and https://github.com/mendersoftware/meta-mender/pull/823 which don't require repeating the tests.

PR #819, opened 2019-09-16
Enabled SRC_URI folder access to arm64 binaries.

kacf:
Changelog: None


meta-mender-commercial/recipes-mender/mender-binary-delta/mender-binary-delta_0.1.1.bb
```diff
@@ -2,6 +2,7 @@ LICENSE = "CLOSED"
 LICENSE_FLAGS = "commercial"
 
 SUB_FOLDER_arm = "armhf"
+SUB_FOLDER_aarch64 = "arm64"
```
lluiscampos: We should also add it in COMPATIBLE_HOSTS

kacf: Indeed we should! Fixed.

kacf:
> We could have called the folder aarch64, if you guys think that the term is more extended than arm64

I checked this specifically, and arm64 appears to be the correct terminology as opposed to armhf. There seems to be two competing terminologies, and arm64 is the Debian term, aarch64 the Yocto term. I have no strong feelings about which one is "correct", but at least we are consistent! :-)

lluiscampos:
I think we should then use aarch64, as our deliverable is for Yocto. The fact that is built in a Debian container should be transparent to our users.

Actually, even inside Debian the terminology is not always consistent, the toolchain uses `arm64` while pkgconfig uses `aarch64`, see [here](https://github.com/mendersoftware/mender-binary-delta/blob/master/.gitlab-ci.yml#L58).

lluiscampos:
I can do it in separate PRs if you wish. Let me know.

kacf:
Yes, please do that. It requires changes in both repos, so I'd rather that happen separately.

No clue why the integration tests failed miserably. Will retry.

PR #815, opened 2019-09-12
MEN-2337: Enable dynamic resizing of the data partition

oleorhagen:
Changelog: Title. This is done through enabling systemd's growfs
feature. Hereforth the feature is enabled by default when inheriting
from 'mender-full'. The feature can be disabled, and must be done so
explicitly through the 'MENDER_DATA_PART_FSTAB_OPTS' variable.

Signed-off-by: Ole Petter <ole.orhagen@northern.tech>

PR #822, opened 2019-09-17
Men 2337 enable dynamic resizing of the data partition

mirzak:
Supersedes #815

### DIFF
--- a/meta-mender-core/classes/mender-setup.bbclass
+++ b/meta-mender-core/classes/mender-setup.bbclass
@@ -262,6 +262,9 @@ python() {
 
         # Setup the systemd machine ID to be persistent across OTA updates.
         'mender-persist-systemd-machine-id',
+
+        # Enable dynamic resizing of the data filesystem through systemd's growfs
+        'mender-growfs-data',
     }
 
     mfe = d.getVar('MENDER_FEATURES_ENABLE')
@@ -392,6 +395,8 @@ python mender_vars_handler() {
         with open (path, 'w') as f:
             json.dump(mender_vars, f, sort_keys=True, indent=4)
 }
+
+MENDER_DATA_PART_FSTAB_OPTS_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-growfs-data', ',x-systemd.growfs', '', d)}"
 
 # Including these does not mean that all these features will be enabled, just
 # that their configuration will be considered. Use DISTRO_FEATURES to enable and
--- a/meta-mender-core/classes/mender-full.bbclass
+++ b/meta-mender-core/classes/mender-full.bbclass
@@ -6,6 +6,7 @@ MENDER_FEATURES_ENABLE_append = " \
     ${_MENDER_IMAGE_TYPE_DEFAULT} \
     mender-install \
     mender-systemd \
+    mender-growfs-data \
 "
 
 _MENDER_IMAGE_TYPE_DEFAULT ?= "mender-image-uefi"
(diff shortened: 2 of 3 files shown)


meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -63,7 +63,7 @@ MENDER_MTD_UBI_DEVICE_NAME ??= "${MENDER_MTD_UBI_DEVICE_NAME_DEFAULT}"
 MENDER_MTD_UBI_DEVICE_NAME_DEFAULT = ""
 
 # Filesystem type of data partition. Used for both FS generation and fstab construction
-# Leave as default (auto) to generate a partition using the same Filesystem as
```
oleorhagen: Nothing major, but in general this will dirty the blame-log.

mirzak: My editor does this, and I was hoping that this would  slip trough :)

oleorhagen: Reason I'm saying it is because I'm getting yelled at for this also. The solution I found was to add a git-hook instead. So that only the staged git diff will be whitespace-filtered ;)

Gonna let it slide for now though :p

mirzak: I have updated :)

oleorhagen:
Looks beautiful ^^

kacf:
Some pretty [massive failures](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/82696292). Let's hope it was some problem with qemu. [Rerunning](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/83102124).

kacf:
This appears to be persistently failing. The log is difficult to interpret, but there are loads of error messages in QEMU boot. One that stands out is `Failed to start Mender service to grow data partition size`.

mirzak:
I will take a closer look.

mirzak:
> Failed to start Mender service to grow data partition size.

Found why this was failing, was using `DEPENDS` instead of `REPENDS` and it worked in my environment because it was polluted with a `IMAGE_INSTALL_append = " parted"`

I have pushed an update, and lets see if it behaves better.

kacf:
This still fails. A couple of messages stand out to me:

```
[ SKIP ] Ordering cycle found, skipping Grow File System on /data
[ SKIP ] Ordering cycle found, skipping Mend…ervice to grow data partition size
...
[  OK  ] Created slice system-systemd\x2dgrowfs.slice.
```

mirzak:
> This still fails. A couple of messages stand out to me:
```
[ SKIP ] Ordering cycle found, skipping Grow File System on /data
[ SKIP ] Ordering cycle found, skipping Mend…ervice to grow data partition size
[  OK  ] Created slice system-systemd\x2dgrowfs.slice.
...

Will take a look. 

It seems that above errors are only shown for `vexpress_qemu` but not on the `qemux86_64` (unclear why this failed).

**Comment by mirzak** [comment]
Been wrestling with the report problems locally and a `vexpress_qemu`. Pushed an update which  
eliminates the `Ordering cycle found...` errors.

**Comment by kacf** [comment]
Alright, finally passed!

File: meta-mender-core/classes/mender-setup.bbclass [hunk 2]
```diff
@@ -393,6 +396,8 @@ python mender_vars_handler() {
             json.dump(mender_vars, f, sort_keys=True, indent=4)
 }
 
+MENDER_DATA_PART_FSTAB_OPTS_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-growfs-data', ',x-systemd.growfs', '', d)}"
```
kkoopa: This is specific to systemd and must also test for `mender-systemd`.

PR #826, opened 2019-09-24
Fix and strict check for mender-migrate-configuration MEN-2757

lluiscampos:
In the one hand, fix the recipe itself by moving the cp statement to
do_compile as required by mender-state-scripts class.

In the other hand, this recipe will only work if split-mender-config is
disabled in the same build. Add a hard check for this making it depend
on mender and document the improve the usage doc in the changelog entry.

Cancel-changelog: 5acbed2

Changelog: Splits the mender.conf configuration file into a transient
configuration /etc/mender/mender.conf and a persistent congiguration in
/data/mender/mender.conf. This split is enabled by default, and it can
be opt-out by adding `PACKAGECONFIG_remove = "split-mender-config"` to
local.conf.

Changelog: Add a new recipe mender-migrate-configuration to ease the
migration of devices with single mender.conf to the new setup of split
configuration files. To build an updgrade for such device, add the
recipe with `IMAGE_INSTALL_append = " mender-migrate-configuration"`,
disable the split feature (for this update only) with
`PACKAGECONFIG_remove = " split-mender-config"`, and specify which
parameters to migrate (at least partitions params) with
`MENDER_PERSISTENT_CONFIGURATION_VARS = "RootfsPartA RootfsPartB"`.

### DIFF
--- a/meta-mender-core/recipes-mender/mender-migrate-configuration/mender-migrate-configuration.bb
+++ b/meta-mender-core/recipes-mender/mender-migrate-configuration/mender-migrate-configuration.bb
@@ -14,6 +14,16 @@ ALLOW_EMPTY_${PN} = "1"
 
 RDEPENDS_${PN} += "jq"
 
+DEPENDS += "mender"
+
+do_check_split_conf() {
+    if [ -f ${STAGING_DIR_TARGET}/data/mender/mender.conf ]; then
+        bbfatal "A persistent Mender configuration file was found on the /data partition. To migrate the configuration, \
+disable split-mender-config by adding PACKAGECONFIG_remove = \" split-mender-config\" to your local.conf"
+    fi
+}
+addtask do_check_split_conf before do_compile
+
 do_compile[vardeps] += "MENDER_PERSISTENT_CONFIGURATION_VARS"
 do_compile() {
 
@@ -31,8 +41,7 @@ do_compile() {
     # Replace the program markers in the script with the jq programs generated above.
     sed -i "s/%jq-program-marker%/${MENDER_JQ_PROGRAM}/" mender-migrate-configuration
     sed -i "s/%jq-delete-fields-marker%/${MENDER_JQ_DELETE}/" mender-migrate-configuration
-}
 
-do_install() {
+    # Deploy script as an State Script
     cp mender-migrate-configuration ${MENDER_STATE_SCRIPTS_DIR}/ArtifactCommit_Enter_10_migrate-configuration
 }
--- a/meta-mender-core/recipes-mender/mender/mender.inc
+++ b/meta-mender-core/recipes-mender/mender/mender.inc
@@ -33,6 +33,8 @@ FILES_${PN} += "${systemd_unitdir}/system/mender.service \
                 ${sysconfdir}/udev/mount.blacklist.d/mender \
                "
 
+SYSROOT_DIRS += "/data"
+
 SRC_URI_append_mender-image_mender-systemd = " file://mender-data-dir.service"
 SRC_URI_append_mender-persist-systemd-machine-id = " file://mender-systemd-machine-id.service \
                                                      file://mender-set-systemd-machine-id.sh"


lluiscampos:
@oleorhagen @drewmoseley @kacf ^

lluiscampos:
Decided not to add an acceptance test. Reasoning: it is not a simple test and support to upgrade from thud to warrior is best-effort only.

Stopping the automated build as there are no tests for this recipe.

meta-mender-core/recipes-mender/mender-migrate-configuration/mender-migrate-configuration.bb
```diff
@@ -14,6 +14,17 @@ ALLOW_EMPTY_${PN} = "1"
...
 RDEPENDS_${PN} += "jq"
 
+# NOTE: Implicitly append split-mender-config (same as in mender.inc) and expect the
+# user to remove it explicitly in the local.conf
+PACKAGECONFIG[split-mender-config] = ",,,"
+PACKAGECONFIG_append = " split-mender-config"
+
+python do_check_config() {
+    if bb.utils.contains('PACKAGECONFIG', 'split-mender-config', True, False, d):
+        bb.fatal("'split-mender-config' needs to be removed for this migration recipe to work. Use PACKAGECONFIG_remove = \" split-mender-config\" in your local.conf")
```
kacf: I think this isn't a good way. `PACKAGECONFIG` is supposed to be package specific, and using a global `PACKAGECONFIG_remove` is against convention. Tweaks are normally added in `.bbappend` files instead. Can we do this in a "softer" way, and in the state script detect if `mender.conf` is already present in `/data`? If it is, we can simply not update it.

lluiscampos: Done.

lluiscampos:
Fixed also the Changelog text following @oleorhagen comments in private chat.

oleorhagen:
```and specify which
parameters to migrate (at least partitions params) with
`MENDER_PERSISTENT_CONFIGURATION_VARS = "RootfsPartA RootfsPartB"`.```
Are we requiring that the user sets the rfsA and B to be persistent now?

**Comment by lluiscampos** [comment]
> ```
> parameters to migrate (at least partitions params) with
> `MENDER_PERSISTENT_CONFIGURATION_VARS = "RootfsPartA RootfsPartB"`.```
> Are we requiring that the user sets the rfsA and B to be persistent now?
> ```

When using split-mender-config, RootfsPartA and RootfsPartB are stored in the persistent config file. So in order to use this feature later on, you have to migrate them during this upgrade.

That was the issue that I was having in my local setup

**Comment by lluiscampos** [comment]
> When using split-mender-config, RootfsPartA and RootfsPartB are stored in the persistent config file. So in order to use this feature later on, you have to migrate them during this upgrade.

That is if you use the default config, of course:
```
MENDER_PERSISTENT_CONFIGURATION_VARS ?= "RootfsPartA RootfsPartB"
```

But I think that clarifying that that depends on your MENDER_PERSISTENT_CONFIGURATION_VARS will make this Changelog way to long and difficult to understand.

So in short: If you follow that steps you have a safe way to migrate. If you are already using MENDER_PERSISTENT_CONFIGURATION_VARS it means that you are an advance user of this feature and probably you already know what you are doing.

**Comment by oleorhagen** [comment]
kk. LGTM. Maybe reword `at least partittions params` to `at least the partition parameters` (or variables or whatever)

**Comment by lluiscampos** [comment]
Done. Merging.

File: meta-mender-core/recipes-mender/mender-migrate-configuration/mender-migrate-configuration.bb [hunk 2]
```diff
@@ -14,6 +14,16 @@ ALLOW_EMPTY_${PN} = "1"
 
 RDEPENDS_${PN} += "jq"
 
+DEPENDS += "mender"
+
+do_check_split_conf() {
+    if [ -f ${STAGING_DIR_TARGET}/data/mender/mender.conf ]; then
```
kacf: `STAGING_DIR_TARGET`, aha!

lluiscampos: Yes. But the key was to add `SYSROOT_DIRS += "/data"` to the mender recipe so that it is available for other recipes

PR #829, opened 2019-09-26
add support for Raspberry Pi 4

mirzak:
Changelog: Title


mirzak:
Should go in after, https://github.com/mendersoftware/meta-mender/pull/828

kacf:
Nice!

For purposes of releasing, do you know if the RPi3 images run on RPi4?

mirzak:
> For purposes of releasing, do you know if the RPi3 images run on RPi4?

Unfortunately not. Because of U-boot.

kacf:
Ok, good to know. I suppose at some point we should start building release images for Rpi4. But I think it's not that urgent yet.

PR #835, opened 2019-10-02
systemd: move network configurations to systemd-conf

jorisoffouga:
systemd-conf is a better place for machine configuration.
See http://lists.openembedded.org/pipermail/openembedded-core/2019-May/282614.html

Signed-off-by: Joris Offouga <offougajoris@gmail.com>


texierp:
Interesting ... It's something I was going to send.

But your PR seems not valable in master since this [commit](https://git.openembedded.org/openembedded-core/commit/?id=d87efd14ce0471135c0aa7fd7b5da2808acb9c76).

See https://github.com/texierp/meta-mender/commit/7326256d75ec674af6ff09ebd7fcb031d6436ae2

kacf:
On the surface this looks alright, but I will have to test it next week with some of our real devices.

kacf:
Tested manually, and works. Thanks, and sorry for the delay!

jorisoffouga:
No problem thanks!

PR #836, opened 2019-10-04
MEN-1670: Enable Artifact Depends and Provides for Yocto builds

oleorhagen:
Enables the ability to set:

* Artifact Depends
* Artifact Provides
* Artifact Name Depends
* Artifact Provides Group
* Artifact Depends Groups

In the Mender Artifact from the Yocto build,
through the build variables:

ARTIFACT_NAME_DEPENDS -- List of names
ARTIFACT_PROVIDES -- Key:Value
ARTIFACT_DEPENDS -- Key:Value
ARTIFACT_PROVIDES_GROUP -- Name
ARTIFACT_DEPENDS_GROUPS -- List of names

Changelog: Commit


oleorhagen:
Depends on: https://github.com/mendersoftware/mender-artifact/pull/234

meta-mender-core/classes/mender-artifactimg.bbclass
```diff
@@ -55,6 +63,26 @@ IMAGE_CMD_mender () {
...
+    if [ -n "${ARTIFACT_PROVIDES_GROUP}" ]; then
+        extra_args="$extra_args --provides-group ${ARTIFACT_PROVIDES_GROUP}"
+    fi
+
+    if [ -n "${ARTIFACT_DEPENDS}" ]; then
+        extra_args="$extra_args --depends ${ARTIFACT_DEPENDS}"
+    fi
+
+    if [ -n "${ARTIFACT_DEPENDS_GROUPS}" ]; then
+        extra_args="$extra_args --depends-groups ${ARTIFACT_DEPENDS_GROUPS}"
```
kacf: If there is more than one entry, don't you need to loop over these and do some joining with `--depends-groups` in between (same for the others)?

oleorhagen:
> I also think this needs an acceptance test where the variables are used, for example in `test_build.py` or `test_mender-artifact.py`. Can do all in the same test to save time, but make sure to have some with single entries and some with multiple entries.

Sure, I would feel much more comfortable with a test here. I am curious as to the variable names though. Should they be `MENDER_*` variables you think?

meta-mender-core/classes/mender-artifactimg.bbclass
oleorhagen: I guess it depends on how we require the input to be formed, right?
It should be fine with a string if the input is formatted like MA expects, I think..

kacf: It doesn't work with `--depends-groups group1 group2`, only with `--depends-groups group1 --depends-groups group2`. It's quite buggy in that respect actually, because it *silently* accepts the incorrect command, and then uses only the first element.

kacf:
> I am curious as to the variable names though. Should they be `MENDER_*` variables you think?

Yes, good point! I agree, they should have `MENDER_` prefixes.

meta-mender-core/classes/mender-artifactimg.bbclass
oleorhagen: haha, lol, Yeh, I discovered it during testing :p

oleorhagen: https://tracker.mender.io/browse/MEN-2849

meta-mender-core/classes/mender-artifactimg.bbclass
```diff
@@ -18,6 +18,31 @@ ARTIFACTIMG_FSTYPE_DEFAULT = "ext4"
...
 
+MENDER_ARTIFACT_NAME_DEPENDS ?= ""
+
+MENDER_ARTIFACT_PROVIDES ?= ""
+MENDER_ARTIFACT_PROVIDES_GROUP ?= ""
+
+MENDER_ARTIFACT_DEPENDS ?= ""
+MENDER_ARTIFACT_DEPENDS_GROUPS ?= ""
+
+apply_arguments () {
```
oleorhagen: Happy to receive some suggestions for improvements on this one

meta-mender-core/classes/mender-artifactimg.bbclass
```diff
@@ -55,6 +80,36 @@ IMAGE_CMD_mender () {
         image_flag=-f
     fi
 
+    if [ -n "${MENDER_ARTIFACT_NAME_DEPENDS}" ]; then
```
oleorhagen: I know we are basically repeating the same thing here, but I like the expliciity of it. Happy to turn it into a function though, if sought for.

meta-mender-core/classes/mender-artifactimg.bbclass
```diff
@@ -18,6 +18,31 @@ ARTIFACTIMG_FSTYPE_DEFAULT = "ext4"
...
+apply_arguments () {
+    #
+    # $1 -- the command line flag to apply to each element
+    # $@ -- the list of arguments to give each its own flag
+    #
+    local res=""
+    flag=$1
+    shift
+    OIFS=$IFS
+    IFS=" "
```
kacf: Is this necessary? I think `IFS` already breaks on spaces if you don't specify it. Also, despite looking correct, I don't think that

```
OIFS="$IFS"
IFS="$OIFS"
```

resets it to the same state, because empty `IFS` is not the same as unset `IFS`, and in that sequence of commands the shell can't tell the difference. Not 100% sure though.

meta-mender-core/classes/mender-artifactimg.bbclass
kacf: I like having a generic function. :-)

meta-mender-core/classes/mender-artifactimg.bbclass
kacf: No, I also like it. It's still pretty minimal, but verbose enough to be explicit.

tests/acceptance/test_build.py
```diff
@@ -567,3 +567,162 @@ def test_correct_partition_types(self, latest_part_image):
...
+            provides_group="providesgroupname",
+            depends={
+                "dependskey1": "dependsval1",
+                "dependskey2": "dependsval2"
+            },
+            depends_groups=["depenceygroup1",
+                            "depenceygroup2"])
+    ]
+
+    @pytest.mark.min_mender_version('1.0.0')
```
kacf: These parameters were introduced in Mender 2.0.0.

tests/acceptance/test_build.py
```diff
@@ -567,3 +567,162 @@ def test_correct_partition_types(self, latest_part_image):
...
+            self.depends_groups = depends_groups
+
+        def _flatten_dict(self, d):
+            l = ""
+            for pair in d.items():
+                l += ":".join(pair)
+                l += " "
+            return l
+
+        def __str__(self):
```
kacf: Nice use of `__str__`!

oleorhagen:
@kacf removed the `IFS` caching, and set min version to `2.0.0`. Should be good to go

PR #772, opened 2019-06-26
Drew sumo wip with license fixes

lluiscampos:
Supersedes #771 adding the correct license checksums for Git recipes

PR #838, opened 2019-10-08
Remove several mender and mender-artifact artifact versions.

kacf:
Changelog: Remove mender-2.0.x and older, and mender-artifact 3.0.x
and older.


lluiscampos:
Nice clean-up. I'll get this change also in warrior

lluiscampos:
> I'll get this change also in warrior

Well, according to PR #772 we keep these recipes in warrior. Can we discuss this a bit? At least we could stop supporting 1.x in warrior (not yet released any tag there)

mirzak:
> I'll get this change also in warrior

I do not think that we should. There are chances that people are still using 1.7 even if they are on warrior and we would break this. It is to late to clean up now :)

lluiscampos:
> > I'll get this change also in warrior
> 
> I do not think that we should. There are chances that people are still using 1.7 even if they are on warrior and we would break this. It is to late to clean up now :)

I agree. We will remove it only in master

kacf:
Yes, recipe removals are only done in master, in preparation for the next Yocto release.

kacf:
Only known spurious errors. Merging.

PR #847, opened 2019-10-23
Actually rename mender-artifact 3.2.0b1 to 3.2.0.

kacf:
This was forgotten in the last commit and will lead to wrong version
number in the built binary.

Changelog: None


kacf:
Only running build. The SHA of the beta and the final is the same, so there is no chance of regressions.

PR #851, opened 2019-10-29
Remove explicit tag 2.1.1 from licenses checksum in mender git recipe

lluiscampos:
It is not really required so let's try to keep this function clean.

Changelog: None


lluiscampos:
Stopping the CI run after the build phase passed.

PR #854, opened 2019-11-04
Update min_mender_version sentinel to 2.2.0 for artifact deps tests

oleorhagen:
The current mender-version guard had a version of 2.0.0, which is too low, since
the changes in mender-artifact which holds the depends and provides
functionality is only available in master at the moment of this writing. Hence
set the tag to 2.2.0, which for now only includes the master branch.

Without this fix, cherry-picks to older branches will fail the acceptance tests,
if they employ older versions of mender-artifact.

Changelog: None

### DIFF
--- a/tests/acceptance/test_build.py
+++ b/tests/acceptance/test_build.py
@@ -707,7 +707,7 @@ deployed-test-dir9/*;renamed-deployed-test-dir9/ \
                             "depenceygroup2"])
     ]
 
-    @pytest.mark.min_mender_version('2.0.0')
+    @pytest.mark.min_mender_version('2.2.0')
     @pytest.mark.parametrize('dependsprovides', test_cases)
     def test_build_artifact_depends_and_provides(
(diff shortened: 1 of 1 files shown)


oleorhagen:
Stopped the integration tests build, though no real integration test changes have been made. Will re-run together with: https://github.com/mendersoftware/mender-artifact/pull/242 before merging.

oleorhagen:
@lluiscampos I can't really see that this should be cherry-picked anywhere. The test lives in master only this far (for meta-mender), and the same goes for the functionality which is being tested with mender-artifact (which still is only on master also).

Am I missing something?

lluiscampos:
Yes, this does not need cherry-picking and can be merged as-is.

It will probably land in warrior together with the test itself, but that will happen on the next round of warrior updates.

oleorhagen:
@lluiscampos great. Could you force-merge this then?

PR #863, opened 2019-11-21
uboot_auto_patch: handle more complex dependency strings in add_kconf… 

joergho:
…ig_option_with_depends script

The change makes the script capable of handling more complex depends-strings which
contain or-operators ("||") where the first alternative is used, parentheses, and
and-operators ("&&").

This makes the script compatible with U-Boot 2019.10.

U-Boot 2019.10 has the following ENV_OFFSET dependencies:
    depends on (!ENV_IS_IN_UBI && !ENV_IS_NOWHERE) || ARCH_STM32MP
(see https://gitlab.denx.de/u-boot/u-boot/blob/v2019.10/env/Kconfig#L469)

In comparison, U-Boot 2019.01 has the following ENV_OFFSET dependencies:
    depends on !ENV_IS_IN_UBI
    depends on !ENV_IS_NOWHERE
(see https://gitlab.denx.de/u-boot/u-boot/blob/v2019.01/env/Kconfig#L436)

Signed-off-by: Joerg Hofrichter <joerg.hofrichter@ni.com>

### DIFF
--- a/meta-mender-core/recipes-bsp/u-boot/files/add_kconfig_option_with_depends.py
+++ b/meta-mender-core/recipes-bsp/u-boot/files/add_kconfig_option_with_depends.py
@@ -16,6 +16,22 @@ parser.add_argument("option", metavar="OPTION", nargs=1,
                     help="Option to add to Kconfig, expressed as KEY=VALUE.")
 args = parser.parse_args()
 
+def parse_dependencies(depends_string):
+    # if depends_string has multiple alternative dependencies ("A || B"),
+    # select the first one ("A")
+    depends_string = re.split("\s*\|\|\s*", depends_string)[0]
+
+    # if there are parentheses around ("(A && B)"), remove them ("A && B")
+    match = re.match("^\(\s*(.*)\s*\)$", depends_string)
+    if match:
+        depends_string = match[1]
+
+    # if depends_string has multiple required dependencies ("A && B"),
+    # return all of them
+    dependencies = re.split("\s*&&\s*", depends_string)
+
+    return dependencies
+
 def add_kconfig_option(option):
     key, value = option.split("=", 1)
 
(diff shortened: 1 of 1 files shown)


oleorhagen:
Hi @joergho 

Thank you for your contribution. It seems to me that you did not add a `Changelog` tag. Could you amend this?

joergho:
@oleorhagen Sure I can add the tag. Is "Changelog: none" correct in this case?

oleorhagen:
I think I would add `Changelog: Title` ;)

mirzak:
I would actually put:

```
Changelog: Update uboot_auto_patch script to be compatible with U-Boot 2019.10
```

Note the Changelog entries are user facing and ends up release change log.

joergho:
> I would actually put:
> Changelog: Update uboot_auto_patch script to be compatible with U-Boot 2019.10

I added this in baa2d77.

mirzak:
The changes look good to me, will start build and tests.

kkoopa:
This is trying to  parse what appears to be a context-free language as a regular language.
`parse_dependencies('(A || B) && C')` gives `['(A']`.

oleorhagen:
> This is trying to parse what appears to be a context-free language as a regular language.
> `parse_dependencies('(A || B) && C')` gives `['(A']`.

Damn, well spotted! 
I guess this is not the intended functionality @joergho

joergho:
> I guess this is not the intended functionality

The function parse_dependencies was tailored around the problems I was seeing with U-Boot 2019.10 but it does not implement sophisticated logic parsing which would be required for other testcases you were testing for. Therefore the failing for more complex terms was to be expected....

I prepared another version where where I use the "boolean" library to do proper parsing:
https://github.com/joergho/meta-mender/tree/u-boot-add-kconfig-fix-v2

Please decide if you think this is the better approach. If so, I'd suggest to close this PR and create a new PR from the other branch. Also, please let me know if you accept the dump copy of the boolean.py file or if you prefer a proper python3-boolean recipe (which I would suggest to place at meta-mender-core/recipes-devtools/python).

oleorhagen:
I will leave @mirzak to make that decision ;)

kacf:
@joergho: Yes, that is a very good approach, looks robust! And including `boolean.py` is fine, since it's BSD-licensed. Thanks for working on this!

joergho:
Ok, then I close this PR in favor of the new PR #865 which uses the "boolean.py" approach.

PR #864, opened 2019-11-28
QA-120: Acceptance tests python3 for mender-convert 

lluiscampos:
Missing fixes from PR #849. See individual commits.

### DIFF
--- a/tests/acceptance/common.py
+++ b/tests/acceptance/common.py
@@ -269,6 +269,9 @@ def common_boot_from_internal(conn):
 def latest_build_artifact(builddir, extension, sdimg_location=None):
     global configuration
 
+    # Force the builddir to be an absolute path
+    builddir = os.path.abspath(builddir)
+
     if configuration.get("conversion"):
         output = subprocess.check_output(["sh", "-c", "ls -t %s/*%s | grep -v data*%s| head -n 1" % (builddir, extension, extension)])
     else:
--- a/tests/acceptance/common.py
+++ b/tests/acceptance/common.py
@@ -281,20 +281,29 @@ def latest_build_artifact(builddir, extension, sdimg_location=None):
     return output
 
 def get_bitbake_variables(target, env_setup="true", export_only=False):
-    current_dir = os.open(".", os.O_RDONLY)
-    os.chdir(os.environ['BUILDDIR'])
-
     global configuration
+    lines = []
 
     if configuration.get("conversion"):
         config_file_path = os.path.abspath(configuration["test_variables"])
         with open(config_file_path, 'r') as config:
-            output = config.readlines()
+            lines = config.readlines()
     else:
+        current_dir = os.open(".", os.O_RDONLY)
+        os.chdir(os.environ['BUILDDIR'])
         ps = subprocess.Popen("%s && bitbake -e %s" % (env_setup, target),
                                   stdout=subprocess.PIPE,
                                   shell=True,
                                   executable="/bin/bash")
+        os.fchdir(current_dir)
+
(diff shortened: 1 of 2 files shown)


tests/acceptance/common.py
```diff
@@ -268,51 +268,53 @@ def common_boot_from_internal(conn):
...
-        output = subprocess.check_output(["sh", "-c", "ls -t %s/%s/*%s | grep -v data*%s| head -n 1" % (builddir, sdimg_location, extension, extension)])
+        output = subprocess.check_output(["sh", "-c", "ls -t %s/*%s | grep -v data*%s| head -n 1" % (builddir, extension, extension)])
     else:
         output = subprocess.check_output(["sh", "-c", "ls -t %s/tmp*/deploy/images/*/*%s | grep -v data*%s| head -n 1" % (builddir, extension, extension)])
     output = output.decode().rstrip('\r\n')
-    print("Found latest image of type '%s' to be: %s" % (extension, output))
-    return output
+    path_to_image = os.path.abspath(output)
+    print("Found latest image of type '%s' to be: %s" % (extension, path_to_image))
+    return path_to_image
```
lluiscampos: @pasinskim FYI this is the change that makes the tests fail. Probably the working dir is not what I assumed to be. I don't have more time today but I'll look into it tomorrow.

tests/acceptance/common.py
```diff
@@ -268,51 +268,53 @@ def common_boot_from_internal(conn):
...
+
     if configuration.get("conversion"):
-        output = subprocess.check_output(["sh", "-c", "ls -t %s/%s/*%s | grep -v data*%s| head -n 1" % (builddir, sdimg_location, extension, extension)])
+        output = subprocess.check_output(["sh", "-c", "ls -t %s/*%s | grep -v data*%s| head -n 1" % (builddir, extension, extension)])
     else:
         output = subprocess.check_output(["sh", "-c", "ls -t %s/tmp*/deploy/images/*/*%s | grep -v data*%s| head -n 1" % (builddir, extension, extension)])
     output = output.decode().rstrip('\r\n')
-    print("Found latest image of type '%s' to be: %s" % (extension, output))
-    return output
+    path_to_image = os.path.abspath(output)
```
pasinskim: Just a slight detail, but in this case. maybe we can remove above `rstrip('\r\n')` call?

tests/acceptance/common.py
```diff
@@ -268,51 +268,53 @@ def common_boot_from_internal(conn):
...
         export_only_expr = "?"
     matcher = re.compile('^(?:export )%s([A-Za-z][^=]*)="(.*)"$' % export_only_expr)
     ret = {}
-    while True:
-        line = ps.stdout.readline()
-        if not line:
-            break
-        line = line.decode().strip()
+    for line in lines:
+        line = line.strip()
```
pasinskim: Can't we strip lines as we read those here: https://github.com/mendersoftware/meta-mender/pull/864/files#diff-1341c7e04ba8ae1f7491966f6793b12dR302

lluiscampos: We would then need to strip there and in the config line parsing `lines = config.readlines()`. It is quite harmless so I'll leave it.

tests/acceptance/common.py
lluiscampos: I have reworked this change.

tests/acceptance/common.py
lluiscampos: Reworked

lluiscampos:
All acceptance tests passed after my rework. Merging!

PR #867, opened 2019-12-03
A few random changes

drewmoseley:


meta-mender-qemu/docker/build-docker
```diff
@@ -76,6 +76,7 @@ if [ $BUILD_DEFAULTS = 1 ]; then
 
     if egrep -q '\bmender-image-uefi\b' <<<"$DISTRO_FEATURES"; then
         DISK_IMG="$IMAGE_NAME-$MACHINE.uefiimg"
+        [ -e $BUILDDIR/tmp/deploy/images/$MACHINE/$DISK_IMG ] || DISK_IMG="$IMAGE_NAME-$MACHINE-grub.uefiimg"
```
kacf: I'm missing some context here. In which case is this needed?

drewmoseley: I tried to run it yesterday on a vexpress-qemu build and it was necessary.

drewmoseley:
I also added a commit here to create a data.tar deploy image. I need it for Toradex Easy Installer integration.

meta-mender-qemu/docker/build-docker
kacf: Ah, it's because `MENDER_DEVICE_TYPE` is sometimes not the same as `MACHINE`. This is not very obvious, and also the fix is not generic. Can I suggest this fix instead:

```
diff --git a/meta-mender-qemu/docker/build-docker b/meta-mender-qemu/docker/build-docker
index 223497e..4679c75 100755
--- a/meta-mender-qemu/docker/build-docker
+++ b/meta-mender-qemu/docker/build-docker
@@ -75,7 +75,9 @@ if [ $BUILD_DEFAULTS = 1 ]; then
     eval "$(bitbake -e $IMAGE_NAME | egrep '^DISTRO_FEATURES=')"
 
     if egrep -q '\bmender-image-uefi\b' <<<"$DISTRO_FEATURES"; then
-        DISK_IMG="$IMAGE_NAME-$MACHINE.uefiimg"
+        # MENDER_DEVICE_TYPE is sometimes not the same as MACHINE, so use
+        # wildcard to select correct image.
+        DISK_IMG="$(basename $BUILDDIR/tmp/deploy/images/$MACHINE/$IMAGE_NAME-$MACHINE*.uefiimg)"
     elif egrep -q '\bmender-image-sd\b' <<<"$DISTRO_FEATURES"; then
         DISK_IMG="$IMAGE_NAME-$MACHINE.sdimg"
     else
```

drewmoseley: Can we guarantee that wildcard will never match more than 1 file?

kacf: I know that Bitbake removes old artifacts automatically, but whether it's truly guaranteed, I can't say. If we want to ensure it we can add this:

```
DISK_IMG="$(basename $(ls --sort=time $BUILDDIR/tmp/deploy/images/$MACHINE/$IMAGE_NAME-$MACHINE*.uefiimg | head -n1))"
```

That picks the latest one which matches criteria.

drewmoseley: Updated as requested

kacf:
Superseded by #877.

meta-mender-qemu/docker/build-docker
kacf: This turned out to be trickier than I thought. Can you take a look at #877, and verify that this still fixes your problem?

PR #868, opened 2019-12-05
u-boot-mender: set correct boot command for fitImage

joergho:
When u-boot is built with KERNEL_IMAGETYPE="fitImage", the wrong boot command
(MENDER_BOOT_KERNEL_TYPE) was chosen in do_provide_mender_defines because
the selection defaults to "bootz". The correct command is "bootm" because a
fitImage is an extended uImage format. For details about this format, see:
https://github.com/u-boot/u-boot/blob/master/doc/uImage.FIT/howto.txt

Changelog: none

Signed-off-by: Joerg Hofrichter <joerg.hofrichter@ni.com>


kacf:
Unrelated spurious error. Merging.

kacf:
Thanks @joergho!

PR #875, opened 2019-12-09
mender: Ensure excluded directories work for all image types.

drewmoseley:
This commit changes the functionality of the directory exclusion to simply backup
the existing one and restoring it when done.  The previous version relied on modifying
the IMAGE_ROOTFS variable but that did not work for image commands such as IMAGE_CMD_tar
since that is defined as a variable rather than a function and would be fully expanded
before we modified the IMAGE_ROOTFS variable.

Changelog: Title

### DIFF
--- a/meta-mender-core/classes/mender-setup-image.inc
+++ b/meta-mender-core/classes/mender-setup-image.inc
@@ -86,26 +86,25 @@ python prepare_excluded_directories() {
     exclude_list = exclude_var.split()
 
     rootfs_orig = d.getVar('IMAGE_ROOTFS')
-    # We need a new rootfs directory we can delete files from. Copy to
-    # workdir.
-    new_rootfs = os.path.realpath(os.path.join(d.getVar("WORKDIR"), "rootfs.%s" % taskname))
+    # We need a backup rootfs directory so we can restore during the cleanup phase.
+    rootfs_backup = os.path.realpath(os.path.join(d.getVar("WORKDIR"), "rootfs.%s" % taskname))
 
-    if os.path.lexists(new_rootfs):
-        shutil.rmtree(os.path.join(new_rootfs))
+    if os.path.lexists(rootfs_backup):
+        shutil.rmtree(os.path.join(rootfs_backup))
 
-    copyhardlinktree(rootfs_orig, new_rootfs)
+    copyhardlinktree(rootfs_orig, rootfs_backup)
 
     for orig_path in exclude_list:
         path = orig_path
         if os.path.isabs(path):
             bb.fatal("IMAGE_ROOTFS_EXCLUDE_PATH: Must be relative: %s" % orig_path)
 
-        full_path = os.path.realpath(os.path.join(new_rootfs, path))
+        full_path = os.path.realpath(os.path.join(rootfs_orig, path))
 
         # Disallow climbing outside of parent directory using '..',
         # because doing so could be quite disastrous (we will delete the
         # directory).
-        if not full_path.startswith(new_rootfs):
+        if not full_path.startswith(rootfs_orig):
             bb.fatal("'%s' points to a path outside the rootfs" % orig_path)
 
         if not os.path.lexists(full_path):
@@ -125,8 +124,7 @@ python prepare_excluded_directories() {
             shutil.rmtree(full_path)
 
     # Save old value for cleanup later.
-    d.setVar('IMAGE_ROOTFS_ORIG', rootfs_orig)
-    d.setVar('IMAGE_ROOTFS', new_rootfs)
+    d.setVar('IMAGE_ROOTFS_BACKUP', rootfs_backup)
 }
 
 python cleanup_excluded_directories() {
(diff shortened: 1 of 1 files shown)


drewmoseley:
See discussion here: https://hub.mender.io/t/data-folder-present-in-mender-artifact-rootfs-update/1329/20
for more context.

kacf:
I don't think this will work correctly, since `do_image` tasks can run in parallel. For example `do_image_tar` may run at the same time as `do_image_dataimg`. I guess this is why the QA pipeline also failed. Theoretically we could get around this with a global file lock, but it's a bit of a nasty solution.

If the entire problem is that `IMAGE_CMD_tar` is a variable, then I suspect OE might accept a patch to change that, since it has virtually zero impact for them, and our reasons are quite solid.

drewmoseley:
> If the entire problem is that `IMAGE_CMD_tar` is a variable, then I suspect OE might accept a patch to change that, since it has virtually zero impact for them, and our reasons are quite solid.

There are quite a few IMAGE_CMD definitions which are variables rather than functions.  I wonder if there is anything else we can do here.  Any thoughts?

kacf:
My only concrete suggestion is the file lock. But maybe there is some way to force reevaluation of variables? Someone would have to dig into the evaluator API of bitbake.

dellgreen:
Would it be possible to fix up the image cmd variable replacing the evaluated IMAGE_ROOTFS path with the alternative path?

drewmoseley:
I'm not sure about modifying the variable.  That seems error prone but I'll give it some thought.

What about using a new task that is inserted before do_image()?  @kacf is there something that causes that not to work?

dellgreen:
@drewmoseley, i had a quick look at testing changing the image cmd variable within the prepare_excluded_directories &  cleanup_excluded_directories functions. 

Although I confirmed that the command was now pointing at the temporary rootfs.image_tar directory, I still ended up with excludes in the tar archive. So its not the whole picture as that change must come too late from with these functions

kacf:
> What about using a new task that is inserted before do_image()? @kacf is there something that causes that not to work?

Hmm, what would we do in this task? You can't assign variables in one task and have it picked up in another.

A couple of suggestions:

* See if you can move [the assignment](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/classes/mender-setup-image.inc#L129) from `prepare_excluded_directories` into [the place where the function is registered](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/classes/mender-setup-image.inc#L172) instead. This executes earlier, so maybe it's early enough that the variable expansion can still happen correctly. This means you will have to move the decision logic also here, so that you only make the assignment if none of the `return` statements in the original `prepare_excluded_directories` would have executed. I'm a bit worried that `BB_CURRENTTASK` won't be correctly defined in here though... you'll just have to try.

* There is actually an automatic [`[lockfiles]`](https://www.yoctoproject.org/docs/latest/mega-manual/mega-manual.html#variable-flags) mechanism available for tasks in Bitbake. This would make that solution a lot less ugly. It will slow down the build though, especially for short round-trip builds during development, since no `do_image` task will be able to execute in parallel. But if it's the only option, I guess we don't have a choice.

drewmoseley:
> Hmm, what would we do in this task? You can't assign variables in one task and have it picked up in another.

I was thinking we could actually remove the excluded directories in a separate task to avoid the need for changing IMAGE_ROOTFS.  Your concern earlier was that multiple do_image tasks could be running simultaneously so cleaning it up in a separate task that is a predecessor to all do_image tasks may address that concern.

kacf:
> I was thinking we could actually remove the excluded directories in a separate task to avoid the need for changing IMAGE_ROOTFS. Your concern earlier was that multiple do_image tasks could be running simultaneously so cleaning it up in a separate task that is a predecessor to all do_image tasks may address that concern.

The problem is that each task requires different content. You can't possibly run them in parallel if they use the same directory, because `do_image_ext4` requires content with exclusions removed, whereas at the same time, `do_image_dataimg` requires full content.

dellgreen:
any further thoughts on this?

drewmoseley:
Sorry, no. I've not had a chance to look any further into this.

dellgreen:
Ok, thanks for replying, I'll try and figure out a work around

dellgreen:
It looks like the issue is caused by the image.bbclass dynamic generation of the do_image_xxx tasks whereby the cmds are joined into a big string  to form the task script long before our functions or our anonymous python functions are called. Attempts so far to change the IMAGE_CMD_tar in the anonymous python block causes build errors with hash mismatch between the two parsing phases noticing that the IMAGE_CMD_tar variable has been changed twice.

I am yet to find a satisfactory solution, i suppose a quick fix for the do_image_tar task would be to add a postfunc that iterates over IMAGE_ROOTFS_EXCLUDE_PATH entries and deletes them from the tar using tar's delete option

dellgreen:
Post altering tar won't work as the compressed versions of the tar will be missed as code for that is also added to the  dynamic image task

dellgreen:
ok, i have an ugly solution hack for my scenario only for tar as I can fine no good solution to this after 2 days solidly on it, the problemn as lalready pointed on stems from the some IMAGE_CMDS_xxx being defined as variables rather than functions. I'll post back my ugly solution on mender hub as its not a fix for mender code

kacf:
Ok, thanks for the update, both here and on Mender Hub. As no solution seems to be in sight, I'm going to close this for now. If a solution comes up, we are of course very interested in a new PR.

PR #881, opened 2019-12-17
Use Python3 when starting the HTTP server and add connection checks

lluiscampos:
We are getting connection refused error some times in the client
acceptance tests. It does not seem possible to reproduce locally.

With this checks we can see if the problem is in the server side or
maybe in the QEMU networking (forwarding 10.0.2.2 to the host).

### DIFF
--- a/tests/acceptance/test_update.py
+++ b/tests/acceptance/test_update.py
@@ -24,6 +24,8 @@ import re
 import shutil
 import subprocess
 import tempfile
+import time
+import requests
 
 from common import *
 
@@ -112,14 +114,22 @@ class Helpers:
             Helpers.upload_to_s3(image)
             http_server_location = "{}/mender/temp".format(s3_address)
         else:
-            http_server = subprocess.Popen(["python", "-m", "SimpleHTTPServer"])
+            http_server = subprocess.Popen(["python3", "-m", "http.server"])
             assert(http_server)
+            time.sleep(1)
+            assert requests.head("http://localhost:8000/%s" % (image)).status_code == 200
 
         try:
             output = conn.run("mender %s http://%s/%s" % (install_flag, http_server_location, image))
             print("output from rootfs update: ", output.stdout)
         finally:
             if http_server:
+                try:
+                    status_code = requests.head("http://localhost:8000/%s" % (image)).status_code
+                    if status_code != 200:
+                        print("warning: http server is not accessible, status code %d" % (status_code))
+                except Exception as e:
+                    print("Exception during request" + str(e))
                 http_server.terminate()
 
 
--- a/tests/acceptance/requirements_py3.txt
+++ b/tests/acceptance/requirements_py3.txt
@@ -4,3 +4,4 @@ pytest-html
 paramiko
 python-lzo
 ubi-reader
+requests


tests/acceptance/test_update.py
```diff
@@ -112,14 +114,17 @@ def install_update(image, conn, http_server, board_type, use_s3, s3_address):
...
             assert(http_server)
+            time.sleep(1)
+            assert requests.head("http://localhost:8000").status_code == 200
 
         try:
             output = conn.run("mender %s http://%s/%s" % (install_flag, http_server_location, image))
             print("output from rootfs update: ", output.stdout)
         finally:
             if http_server:
+                assert requests.head("http://localhost:8000").status_code == 200
```
kacf: Why assert here? It's not very good style to have it in a cleanup function.

lluiscampos: Yes, that is not nice at all. But somewhat I want to check that the server is still up and running. I will add a simple print instead

lluiscampos: I reworked it and pushed. I will let all tests to run again to see if by a chance the issue appears.

tests/acceptance/test_update.py
```diff
@@ -112,14 +114,19 @@ def install_update(image, conn, http_server, board_type, use_s3, s3_address):
...
+            assert requests.head("http://localhost:8000/%s" % (image)).status_code == 200
 
         try:
             output = conn.run("mender %s http://%s/%s" % (install_flag, http_server_location, image))
             print("output from rootfs update: ", output.stdout)
         finally:
             if http_server:
+                status_code = requests.head("http://localhost:8000/%s" % (image)).status_code
+                if status_code != 200:
+                    print("warning: http server is not accessible, status code %d" % (status_code))
```
kacf: I *think* you will not typically reach this. `requests` will return normally as long as the server returns something, even if it's an error, but I think you will get an exception if the server can't be contacted at all. I think it's better to wrap this in:

```python
if http_server:
    try:
        status_code = requests.head("http://localhost:8000/%s" % (image)).status_code
        if status_code != 200:
            print("warning: http server is not accessible, status code %d" % (status_code))
    except Exception as e:
        print("Exception during request" + e.message())
    http_server.terminate()
```

or something along those lines (`e.message()` is probably not correct, I don't remember the notation off the top of my head).

lluiscampos: You are right. Thanks.

tests/acceptance/test_update.py
```diff
@@ -112,14 +114,24 @@ def install_update(image, conn, http_server, board_type, use_s3, s3_address):
...
             if http_server:
+                if http_server:
+                    try:
+                        status_code = requests.head("http://localhost:8000/%s" % (image)).status_code
+                        if status_code != 200:
+                            print("warning: http server is not accessible, status code %d" % (status_code))
+                    except Exception as e:
+                        print("Exception during request" + str(e))
+                    http_server.terminate()
                 http_server.terminate()
```
kacf: Hehe, sorry again, copy paste error, you have a double `if` statement and double `terminate` as well. :-)

Otherwise it looks good now.

lluiscampos: Lazy copy-pasting... Sorry.

lluiscampos:
I am stopping the builds this time. @kacf please have a last look and I'll merge.

PR #887, opened 2020-01-02
MEN-2799: Work around missing LSB support in zeus.

kacf:
Plus some other, smaller fixes. This is the most important one:

```
    Work around missing LSB support in zeus.
    
    This is a very intrusive change, which creates the `/lib64` symlink on
    the root filesystem for 64-bit systems. This symlink is normally
    missing when building with Yocto, but was provided by the LSB package
    previously. The directory is necessary to remain compatible with
    software built outside of Yocto. The Mender demo artifact, as well as
    the binary delta update module are examples of such components.
    
    If the link turns out to conflict with another package, it should be
    possible to work around the problem by adding this to the other
    package, either in the original `.bb` file or a `.bbappend` file:
    
    ```
    RPROVIDES_${PN} += "lsb-ld"
    ```
    
    Changelog: Commit
    
```

Note that I have deliberately omitted the patch which lists zeus as a compatible branch, because I don't want people to start using this until it's ready and tested.


kacf:
@lluiscampos, @mirzak, @drewmoseley: This is my suggested approach for dealing with missing LSB support in zeus onwards (see "Work around missing LSB support in zeus"). It's not pretty, and quite intrusive, but I cannot think of any other way to do it. It only affects x86-64, and I have tested that both x86-64 and ARM (RPi) works without the LSB package when using this alternative solution (for the latter it has no effect).

There some other smaller fixes too, but that is the main one I would like feedback on.

drewmoseley:
The meta-data itself looks straight forward; nice and simple.
Do we have any more context as to why it was removed?

kacf:
> Do we have any more context as to why it was removed?

The LSB spec is huge, this is only a small part. From Yocto's side it looks from the thread as if it's a combination of the spec being outdated, and finding a maintainer: see https://patches.openembedded.org/patch/164298/

kacf:
Thanks!

PR #890, opened 2020-01-04
Added handling for easily adding partitions

nandra:
This is not final stuff just RFC. It's done and added only for SDcard partitions (not ubi handling ...).

Idea:
- added new variable ```MENDER_EXTRA_PARTS``` which contains only partitions type, size (can be added also name of directory which should be added to partitions - not part of this PR)
e.g.: ```MENDER_EXTRA_PARTS = "--fixed-size 1024M --label=static --fstype=ext4;--fixed-size 12G --label=update --fstype=vfat"```

- separation for partitions is ```;``` character (handled in small python helper)
- user can add as much as needed
- not yet added automount handling of extra partitions

Just want to share this idea if it's preferred way to go. Thanks for any comment.


meta-mender-core/classes/mender-part-images.bbclass
```diff
@@ -19,6 +20,15 @@ python() {
             bb.fatal('Detected use of deprecated var %s, please replace it with %s in your setup' % (varname, newvarname))
 }
 
+def get_extra_parts(d):
+    final_parts = []
+    parts = d.getVar("MENDER_EXTRA_PARTS")
+    parts = parts.split(";")
+    for part in parts:
+      final_parts.append("part --ondisk \"$ondisk_dev\" --align $alignment_kb {}".format(part))
```
kacf: Nitpick: Should be indented eight spaces, not six.

meta-mender-core/classes/mender-part-images.bbclass
```diff
@@ -137,6 +147,15 @@ EOF
 
     cat >> "$wks" <<EOF
 part --source rawcopy --sourceparams="file=${IMGDEPLOYDIR}/${IMAGE_LINK_NAME}.dataimg" --ondisk "$ondisk_dev" --align $alignment_kb --fixed-size ${MENDER_DATA_PART_SIZE_MB} --mkfs-extraopts='${MENDER_DATA_PART_FSOPTS}' $part_type_params
+EOF
+# added extra partitions if exists
```
kacf: Nitpick: Should be indented four spaces.

nandra:
@kacf thanks for comments. I think it's better idea to use variable as you suggested. I'll rework it and test. What about automount? I was thinking (at least case I have) that I prefer to use systemd mount services for those new partitions with possibility to override (add checks, create some directories when it e.g. cannot be mounted and re-formatted ...). It will give more possibility for user if it's not automounted by /etc/fstab. Does that also sounds reasonable? Thanks.

kacf:
The idea sounds reasonable, but it's not clear to me how it would look. The fstab variant is straightforward, but in systemd there are so many options. My gut feeling is that it's better to just add systemd mount files as one would for any other service. But maybe I'm misunderstanding your intention. What exactly did you have in mind?

dellgreen:
I would be concerned that the wider acceptance criteria would be for them to go into /etc/fstab by default, and add enough support to allow disabling of automount via MENDER_EXTRA_PARTS so those that need custom systemd mounts can add them themselves.

nandra:
@kacf usually you need to write systemd mount file (idea was to be generated automagically) which contains info what to mount where and that's it. 
My use case is that if I cannot mount that partition then I perform re-format and create new one (this shouldn't happen but we want to use fat32 and this can be fragile).
If this is done by automount in /etc/fstab then systemd goes to service mode and device fails to start (had experience this issue already).

@dellgreen there will be default mount services which mount those partitions by default (as they would be in /etc/fstab) but then we have /data mounted by /etc/fstab and others by service files (maybe move /data mount also to mount service file would be an option?). 

Format of ```MENDER_EXTRA_PARTS``` is in wic format what about hide it from user and be more general to setup size, type, label and maybe partition whcih needs to be mounted and create wic format completely in bbclass? This is nice from user point of vies but if anyone would need to add something extra handling must be extended. Ideas to keep it like it is or? Thanks

dellgreen:
@nandra I maybe misunderstanding (and apologies for that if I am) but my feeling is that  /etc/fstab with (systemd-fstab-generator) by default as it currently is, may be the path to least disruption/resistance.

Either way whatever the consensus is I will be happy to work with :)

nandra:
I did small refactor and use variable as @kacf suggested. Move helpers to mender-helpers class. Add update of fstab with partitions. There are few weak points:
- mender-partuuid is not handled
- numbering device (/dev/mmcblk0X) is handled by +1 only
- because extended partition is created /data mount point must be shifted also by one.

Comments welcomed ;)

dellgreen:
Why does  date device number need to be incremented as I was under the impression that MENDER_DATA_PART_NUMBER calculation already took into account if extended partition is required because an optional swap partition can be inserted before the data partition already?

nandra:
@dellgreen thanks for pointer I checked ```mender_get_data_part_num``` which handle it. I'll take a look and update. :+1:

nandra:
@dellgreen updated.

nandra:
@kacf @dellgreen one more point I would like to highlight is tha tI would love to have some way to setup fstype and fstab options in variable. My first idea is something like:

```
MENDER_EXTRA_PARTS[part1] = "--fixed-size 1024M --label=static --fstype=ext4;auto;default,noauto"
```

if item is empty default value will be used. Or preferred is to create new variable names? Thanks.

dellgreen:
@nandra . Not advocating for this, but for reference the following was the kind of thing i was working on previously, which could easily and safely be converted to python dictionary with 'ast.literal_eval()', but the focus for that implementation was to have finer control over all the partition layouts as different people had different requirements as to how they wanted partitions laid 

> MENDER_CUSTOM_PARTITIONS = "{ \\
   1:{'fstype':'ext4', 'mvar':'MENDER_BOOT_PART'}, \\
   2:{'fstype':'ext4', 'mvar':'MENDER_ROOTFS_PART_A'}, \\
   3:{'fstype':'ext4', 'mvar':'MENDER_ROOTFS_PART_B'}, \\
   4:{'fstype':'swap', 'mvar':'MENDER_SWAP_PART', 'size':'4096'}, \\
   5:{'fstype':'ext4', 'mvar':'MENDER_DATA_PART'}, \\
}"

nandra:
Looks nice. I think current mender setup doesn't give user much options to setup partitions layout. I would keep it simple as is now just wan to extend for add user ability to not mount automatically. Maybe I can omit also fstype as auto can be used by default. I know it's not perfect solution I would see it as a base for other extensions ;). Thx

meta-mender-core/classes/mender-helpers.bbclass
```diff
@@ -242,12 +242,16 @@ def mender_get_data_part_num(d):
...
+    extra_parts = d.getVar('MENDER_EXTRA_PARTS')
     if (boot_part_size and boot_part_size != '0'): n += 1
     if (swap_part_size and swap_part_size != '0'): n += 1
+    if extra_parts: n += 1
 
     #is an msdos extended partion going to be required
     if n <= 4: return n
     if mender_is_msdos_ptable_image(d): n += 1
+    # extra partitions are added after data partition
+    if extra_parts: n -= 1
```
kacf: Is this correct? Even if the data partition has extra partitions following it, it would keep the same index, right?

meta-mender-core/classes/mender-part-images.bbclass
kacf: Actually I was talking about the comment, not the `${@get_extra_parts_wks(d)}` part below. The comment should be indented four spaces (outside here-document), the `${@get_extra_parts_wks(d)}` should not (inside here-document).

meta-mender-core/classes/mender-setup-image.inc
```diff
@@ -44,6 +44,18 @@ mender_update_fstab_file() {
 
     mkdir -p ${IMAGE_ROOTFS}/data
     printf "%-20s %-20s %-10s %-21s %-2s %s\n" ${tmpDataPart} /data ${MENDER_DATA_PART_FSTYPE} ${MENDER_DATA_PART_FSTAB_OPTS} 0 0 >> ${IMAGE_ROOTFS}${sysconfdir}/fstab
+
+    # TODO: skip if mender-partuuid is used
+    if [ -n "${MENDER_EXTRA_PARTS}" ]; then
```
kacf: Can you add a test for this and error out with `bbfatal` if both are set?

meta-mender-core/classes/mender-helpers.bbclass
nandra: Yes it is correct. In case we enable extra partition we need to add + 1 and in my case we have boot partition also it add +1 so we have 5. check if n <= 4 is false and then  mender_is_msdos_ptable_image add n+1 which is not right because then data will be mount on 6 and it should be on 5 (as 4 is extended partition). Point is that other partitions are added after data so thus -1.

meta-mender-core/classes/mender-setup-image.inc
nandra: OK will do. Thanks.

nandra: Also for comment alignment something like that:
```
diff --git a/meta-mender-core/classes/mender-part-images.bbclass b/meta-mender-core/classes/mender-part-images.bbclass
index 964b14f..ee10283 100644
--- a/meta-mender-core/classes/mender-part-images.bbclass
+++ b/meta-mender-core/classes/mender-part-images.bbclass
@@ -139,9 +139,9 @@ EOF
     cat >> "$wks" <<EOF
 part --source rawcopy --sourceparams="file=${IMGDEPLOYDIR}/${IMAGE_LINK_NAME}.dataimg" --ondisk "$ondisk_dev" --align $alignment_kb --fixed-size ${MENDER_DATA_PART_SIZE_MB} --mkfs-extraopts='${MENDER_DATA_PART_FSOPTS}' $part_type_params
 EOF
-# added extra partitions if exists
+    # added extra partitions if exists
     cat >> "$wks" <<EOF
-    ${@get_extra_parts_wks(d)}
+${@get_extra_parts_wks(d)}
 EOF
 
     cat >> "$wks" <<EOF

```

Thanks.

meta-mender-core/classes/mender-helpers.bbclass
dellgreen: I'n not convinced this is the correct way to go. mender_get_data_part_num() is dedicated to getting the data partition index and should not need to know nothing about  extra partitions if they always go after the data partition.  if extra partitions added, the data partition index shouldn't need to change. Later code for creating partitions should take result of mender_get_data_part_num() and calculate their indexes accordingly. :)

nandra: IMO it must be influenced because extra partitions maybe require to create extended partition. As you describe it I have it implemented and it was commented so I dropped it but certainly some action in this method (or other) must be done. If I drop it from mender_get_data_part_num without knowing about extra partitions then /data will be mounted on e.g. /dev/mmcblk0p4 which is not correct. So based on my testing this is correct solution :) (not saying it's elegant way). Let's think about proper one :+1:

dellgreen: Is it not broken if both boot_part_size and swap_part_size are not declared resulting in data part 4 being returned when it should be 3?

nandra: @dellgreen  yes this is right. I'll try to check it and fix for all configs.

nandra: ok created this : https://pastebin.com/hAYV49gG to verify fix (oneliner only) and it looks like now it should be OK. Any other comments to this function? I also realize that would need some helper because if board don't have boot part and also not swap and user define 2 extra partition code in updating fstab is not correct in this case.

meta-mender-core/classes/mender-setup-image.inc
kacf: Correct!

meta-mender-core/classes/mender-helpers.bbclass
nandra: @kacf looks good to me. One small nitpick which I see is:
```
    #is an msdos extended partion going to be required
-    if n <= 4: return n
-    if mender_is_msdos_ptable_image(d): n += 1
-    return n
+    if parts_num > 4 and mender_is_msdos_ptable_image(d):
+        parts_num += 1 <----- we don't need this line IMO
+        if data_pos >= 4:
+            data_pos += 1
+    return data_pos
```

kacf: Well spotted, feel free to remove it in your update.

dellgreen: Are you saying this line is not needed because its wrong, or because its redundant? 

if redundant then its probably worth keeping this line in for any code that may be added to this function in the future that may depend on an accurate parts_num calculation.

nandra: Line is redundant because variable is not used anywhere after that point. Function should be used as name says for return data offset. For partition number I'll probably create new helper which return number of partitions in total (used for fstab update)

nandra:
Squashed everything to one commit. Tested on my setup. Also verify helpers for computation return valid results in all combinations. Pls comment. Thanks.

nandra:
added missing changelog

kacf:
Resubmitted rebased version at #899 (checksum mismatch). Will close this if it passes.

kacf:
Merged!

kacf:
@nandra: Maybe you also want to add the variable to [this list](https://github.com/mendersoftware/mender-docs/blob/master/04.Artifacts/01.Yocto-project/99.Variables/docs.md), so people can find it.

nandra:
@kacf added : https://github.com/mendersoftware/mender-docs/pull/849 Thanks

MikelOrtega99:
@nandra @kacf 
are this extra partitions able to be used in updates? or are just for storing data etc
Thanks!

kacf:
Not using core Mender, no. But you can use one of the Update Modules to put files there, for example the [directory module](https://hub.mender.io/t/directory/325/54).

PR #892, opened 2020-01-07
mender-artifact: change to use OE go buildsystem

liuming50:
A build error was observed for mender-artifact-native, as follows:
| # github.com/mendersoftware/mender-artifact/cli/mender-artifact
| .../tmp/work/x86_64-linux/mender-artifact-native/3.2.1-r0/recipe-sysroot-native/usr/lib/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
| .../tmp/hosttools/ld: cannot find -llzma
| ...//tmp/hosttools/ld: cannot find -llzma

this issue is due to missing "-extldflags '${GO_EXTLDFLAGS}'" since
it's not passed to Makefile because do_compile from go.bbclass is not
used.

There is a comment saying using do_compile from go.bbclass would prevent
us from always building the most recent version due to recursive
expansion if SRCPV, but I could not see how that's happening, if that's
all about the "-X main.Version" string, then we could add a prefuncs
extract_main_version to handle that.

Also since the mender-artifact only contains a test go file, we need
create a dummy.go to let it build with go-1.2 and upper, see:
https://github.com/golang/go/issues/22409

Signed-off-by: Ming Liu <liu.ming50@gmail.com>

### DIFF
--- a/meta-mender-core/recipes-mender/mender-artifact/mender-artifact.inc
+++ b/meta-mender-core/recipes-mender/mender-artifact/mender-artifact.inc
@@ -4,24 +4,21 @@ GO_IMPORT = "github.com/mendersoftware/mender-artifact"
 inherit go
 
 S = "${WORKDIR}/git"
-B = "${WORKDIR}/build"
 
-GOPATHDIR = "${B}/src/${GO_IMPORT}"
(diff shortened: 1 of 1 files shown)


kacf:
I'm not exactly sure what the comment is referring to, but regardless of that, why not put this in the Makefile instead? That should be much cleaner, and a smaller change.

oleorhagen:
Closed due to inactivity

PR #896, opened 2020-01-14
Enable the new formatting check for Python files

oleorhagen:
This adds the new formatting check, and formats the files which are out of sync.


oleorhagen:
Cancelled the pipeline, as it is not applicable to this change

oleorhagen:
@kacf that means you have to force-merge it

PR #897, opened 2020-01-16
tests for snapshot feature + improvements

kacf:


kacf:
Cancelling tests for now. I know this won't pass because of some issues in the client.

PR #909, opened 2020-01-30
Added extra parts fstab options

nandra:
This PR adds handling for user to define fstab options like:
```
MENDER_EXTRA_PARTS_FSTAB = "part1 part2"
MENDER_EXTRA_PARTS_FSTAB[part1] = "auto default,ro"
MENDER_EXTRA_PARTS_FSTAB[part2] = "auto 123,4"
```

part1 and part2 should match ones in ```MENDER_EXTRA_PARTS``` and user can define only selected (other ones are defaults). Pls comment.


meta-mender-core/classes/mender-helpers.bbclass
```diff
@@ -350,19 +357,50 @@ def get_extra_parts_by_id(d, id):
...
+def get_extra_parts_fstab(d):
+    out = []
+    extra_parts_offset = mender_get_extra_parts_offset(d)
+    device = d.getVar('MENDER_STORAGE_DEVICE_BASE')
+    for part in get_extra_parts_flags(d):
+       label = get_extra_parts_label(d, part)
+       fstype_opts = get_extra_parts_fstab_opts(d, part).split()
+       if len(fstype_opts) < 2:
+            bb.warn("MENDER_EXTRA_PARTS_FSTAB[%s] invalid." % part)
+            continue
```
kacf: I think this should either be a failure (`bbfatal`) or a graceful fallback to using `auto default`. Personally I think I vote for the former. This could perhaps be a reason for moving that logic into `get_extra_parts_fstab_opts` instead of here.

What will happen now is that the build will continue, but you will be missing a whole partition. Definitely not what you want.

nandra: @kacf ok thanks a lot for comment. I'll update in next few days :+1:

nandra:
@kacf updated. I've moved handling and added bbfatal. When I should squash last commit to previous one just tell and I'll do it. Thanks.

meta-mender-core/classes/mender-helpers.bbclass
```diff
@@ -350,19 +357,49 @@ def get_extra_parts_by_id(d, id):
...
+            labels.append(get_extra_parts_label(d, id))
     return ' '.join(labels)
 
+def get_extra_parts_label(d, id = None):
+    part = get_extra_parts_by_id(d, id)
+    if part:
+        s = part.find("--label")
+        if s > 0:
+            e = part.find("--", s + 2)
+            return part[s + len("--label="):e]
```
kacf: Upon trying this I realized that this doesn't work if the `--label` argument is the last on the line. The indexed searches are also a bit hard to follow, so I suggest changing it to something that works in both cases and is easier to read, like this:

```diff
diff --git a/meta-mender-core/classes/mender-helpers.bbclass b/meta-mender-core/classes/mender-helpers.bbclass
index 3100fc55..3f75f2f3 100644
--- a/meta-mender-core/classes/mender-helpers.bbclass
+++ b/meta-mender-core/classes/mender-helpers.bbclass
@@ -366,10 +366,9 @@ def get_extra_parts_labels(d):
 def get_extra_parts_label(d, id = None):
     part = get_extra_parts_by_id(d, id)
     if part:
-        s = part.find("--label")
-        if s > 0:
-            e = part.find("--", s + 2)
-            return part[s + len("--label="):e]
+        import re
+        match = re.search(r'--label=(\S+)', part)
+        return match.group(1) if match else ""
     return ""
 
 def get_extra_parts_wks(d):
```

The overhead of using a regex here is negligible anyway.

nandra: OK fine I'll squash and update

meta-mender-core/classes/mender-helpers.bbclass
```diff
@@ -350,19 +357,36 @@ def get_extra_parts_by_id(d, id):
...
             final_parts.append("part --ondisk \"$ondisk_dev\" --align \"$alignment_kb\" {}".format(part))
     return '\n'.join(final_parts)
+
+def get_extra_parts_fstab(d):
+    out = []
+    extra_parts_offset = mender_get_extra_parts_offset(d)
+    device = d.getVar('MENDER_STORAGE_DEVICE_BASE')
+    for part in get_extra_parts_flags(d):
+       label = get_extra_parts_label(d, part)
+       out.append("{:20s} {:20s} {:10s} {:21s} 0 0".format(device + str(extra_parts_offset), "/mnt/{}".format(label), "auto", "default"))
```
nandra: @kacf this is also not necessary to align as same is done in fstab postprocess callback and I tested it and work fine. Will drop this in final version something like:
```
--- a/meta-mender-core/classes/mender-helpers.bbclass
+++ b/meta-mender-core/classes/mender-helpers.bbclass
@@ -399,7 +399,7 @@ def get_extra_parts_fstab(d):
     for part in get_extra_parts_flags(d):
        label = get_extra_parts_label(d, part)
        fstype_opts = get_extra_parts_fstab_opts(d, part).split()
-       out.append("{:20s} {:20s} {:10s} {:21s} 0 0".format(device + str(extra_parts_offset), "/mnt/{}".format(label), fstype_opts[0], fstype_opts[1]))
+       out.append("{} {} {} {} 0 0".format(device + str(extra_parts_offset), "/mnt/{}".format(label), fstype_opts[0], fstype_opts[1]))
        extra_parts_offset += 1

```

kacf: Ok!

nandra:
@kacf squashed and updated according comments. Thanks.

nandra:
@kacf thanks. Yes when it will be merged I'll update documentation (also will do for partuuid as this nice feature is missing in documentation also ;)

kacf:
Thanks @nandra!

PR #912, opened 2020-02-10
In demo mode, put demo certificate in same directory as Debian package.

kacf:
This fixes an issue with `mender setup`.

Changelog: Title

MEN-3048

### DIFF
--- a/meta-mender-core/recipes-mender/mender/mender.inc
+++ b/meta-mender-core/recipes-mender/mender/mender.inc
@@ -10,7 +10,7 @@ def cert_location_if_server_crt_in(src_uri, d):
     return ""
 
 MENDER_SERVER_URL ?= "https://docker.mender.io"
-MENDER_CERT_LOCATION ?= "${@cert_location_if_server_crt_in('${SRC_URI}', d)}"
+MENDER_CERT_LOCATION ??= "${@cert_location_if_server_crt_in('${SRC_URI}', d)}"
 # Tenant token
 MENDER_TENANT_TOKEN ?= "dummy"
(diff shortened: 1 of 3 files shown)


meta-mender-core/recipes-mender/mender/mender.inc
```diff
@@ -10,7 +10,7 @@ def cert_location_if_server_crt_in(src_uri, d):
     return ""
 
 MENDER_SERVER_URL ?= "https://docker.mender.io"
-MENDER_CERT_LOCATION ?= "${@cert_location_if_server_crt_in('${SRC_URI}', d)}"
+MENDER_CERT_LOCATION ??= "${@cert_location_if_server_crt_in('${SRC_URI}', d)}"
```
lluiscampos: For my own education, but why do we change it to weak assignment?

meta-mender-core/recipes-mender/mender/mender.inc
```diff
@@ -10,7 +10,7 @@ def cert_location_if_server_crt_in(src_uri, d):
     return ""
```
lluiscampos: I was looking back at forth at this function. But I agree that we shall not modify this, so that a production cert is still installed in `/etc/mender/server.crt` :+1:

kacf:
> Have you manually run `mender setup` after a build and verified that it just "works"?

I did, everything works!

meta-mender-core/recipes-mender/mender/mender.inc
kacf: Because of the other weak assignment I introduced in the same commit. We want it to override this one, but we don't want it to override the users' assignments using `=`.

PR #914, opened 2020-02-10
Secure boot

drewmoseley:


drewmoseley:
@kacf since the grub-mender-grubenv changes here are specific to Yocto, I opted to keep them in a patch applied here. Let me know if you would prefer that to go directly into the upstream repo instead.

meta-mender-core/recipes-bsp/grub-mender-grubenv/grub-mender-grubenv.inc
```diff
@@ -93,3 +94,19 @@ do_install() {
...
+}
+do_sign[prefuncs] += "${@bb.utils.contains('DISTRO_FEATURES', 'efi-secure-boot', 'check_deploy_keys', '', d)}"
+do_sign[prefuncs] += "${@'check_boot_public_key' if d.getVar('GRUB_SIGN_VERIFY', True) == '1' else ''}"
+
+fakeroot do_chownboot() {
+    if "${@bb.utils.contains('DISTRO_FEATURES', 'efi-secure-boot', 'true', 'false', d)}"; then
+        chown root:root -R "${D}${GRUB_CONF_LOCATION}/grub.cfg${SB_FILE_EXT}"
+    fi
+}
+addtask chownboot after do_sign before do_package
```
kacf: I believe you can simplify this considerably by doing this instead:

```python
python() {
    if bb.utils.contains('DISTRO_FEATURES', 'efi-secure-boot', True, False, d):
        bb.build.addtask('do_sign', 'do_deploy do_package', 'do_install', d)
        bb.build.addtask('do_chownboot', 'do_package', 'do_sign', d)
}
```

This conditional adding means that you can get rid of all the `efi-secure-boot` checks elsewhere.

meta-mender-core/recipes-bsp/grub-mender-grubenv/files/01_efi_secure_boot_grub.cfg
```diff
@@ -0,0 +1,16 @@
...
+if [ "${secured}" = "1" ]; then
+    if [ -s "${prefix}/password.inc" ]; then
+        source "${prefix}/password.inc"
+    fi
+fi
+
+get_efivar -f uint8 -s unprovisioned SetupMode
+
+if [ "${unprovisioned}" = "1" ]; then
+    chainloader "${prefix}/LockDown.efi"
```
kacf: Could you give a bit more background about the `password.inc` and `LockDown.efi` files? Could really use some comments here.

drewmoseley:
nce the grub-mender-grubenv changes here are specific to Yocto, I opted to keep them in a patch applied here. Let me know if you would prefer that to go directly into the upstream repo instead.
> 
> I'm ok with that, but can you elaborate on exactly what is specific to Yocto?

The entire infrastructure needed to setup the keys, build an appropriate LockDown.efi, etc is part of the meta-secure-core layer.  I doubt there will be much overlap if/when we implement this for a non-Yocto system.  I suppose if that assertion turns out to be incorrect, we can always refactor and move the changes into grub-mender-grubenv at that time.

meta-mender-core/recipes-bsp/grub-mender-grubenv/grub-mender-grubenv.inc
drewmoseley: Fixed in the latest push.

meta-mender-core/recipes-bsp/grub-mender-grubenv/files/01_efi_secure_boot_grub.cfg
drewmoseley: Fixed in the latest push

drewmoseley:
@kacf  I also squashed the two commits into one since there is no value in them being separate and both are required for it to work anyway.

PR #921, opened 2020-02-18
rpi: Add workaround for display corruption in Warrior branch.

drewmoseley:
Changelog: title


mirzak:
Would it be more appropriate to put this patch in meta-raspberrypi (if U-Boot is enabled)? Nothing Mender specific here as I see it?

drewmoseley:
I already submitted the kernel fix there.  This is just a workaround since we require u-boot.

mirzak:
> I already submitted the kernel fix there. This is just a workaround since we require u-boot.

Yes, but would it not be useful for other users of meta-raspberrypi? You can enable U-Boot without Mender as well and hence why I am pushing to put it there.

drewmoseley:
Once the kernel fix is accepted there this will not be needed.  I'm suggesting it here as a workaround since we control when the fix goes in here but not in meta-raspberrypi.

mirzak:
I misunderstood, I though that both changes where needed. Then I do not have any further comments :)

drewmoseley:
Thanks.  And sorry for not clarifying earlier.

kacf:
Alright, merging then, since this is only for warrior, master is still prepared to use the upstream fix.

drewmoseley:
Right.  In fact I just got the notification that my PR was merged.  I'll verify that it works and then revert this if possible.

PR #922, opened 2020-02-18
MENDER_EXTRA_PARTS feature

kacf:
Resubmission of #909, with test.


meta-mender-core/classes/mender-helpers.bbclass
```diff
@@ -235,7 +235,7 @@ mender_get_clean_kernel_devicetree() {
 
 def mender_is_msdos_ptable_image(d):
     mptimgs = 'mender-image-sd mender-image-bios'
-    return bb.utils.contains_any('MENDER_FEATURES_ENABLE', mptimgs , True, False, d)
+    return bb.utils.contains_any('DISTRO_FEATURES', mptimgs , True, False, d)
```
kacf: @nandra: FYI.

nandra: @kacf oh my sorry about that and thanks for spotting that

kacf:
Unrelated error, merging.

PR #923, opened 2020-02-19
Adjust licenses file checksum after vendor update in mender

lluiscampos:
Together with https://github.com/mendersoftware/mender/pull/499


kacf:
For simplicity, maybe you should just change the license for 2.2.x as well, and test both master and 2.2.x client and then merge them at the same time? Because you are going to put this in 2.2.x as well, right?

lluiscampos:
> For simplicity, maybe you should just change the license for 2.2.x as well, and test both master and 2.2.x client and then merge them at the same time? Because you are going to put this in 2.2.x as well, right?

Right!

This is what I wanted to discuss in the stand-up, but I didn't manage to formulate the question properly. I will do as you suggest, thanks.

PR #925, opened 2020-02-21
Warrior backports

drewmoseley:


drewmoseley:
Yes I need these features as well as the bug fixes.

kacf:
Ok. I think most are relatively small and isolated, with the exception of "mender-part-images: Added handling for extra partitions". This one I'm *very* reluctant to add. It has already caused multiple problems and followup fixes (579661b4007a850a47e35c931bfd9c7f5139893f and ea6f936ff38487f8009ebcdb565d908a7de69155), so I'm not convinced of its stability.

As I have spent considerable time lately getting zeus into better shape. If we release that soon, would it be usable for you, or is warrior the only option?

drewmoseley:
Zeus may be an option but I definitely need the multiple partition support sooner than later on this branch.  Note that I am reworking it significantly to make it more expandable and will hopefully have something to review soon.  I guess we can hold off on this one until I get that rework up for review.

kacf:
Ok, this sounds good.

Now that zeus has finally received some development time, hopefully we can keep this out of warrior as it is getting quite old. Zeus will be branched from master, so the multi-partition support is already there, and you have some time to prepare your rework as well. I'll make sure to coordinate with you before I publish any official zeus branch.

drewmoseley:
@kacf I removed the MENDER_EXTRA_PARTS stuff from this PR.  For my needs on warrior I'm going to just add new partitions directly in mender-partimage.bbclass, with similar config parameters to the existing parameters. It's not the full generic solution but should work for my needs at the moment.  I'll post them for review once they are ready.  In the meantime if we can get _this_ PR merged that would be helpful.

kacf:
Alright, that should be doable. You still have some commit message issues in your PR, and need to rebase, but it's just mechanical stuff, so I saved you the turnaround time: #927. :-)

drewmoseley:
Thanks.  Closing in deference to #927

PR #926, opened 2020-02-26
Improve dependency logic to not require U-Boot unconditionally on ARM.

kacf:
This will help when not using `mender-grub`, and instead using
Barebox, for example.

Changelog: Commit


kacf:
This was prompted by [this discussion](https://hub.mender.io/t/barebox-support/531/4).

PR #928, opened 2020-02-27
mender: Add support for additional partitions to hold kernel binaries.

drewmoseley:
Changelog: Title


drewmoseley:
@kacf thoughts on this? It's obviously hard coded to allow only 2 extra partitions for the kernel binaries but at least the sizes and mount options are as configurable as the other partitions.  This will give me what I need for the dm-verity implementation on warrior while waiting until Zeus for the full MENDER_EXTRA_PARTS support.

Ideally we can have this merged into warrior rather than keeping it on a customer-specific branch.

kacf:
I think I may have misunderstood [your earlier comment](https://github.com/mendersoftware/meta-mender/pull/925#issuecomment-590871494). I thought you meant that you were reworking it for master?

Codewise this looks good, but I think we should not merge this into warrior, for several reasons (in addition to the stable branch policy):

1. It's only required by one customer.
2. It introduces a way to define partitions that we will drop again in zeus (unless I've misunderstood?)

I would have no qualms about delivering this to the customer, but it should come with a "this will change in zeus" label attached to it, and I think from a planning perspective it's better if we don't expose this to everybody.

Does that make sense, or am I being unreasonable here?

drewmoseley:
That was my plan. ie have this as a one-off in warrior to be replaced with the generic stuff in zeus.  But I have since thought that maybe we don't need this at all. I think we can just store the signed kernels in /data and bypass all this.

PR #930, opened 2020-03-02
Adjust licenses file checksum after vendor update in mender

lluiscampos:
Cherry picking #923


kacf:
Already merged here: https://github.com/mendersoftware/meta-mender/pull/927

lluiscampos:
Thanks. It was blocking the release that is why I launched an independent PR.

But it should be fine now. Rebuilding :)

kacf:
Yep, it was just good timing! ;-) I would not have waited either if there was a problem.

PR #933, opened 2020-03-02
Dynamic grub device

drewmoseley:
This allows the grub device to be determined dynamically.
This fixes cases where the EFI renumbers the devices, normally when a USB drive is connected to the system. Not all BIOSes do this but this patch bases everything on whatever grub itself is booted from so this should work for all cases.


meta-mender-core/classes/mender-setup-grub.inc
```diff
@@ -1,11 +1,5 @@
 EXTRA_IMAGEDEPENDS_append_mender-grub = " grub-efi"
 
-# The device number of the interface to load partitions from. This is normally
-# empty, in which case it is deduced from MENDER_STORAGE_DEVICE. Only use this
-# if the indexing of devices is different in Grub and in the Linux kernel.
-MENDER_GRUB_STORAGE_DEVICE ??= "${MENDER_GRUB_STORAGE_DEVICE_DEFAULT}"
```
kacf: Note that this is a [documented variable](https://github.com/mendersoftware/mender-docs/search?q=MENDER_GRUB_STORAGE_DEVICE&unscoped_q=MENDER_GRUB_STORAGE_DEVICE), so it should probably be marked as deprecated.

I would also, in addition to your existing `Changelog: Title`, add:

```
Changelog: Deprecate and remove `MENDER_GRUB_STORAGE_DEVICE` variable.
```

Just to be explicit about it in the changelog (yes, you can have two changelog entries in one commit!).

meta-mender-core/recipes-bsp/grub-mender-grubenv/grub-mender-grubenv.inc
```diff
@@ -52,20 +52,12 @@ do_configure() {
...
-        mender_grub_storage_device=$(get_grub_device_from_device_base ${MENDER_STORAGE_DEVICE_BASE})
-    fi
-
     cat > ${B}/mender_grubenv_defines <<EOF
 mender_rootfsa_part=$mender_rootfsa_part
 mender_rootfsb_part=$mender_rootfsb_part
 mender_kernel_root_base=${MENDER_STORAGE_DEVICE_BASE}
-mender_grub_storage_device=$mender_grub_storage_device
 kernel_imagetype=${KERNEL_IMAGETYPE}
+regexp (.*),(.*) \$root -s mender_grub_storage_device
```
kacf: I love this patch!!

But do you think it could go as an extra file in [grub-mender-grubenv](https://github.com/mendersoftware/grub-mender-grubenv) instead, say `04_grub_storage_device`? Then it would also benefit mender-convert.

Testing them together with the rest of this patch is a bit of a PITA, but I can help with that.

drewmoseley: The only concern I have is can we guarantee that the regexp module is available there? I explicitly had to turn it on in Yocto so if we are using distro-provided Grub binaries we should probably take a look to see if regexp is reasonably standard.

kacf: Good point. I have previously argued that the list of modules should live in grub-mender-grubenv, but we have not had the time to make this happen, so it's duplicated in meta-mender and in mender-convert-integration-scripts ATM. If you could fix *that* I would of course be super happy, but then I'm fine with keeping it here also for the time being! :-)

meta-mender-core/classes/mender-setup-grub.inc
drewmoseley: Updated changelog on its was as well as a mender-docs PR.

drewmoseley: Docs: https://github.com/mendersoftware/mender-docs/pull/894

meta-mender-core/recipes-bsp/grub-mender-grubenv/grub-mender-grubenv.inc
drewmoseley: This goes in with https://github.com/mendersoftware/mender-docs/pull/894

drewmoseley:
Closing in deference to #937

kacf:
Unfortunately it looks like this breaks BIOS support. I'm not sure about the exact reason, but my suspicion is that `$root` is either not set, or set to something unexpected, because GRUB loads from a first stage boot loader in the MBR.

Try using `MENDER_FEATURES_ENABLE_append = " mender-bios"` and see where you get. Alternatively, if `$root` is unusable, we may have to keep the old code around for that special case.

drewmoseley:
@kacf I'm not sure how to run a mender-bios image.  I added that setting but with or without my patch, I cannot figure out how to properly launch the system.  What target are you running to test this and how?

kacf:
Actually, looks like what you should use is `INHERIT += "mender-full-bios"` in `local.conf`. This creates a `biosimg` compatible with traditional BIOS loaders. Many x86 boards still allow you to boot this, but you may have to turn it on in the Boot Setup.

However, it's also possible that your board/laptop/whatever doesn't have this mode anymore. If so you can use `MACHINE = "qemux86-64"` and launch it using `meta-mender-qemu/scripts/mender-qemu`.

PR #934, opened 2020-03-03
MEN-2948: Rename mender -> mender-client

oleorhagen:
Hi, I would like some comments on the direction here, as I am a little bit torn.

I cannot decide if we should rename more than the recipes themselves to mender-client.
Should everything now have the `mender-client` prefix? IMO this sounds better, and is more descriptive.

Or should simply the recipe, and the sytemd service file be renamed? (Which was the original description in the ticket).

I'm personally in favour of using `mender-client` for everything. So that a user would now:
```bash
INHERIT+="mender-client-full"
```

Right now it is somewhat in between (in this PR). And I will leave it like this until we have discussed this.

Ticket: https://tracker.mender.io/browse/MEN-2984

-O


drewmoseley:
I'm all for consistency so if we do this, then I recommend doing it everywhere.

meta-mender-demo/conf/layer.conf
```diff
@@ -17,4 +17,4 @@ IMAGE_OVERHEAD_FACTOR = "1.0"
 IMAGE_FEATURES += "splash"
 
 LAYERSERIES_COMPAT_mender-demo = "warrior"
-LAYERDEPENDS_mender-demo = "mender"
+LAYERDEPENDS_mender-demo = "mender-client"
```
kacf: This "mender" reference is for the layer as a whole, including mender-artifact, so this should definitely be just "mender".

Same goes for the other `LAYERDEPENDS` changes.

meta-mender-core/classes/mender-full.bbclass
```diff
@@ -12,7 +12,7 @@ MENDER_FEATURES_ENABLE_append = " \
 _MENDER_IMAGE_TYPE_DEFAULT ?= "mender-image-uefi"
 _MENDER_BOOTLOADER_DEFAULT ?= "mender-grub"
 
-_MENDER_GROWFS_DATA_DEFAULT ?= "${@'' if d.getVar('MENDER_EXTRA_PARTS') else 'mender-growfs-data'}"
+_MENDER_GROWFS_DATA_DEFAULT ?= "${@'' if d.getVar('MENDER_EXTRA_PARTS') else 'mender-client-growfs-data'}"
```
kacf: You don't need the client to grow the filesystem, so I'd argue "mender" here as well.

oleorhagen:
@drewmoseley 

We have discussed this in the office, the three of us, and decided to (mostly stick with the Mender) prefix naming. 

Reason being that most of the features are -- even though required by the client -- mostly not client specific, like u-boot never touches the client. It is but a client dependency. 

However, the `mender-install` feature, will now be renamed to `mender-client-install`.

If this is completely inappropriate, please speak out.

-O

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -202,7 +202,7 @@ MENDER_PERSISTENT_CONFIGURATION_VARS ?= "RootfsPartA RootfsPartB"
 
 # --------------------------- END OF CONFIGURATION -----------------------------
 
-IMAGE_INSTALL_append = " mender"
+IMAGE_INSTALL_append = " mender-client"
```
oleorhagen: Reviwers Note: Should we, or should we not rename this?

meta-mender-core/conf/layer.conf
```diff
@@ -11,11 +11,11 @@ LAYERDIR_MENDER := "${LAYERDIR}"
 BBFILES += "${LAYERDIR}/recipes-*/*/*.bb \
 	${LAYERDIR}/recipes-*/*/*.bbappend"
 
-BBFILE_COLLECTIONS += "mender"
-BBFILE_PATTERN_mender = "^${LAYERDIR}/"
-BBFILE_PRIORITY_mender = "6"
+BBFILE_COLLECTIONS += "mender-client"
```
oleorhagen: Reviewers Note: Also, should this be renamed or not?

meta-mender-core/recipes-mender/mender-client/mender-client.inc
```diff
@@ -25,27 +25,42 @@ inherit go
...
 
-SYSTEMD_SERVICE_${PN} = "mender.service"
-FILES_${PN} += "${systemd_unitdir}/system/mender.service \
+SYSTEMD_SERVICE_${PN} = "mender-client.service"
+FILES_${PN} += "${systemd_unitdir}/system/mender-client.service \
                 ${sysconfdir}/mender.conf \
                 /data/mender/device_type \
                 /data/mender/mender.conf \
                 ${sysconfdir}/udev/mount.blacklist.d/mender \
+                /usr/share/mender/identity \
```
oleorhagen: Reviewers Note: I have no idea why these were not needed previously...

meta-mender-core/recipes-mender/mender-client/mender-client.inc
```diff
@@ -25,27 +25,42 @@ inherit go
...
                "
 
 SYSROOT_DIRS += "/data"
 
-SRC_URI_append_mender-image_mender-systemd = " file://mender-data-dir.service"
-SRC_URI_append_mender-persist-systemd-machine-id = " file://mender-systemd-machine-id.service \
-                                                     file://mender-set-systemd-machine-id.sh"
-FILES_${PN}_append_mender-image_mender-systemd = " ${systemd_unitdir}/system/mender-data-dir.service \
-                                                   ${systemd_unitdir}/system/mender.service.wants/mender-data-dir.service"
+SRC_URI_append_mender-image_mender-systemd = " file://mender-client-data-dir.service"
```
oleorhagen: Reviewers Note: Will clean up.

meta-mender-demo/conf/layer.conf
oleorhagen: I guess that answers the above. Changing.

oleorhagen:
Relates to:
Mender-dist-packages: https://github.com/mendersoftware/mender-dist-packages/pull/27
Mender: https://github.com/mendersoftware/mender/pull/503

meta-mender-core/recipes-mender/mender-client/mender-client.inc
oleorhagen: Or whether or not they are all needed now...

oleorhagen:
Build along with https://github.com/mendersoftware/mender/pull/503

https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/123546192

meta-mender-core/conf/layer.conf
kacf: No, this should definitely be "mender", this refers to the name of the layer, not the client.

meta-mender-core/recipes-mender/mender-client/mender-client.inc
kacf: It has become stricter in zeus. I had to add several of these as well.

kacf: Btw, I think you can just add `/usr/share/mender/*`.

oleorhagen:
@kacf Green at last!

Mind having another look?

meta-mender-core/recipes-mender/mender-client/mender-client.inc
oleorhagen: Should we not keep them explicit?

meta-mender-core/recipes-mender/mender-client/mender-client.inc
```diff
@@ -25,27 +25,42 @@ inherit go
 
 inherit systemd
 
-SYSTEMD_SERVICE_${PN} = "mender.service"
-FILES_${PN} += "${systemd_unitdir}/system/mender.service \
+SYSTEMD_SERVICE_${PN} = "${MENDER_CLIENT}.service"
```
oleorhagen: Question: I'm not sure if `MENDER_CLIENT` is the best name. Other names I can think of grow pretty long though. Like `MENDER_CLIENT_RECIPE_NAME`? Soo... Hmmm

oleorhagen:
Also, @kacf. Could adding a `systemd` `Alias=mender-client` to the older service files be an idea?

meta-mender-core/recipes-mender/mender-client/mender-client.inc
kacf: `MENDER_CLIENT` sounds good to me.

kacf:
> Also, @kacf. Could adding a `systemd` `Alias=mender-client` to the older service files be an idea?

Not sure if it's worth the trouble. If they are on a branch which has this change, then they will also have the version which has the actual fix. So the only reason they would be using the old one is if they actively selected it over the new one. In which case I think we might as well just leave it alone.

meta-mender-core/recipes-mender/mender-client/mender-client.inc
kacf: It might be better for visibility and catching errors, and there aren't a huge number, so sure, it makes sense to keep them explicit.

PR #940, opened 2020-03-05
vendor mender-artifact: update md5sum

alfrunes:
changelog: none


meta-mender-core/recipes-mender/mender-artifact/mender-artifact_git.bb
```diff
@@ -61,7 +61,7 @@ def mender_license(branch):
         }
     else:
         return {
-                   "md5": "99143e34cf23a99976a299da9fa93bcf",
+                   "md5": "9e55056dd32c74785680c4121ff00428",
```
kacf: Could you add a `3.3.x` branch here with the old checksum? It's going to break for that branch otherwise.

PR #943, opened 2020-03-10
Move the leftover recipes to mender-client

oleorhagen:
A few files were merged while the PR for renaming the recipe files was in
flight. This PR renames the two remaining recipes.

Changelog: None


oleorhagen:
I guess the beta recipe should be removed also?

lluiscampos:
> Ah, a classic case of two PRs working separately, but not when combined. I haven't seen too many of those in my time here though.

I have seen a few...

PR #946, opened 2020-03-13
MEN-3263: Explicitly add liblzma to RDEPENDS

oleorhagen:
From the Yocto Manual changes to zeus:

``` The file-rdeps runtime dependency check no longer expands RDEPENDS
recursively as there is no mechanism to ensure they can be fully computed, and
thus races sometimes result in errors either showing up or not. Thus, you might
now see errors for missing runtime dependencies that were previously satisfied
recursively. Here is an example: package A contains a shell script starting with
B, which does depend on bash. You need to add the missing dependency or
dependencies to resolve the warning.
```

Which is what did bite us this time, as the `xz` dependency did resolve it
recursively previously.

Changelog: None


oleorhagen:
Builds fine locally

PR #949, opened 2020-03-16
Amend test_too_big_image_update to accept both stdout/stderr

lluiscampos:
Commit cd5c489 made this test strictly not compatible with older mender
releases (where the logger still used stdout).

Furthermore, the commit introduced an error in the second assert, as the
"ret_code=0" will not be seen in stderr in any case.

See #942

### DIFF
--- a/tests/acceptance/test_update.py
+++ b/tests/acceptance/test_update.py
@@ -267,10 +267,15 @@ class TestUpdates:
             output = connection.run(
                 "mender %s /var/tmp/image-too-big.mender ; echo 'ret_code=$?'"
                 % install_flag
-            ).stderr
+            )
 
-            assert "no space left on device" in output, output
-            assert "ret_code=0" not in output, output
+            assert any(
+                [
+                    "no space left on device" in out
+                    for out in [output.stderr, output.stdout]
+                ]
+            ), output
+            assert "ret_code=0" not in output.stdout, output
 
         finally:
             # Cleanup.


oleorhagen:
Hmm, why not add the output from stdout also, and then do an or?

lluiscampos:
I don't think accepting both is a good idea, as that will make any future change in the logger go unnoticed. The correct way of doing it would be to get from the test framework with version of mender are we testing, and from there expect the string in `stdout` or `stderr`. As I didn't know how to do this (I didn't investigate much, neither...) I thought that disabling it would be okey.

If you think that this is an important test for old releases, I will then investigate a bit more from where to get the mender version under test.

oleorhagen:
> I don't think accepting both is a good idea, as that will make any future change in the logger go unnoticed. The correct way of doing it would be to get from the test framework with version of mender are we testing, and from there expect the string in `stdout` or `stderr`. As I didn't know how to do this (I didn't investigate much, neither...) I thought that disabling it would be okey.
> 
> If you think that this is an important test for old releases, I will then investigate a bit more from where to get the mender version under test.

It's not super important for me, it just makes me a little unease, as this is the only test we have (I think), which does exactly this. So maybe we should have it run proper.

kacf:
I think I would consider an 'or' here. I can't see any bad consequence of allowing both, they're both equally valid, just different quirks of the logger.

lluiscampos:
> I think I would consider an 'or' here. I can't see any bad consequence of allowing both, they're both equally valid, just different quirks of the logger.

All right! Thanks for shipping in your opinion.

I will then implement the or'ing that both of you are suggesting :)

lluiscampos:
Reworked :+1: 

@oleorhagen  See also commit message regarding the second assert; I believe this was bogus since the stdout to stderr.

tests/acceptance/test_update.py
```diff
@@ -267,10 +267,15 @@ def test_too_big_image_update(self, bitbake_variables, connection):
...
                 % install_flag
-            ).stderr
+            )
 
-            assert "no space left on device" in output, output
-            assert "ret_code=0" not in output, output
+            assert any(
+                [
+                    "no space left on device" in out
+                    for out in [output.sterr, output.stout]
```
oleorhagen: btw, should it not be `stderr` and `stdout` ?

lluiscampos: :man_facepalming:

lluiscampos: Pushed a fix. Thanks!

PR #950, opened 2020-03-16
Updated the license checksum in the client recipe

oleorhagen:
This goes along with the update to the logrus dependency in the client.

Changelog: None


oleorhagen:
Already updated

PR #952, opened 2020-03-20
[Final] Add coverage instrumentation to the Mender client

oleorhagen:
This PR adds the ability to collect coverage through running the mender client like usual.

This PR is now ready for final review.

:slightly_smiling_face:

### DIFF
--- a/tests/acceptance/common.py
+++ b/tests/acceptance/common.py
@@ -37,12 +37,12 @@ configuration = {}
 
 
 def start_qemu(qenv=None, conn=None):
-    """Start qemu and return a subprocess.Popen object corresponding to a running
+    """Start QEMU and return a subprocess.Popen object corresponding to a running
     qemu process. `qenv` is a dict of environment variables that will be added
     to `subprocess.Popen(..,env=)`.
 
-    Once qemu is stated, a connection over ssh will attempted, so the returned
-    process is actually a qemu instance with fully booted guest os.
+    Once qemu is started, a connection over ssh will attempted, so the returned
+    process is actually a QEMU instance with a fully booted guest OS.
 
     The helper uses `meta-mender-qemu/scripts/mender-qemu` to start qemu, thus
     you can use `VEXPRESS_IMG`, `QEMU_DRIVE` and other environment variables to
@@ -255,7 +255,7 @@ def manual_uboot_commit(conn):
 
 def common_board_setup(conn, files=None, remote_path="/tmp", image_file=None):
     """
-    Deploy and activate an image to a board that usese mender-qa tools.
+    Deploy and activate an image to a board that uses mender-qa tools.
 
     :param image_file: IMAGE_FILE as passed to deploy-test-image, can be None
     :param remote_path: where files will be stored in the remote system
--- a/tests/acceptance/fixtures.py
+++ b/tests/acceptance/fixtures.py
@@ -131,6 +131,12 @@ def setup_qemu(request, build_dir, conn):
         def qemu_finalizer_impl(conn):
             try:
                 manual_uboot_commit(conn)
+                # Collect the coverage files from /data/mender/ if present
+                try:
+                    conn.run("ls /data/mender/cover*")
+                    get_no_sftp("/data/mender/cover*", conn, local="coverage")
+                except:
+                    pass
                 conn.run("poweroff")
                 halt_time = time.time()
                 # Wait up to 30 seconds for shutdown.
--- a/tests/meta-mender-ci/recipes-testing/mender-coverage-client/gobinarycoverage.bb
+++ b/tests/meta-mender-ci/recipes-testing/mender-coverage-client/gobinarycoverage.bb
@@ -1,0 +1,11 @@
+DESCRIPTION = "Mender image artifact library"
+GO_IMPORT = "github.com/mendersoftware/gobinarycoverage"
+
+inherit go
+
+SRC_URI = "git://${GO_IMPORT};protocol=git"
+SRCREV = "${AUTOREV}"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://src/${GO_IMPORT}/LICENSE;md5=86d3f3a95c324c9479bd8986968f4327"
+
+BBCLASSEXTEND = "native nativesdk"
--- a/tests/meta-mender-ci/recipes-testing/mender-coverage-client/mender-client_%.bbappend
+++ b/tests/meta-mender-ci/recipes-testing/mender-coverage-client/mender-client_%.bbappend
@@ -1,0 +1,60 @@
+SUMMARY = "Build a Mender client which records coverage during its invocation"
+
+DEPENDS += "gobinarycoverage-native rsync-native"
+
+do_instrument_client[dirs] =+ "${GOTMPDIR}"
+do_instrument_client[doc] = "Modifies the Mender client source to enable coverage analysis"
+do_instrument_client () {
+    cd ${B}/src/${GO_IMPORT}
+    bbwarn "Building the Mender client with coverage analysis"
+    oe_runmake instrument-binary
+}
+
+do_configure_prepend () {
+    # Remove all the src present in build if it is not a symbolic link to ${S}
+    if [ -d ${B}src ]; then
+        rm -rf ${B}src
+    fi
+
+}
+
+do_configure_append () {
+    # Remove the symbolic link created by go.bbclass in do_configure
+    if [ -h ${B}src ]; then
+        rm ${B}src
+    fi
+    mkdir -p ${B}src/${GO_IMPORT}
+    rsync --archive --recursive --delete ${S}/src/${GO_IMPORT}/ ${B}/src/${GO_IMPORT}/
+}
+
+python () {
+    # Only add coverage analysis if the client is newer than 2.2.x
+    if not mender_is_2_2_or_older(d):
+        # Coverage instrument the client before compiling it
+        d.prependVarFlag('do_configure', 'postfuncs', "do_instrument_client ")
+    else:
+        bb.debug(2, "The client will not be built with coverage functionality. It is too old.")
+
+}
+
+
+# ---------------------------------------- #
+#                                          #
+#           Utility functions              #
+#                                          #
+# ---------------------------------------- #
+
+def mender_is_2_2_or_older(d):
+    # Due to some infinite recursion we can only check PV when using a non-git
+    # recipe, and to detect this we check MENDER_BRANCH, which is only available
+    # for Git recipes.
+    version = d.getVar('MENDER_BRANCH')
+    if not version:
+        version = d.getVar('PV')
+
+    return (
+        version.startswith("1.")
+        or version.startswith("2.0.")
+        or version.startswith("2.1.")
+        or version.startswith("2.2.")
+    )


tests/acceptance/common.py
```diff
@@ -282,6 +282,9 @@ def common_board_setup(conn, files=None, remote_path="/tmp", image_file=None):
 
 
 def common_board_cleanup(conn):
+    # Get the coverage files from the client before the device is cleaned up
+    get_no_sftp("/data/mender/coverage*", conn, local=".")
```
kacf: I think this is only used by physical boards, which has not been used for ages and is probably buggy beyond belief by now. You should probably add this to `qemu_finalizer` instead.

oleorhagen:
> > * Also, right now, the recipe is included in `meta-mender-ci`. This should, I guess, either be a layer of it's own, or conditionally included. I can't decide which.
> 
> This is a good location considering that we are now using this for acceptance tests.
> 
> However, thinking ahead a little bit, this may not be the best location long term. Just like R/O filesystem, we want to expand this into covering integration tests as well (in particular state scripts are tested heavily here, for example), but the method in this PR will not work for that purpose. The idea I had in mind was to provide a completely separate test binary in some stowed away location, and, when `mender-qemu` is given the right flag, use a systemd service file to replace the original `mender` binary with the test binary before launching. This means two things:
> 
>     * Running vanilla `mender-qemu` launches the normal binary.
> 
>     * We can use the flag to get coverage results from integration tests as well, since they are normally built without the `meta-mender-ci` layer.
> 
> 
> What do you think?

I'm not completely sure I'm following here. 
1) State-script tests. Most likely we will not be able to get the coverage here anyways, since we kill the client through the kernel API, and I don't see any way of catching this through the test framework before rebooting.
2) `mender-qemu` flag. I think a separate layer would be more transparent in this case to be honest, just include it, and everything works as normal, no extra flags, or systemd trickery needed. If you don't want the coverage. Don't include the layer, or simply ignore the coverage files. Then this can be done the same way for integration and acceptance tests.

lluiscampos:
This is looking very promising @oleorhagen!

I don't have a lot to add besides what you and @kacf are already discussing.

> > * Yet it does not upload anything to `codecov`. I guess this should be done in `mender-qa`
> 
> Hmm, yes, this one is a bit tricky. I'm not sure if it's possible to update the coverage from a different repository (mender-qa -> mender). You'll have to look at how the upload is done for mender, and then tweak the call, so maybe you can get it uploaded from mender-qa. Another complication is that it needs to be in addition to the existing code coverage from the unit tests. Interesting challenge! :-)

The technical part should not be a big deal. The easiest: cd into mender client directory and call codecov from there (pointing to the coverage files) and the correct token. The report will be then be merged (aggregated) to the unit tests ones. The only inconvenient is that most likely we will get two PR comments (I don't think in our case it will work [as documented](https://docs.codecov.io/docs/merging-reports) because we are using to independent pipelines). I can assist on this point.

One think. I didn't specify it in the task, but I think it will be enough with qemux86_64_uefi_grub results. Do you guys think that we should aggregate results from all configurations?

> 
> * `mender-qemu` flag. I think a separate layer would be more transparent in this case to be honest, just include it, and everything works as normal, no extra flags, or systemd trickery needed. If you don't want the coverage. Don't include the layer, or simply ignore the coverage files. Then this can be done the same way for integration and acceptance tests.

I don't have an strong opinion. For simplicity, I think it is best to avoid the flag and eventually just publish the `mender-client-qemu` Docker image with coverage support. The image is intended only for testing or demo purposes, so why not?

oleorhagen:
> This is looking very promising @oleorhagen!
> 
^^
> I don't have a lot to add besides what you and @kacf are already discussing.
> 
> > > * Yet it does not upload anything to `codecov`. I guess this should be done in `mender-qa`
> > 
> > 
> > Hmm, yes, this one is a bit tricky. I'm not sure if it's possible to update the coverage from a different repository (mender-qa -> mender). You'll have to look at how the upload is done for mender, and then tweak the call, so maybe you can get it uploaded from mender-qa. Another complication is that it needs to be in addition to the existing code coverage from the unit tests. Interesting challenge! :-)
> 
> The technical part should not be a big deal. The easiest: cd into mender client directory and call codecov from there (pointing to the coverage files) and the correct token. The report will be then be merged (aggregated) to the unit tests ones. The only inconvenient is that most likely we will get two PR comments (I don't think in our case it will work [as documented](https://docs.codecov.io/docs/merging-reports) because we are using to independent pipelines). I can assist on this point.

Alright, great!
> 
> One think. I didn't specify it in the task, but I think it will be enough with qemux86_64_uefi_grub results. Do you guys think that we should aggregate results from all configurations?

Initially, I thought we wanted, this. I guess. Does it really matter, if we already do it for one? It's the same codepath, right?
> 
> > * `mender-qemu` flag. I think a separate layer would be more transparent in this case to be honest, just include it, and everything works as normal, no extra flags, or systemd trickery needed. If you don't want the coverage. Don't include the layer, or simply ignore the coverage files. Then this can be done the same way for integration and acceptance tests.
> 
> I don't have an strong opinion. For simplicity, I think it is best to avoid the flag and eventually just publish the `mender-client-qemu` Docker image with coverage support. The image is intended only for testing or demo purposes, so why not?

plus one on this

kacf:
>     1. State-script tests. Most likely we will not be able to get the coverage here anyways, since we kill the client through the kernel API, and I don't see any way of catching this through the test framework before rebooting.

Sure, but only one test does this. There are plenty of other tests that are just doing normal execution.

> > * `mender-qemu` flag. I think a separate layer would be more transparent in this case to be honest, just include it, and everything works as normal, no extra flags, or systemd trickery needed. If you don't want the coverage. Don't include the layer, or simply ignore the coverage files. Then this can be done the same way for integration and acceptance tests.
> 
> I don't have an strong opinion. For simplicity, I think it is best to avoid the flag and eventually just publish the `mender-client-qemu` Docker image with coverage support. The image is intended only for testing or demo purposes, so why not?

But then there is no image that has the real client, which is important for manual testing. I would rather keep it in `meta-mender-ci` in that case.

For integration tests I don't think we have a choice, at least not without making pretty substantial changes to our build process. We don't build a dedicated image for integration tests, so it can't be in a separate layer. It's not the first time this has been done, the reboot detector works in the same way: It is always included, but only enabled in tests.

I'm ok with leaving it as it is for now, since it's not really part of this task.

> Initially, I thought we wanted, this. I guess. Does it really matter, if we already do it for one? It's the same codepath, right?

We should have all if possible. The Raw Flash path is only covered in the `vexpress-qemu-flash` target.

oleorhagen:
> I'm ok with leaving it as it is for now, since it's not really part of this task.
> 
Okay, let's. There are a few other remaining issues that will take enough energy as is :)
> > Initially, I thought we wanted, this. I guess. Does it really matter, if we already do it for one? It's the same codepath, right?
> 
> We should have all if possible. The Raw Flash path is only covered in the `vexpress-qemu-flash` target.

Good point!

oleorhagen:
@kacf @lluiscampos 
There is a problem with this approach, stemming from go tests functionality:
```
All test output and summary lines are printed to the go command's
standard output, even if the test printed them to its own standard
error. (The go command's standard error is reserved for printing
errors building the tests.)
```

Which obviously creates some problems. In order not to destroy your opportunities at originality here, I'm asking. How would you solve this? I have an idea, but I don't like it.

lluiscampos:
> @kacf @lluiscampos
> There is a problem with this approach, stemming from go tests functionality:
> 
> ```
> All test output and summary lines are printed to the go command's
> standard output, even if the test printed them to its own standard
> error. (The go command's standard error is reserved for printing
> errors building the tests.)
> ```
> 
> Which obviously creates some problems. In order not to destroy your opportunities at originality here, I'm asking. How would you solve this? I have an idea, but I don't like it.

(Related to a recent discussion in [this PR](https://github.com/mendersoftware/meta-mender/pull/949) with `test_too_big_image_update`)
Will this be affecting many tests?  Maybe we can just change test expectations and disregard if the out is coming from stdout or stderr. For when this is important we can then test on unit level.

kacf:
> ```
> All test output and summary lines are printed to the go command's
> standard output, even if the test printed them to its own standard
> error. (The go command's standard error is reserved for printing
> errors building the tests.)
> ```
> 
> 
> Which obviously creates some problems. In order not to destroy your opportunities at originality here, I'm asking. How would you solve this? I have an idea, but I don't like it.

I think the test binary should behave as closely as possible to the real thing, so I would consider reopening all three of `os.Stdin`, `os.Stdout`, and `os.Stderr` to their 0, 1, and 2 system descriptors. Perhaps inside `TestMainInstrumented`?

oleorhagen:
> The technical part should not be a big deal. The easiest: cd into mender client directory and call codecov from there (pointing to the coverage files) and the correct token. The report will be then be merged (aggregated) to the unit tests ones. The only inconvenient is that most likely we will get two PR comments (I don't think in our case it will work [as documented](https://docs.codecov.io/docs/merging-reports) because we are using to independent pipelines). I can assist on this point.
> 

@lluiscampos I think I will take you up on this offer now.

I've had a quick look in 'mendertesting', where the publish stage has a codecov upload functionality already. Is this the correct place to put it?

Also, the link is dead (codedov docs). In fact, I can't get onto their docs site at all. Not even from their home page

lluiscampos:
> @lluiscampos I think I will take you up on this offer now.
> 
> I've had a quick look in 'mendertesting', where the publish stage has a codecov upload functionality already. Is this the correct place to put it?

Client acceptance tests are run from mender-qa/integration pipeline, so it is there that we need to collect and publish the results. The best would be to make a sync call and see from where can I take over for this part.

> Also, the link is dead (codedov docs). In fact, I can't get onto their docs site at all. Not even from their home page

Works for me.

oleorhagen:
> Client acceptance tests are run from mender-qa/integration pipeline, so it is there that we need to collect and publish the results. The best would be to make a sync call and see from where can I take over for this part.
> 

Alright, sounds good. For now, the acceptance tests will save all the coverage files in `tests/acceptance/coverage`. Is that a fair point to sync up you think?

lluiscampos:
> > Client acceptance tests are run from mender-qa/integration pipeline, so it is there that we need to collect and publish the results. The best would be to make a sync call and see from where can I take over for this part.
> 
> Alright, sounds good. For now, the acceptance tests will save all the coverage files in `tests/acceptance/coverage`. Is that a fair point to sync up you think?

That is perfect. Let me know when I can look at this.

tests/meta-mender-ci/recipes-testing/mender-coverage-client/gobinarycoverage.bb
```diff
@@ -0,0 +1,16 @@
...
+
+SRC_URI = "git://${GO_IMPORT};protocol=https;destsuffix=${PN}-${PV}/src/${GO_IMPORT}"
+SRCREV = "${AUTOREV}"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://src/${GO_IMPORT}/LICENSE;md5=86d3f3a95c324c9479bd8986968f4327"
+
+# S = "${WORKDIR}/git"
+# B = "${WORKDIR}/build"
+
+# GOPATHDIR = "${B}/src/${GO_IMPORT}"
```
kacf, line 14: Remove?

tests/acceptance/fixtures.py
```diff
@@ -131,6 +131,8 @@ def qemu_finalizer():
         def qemu_finalizer_impl(conn):
             try:
                 manual_uboot_commit(conn)
+                # Collect the coverage files from /data/mender/ if present
+                get_no_sftp("/data/mender/cover*", conn, local="coverage")
```
kacf, line 135: If there are no files, will this throw, or just silently fail? It's a cleanup function, so important that the rest also runs.

tests/meta-mender-ci/recipes-testing/mender-coverage-client/mender-client_%.bbappend
```diff
@@ -9,5 +9,18 @@ do_compile_prepend() {
     test -e ${B}/src/${GO_IMPORT} || ln -s ${S} ${B}/src/${GO_IMPORT}
     cd ${B}/src/${GO_IMPORT}
 
+    # The coverage functionality can only be applied on a clean repo
+    git reset --hard HEAD
```
kacf, line 15: Wow, risky! If you've got externalsrc set, you risk erasing all of your work every time you do a Yocto build!

A better approach would be to copy the directory instead of symlinking it. Then you get a `${B}` folder which is totally independent, and you can do whatever you want there. This also means you don't actually need to call `git reset --hard`, because it will be re-copied every time.

tests/meta-mender-ci/recipes-testing/mender-coverage-client/mender-client_%.bbappend
```diff
@@ -9,5 +9,18 @@ do_compile_prepend() {
...
     oe_runmake instrument-binary
 }
+
+do_install_append() {
+
+    cd ${B}/src/${GO_IMPORT}
+    # Due to the gobinarycoverage tool dirtying the externalsrc tree
+    # Two builds cannot be done in a row, without cleaning
+    git reset --hard HEAD
+    cd -
```
kacf, line 27: This can probably be removed if copying into the `${B}` folder.

tests/meta-mender-ci/recipes-testing/mender-coverage-client/gobinarycoverage.bb
oleorhagen: Yes, this was simply for testing it out.

tests/meta-mender-ci/recipes-testing/mender-coverage-client/mender-client_%.bbappend
oleorhagen: haha, yeh, not planning on keeping it like this for production :p 

On the bright side. This worked, and all regular acceptance tests passed: https://northern.tech.gitlab.io/-/Mender/mender-qa/-/jobs/510449468/artifacts/report_accep_qemux86_64_bios_grub.html
except for the versioned clients (which do not have this functionality as of yet).

Which begs the question, should all the released clients have this functionality present also?  Which means the PR to Mender has to be cherry picked.

Or should we change the recipe here to only build on master for now?

kacf: Hehe, sorry, I think I was a bit overeager, you had not asked for a review yet!

> Which begs the question, should all the released clients have this functionality present also? Which means the PR to Mender has to be cherry picked.
> 
> Or should we change the recipe here to only build on master for now?

No no, master only.

kacf: > On the bright side. This worked, and all regular acceptance tests passed: 

Nice! :-D

tests/acceptance/fixtures.py
oleorhagen: To be honest, I'm not sure yet, but it's on the roadmap to figure it out ;)

tests/meta-mender-ci/recipes-testing/mender-coverage-client/mender-client_%.bbappend
```diff
@@ -0,0 +1,67 @@
+SUMMARY = "Build a Mender client which records coverage during its invocation"
+
+DEPENDS += "gobinarycoverage rsync-native"
+
+do_copy_src_to_build[nostamp] = "1"
```
kacf, line 5: Why do you need `nostamp`?

tests/meta-mender-ci/recipes-testing/mender-coverage-client/mender-client_%.bbappend
```diff
@@ -0,0 +1,67 @@
...
+do_copy_src_to_build[doc] = "Copies the Mender client source from ${S}/src/${GO_IMPORT} \
+to ${B}/src/${GO_IMPORT}"
+do_copy_src_to_build[dirs] = "${B}src/github.com/mendersoftware/mender"
+do_copy_src_to_build () {
+    bbwarn "Copying source to the build dir..."
+    rsync -a ${S}/src/${GO_IMPORT}/ ${B}/src/${GO_IMPORT}/
+    cd ${B}/src/${GO_IMPORT}/
+    # Why is this needed?
+    git reset --hard HEAD
```
kacf, line 14: I also don't see why! What happens if you remove it?

tests/meta-mender-ci/recipes-testing/mender-coverage-client/mender-client_%.bbappend
```diff
@@ -0,0 +1,67 @@
...
+    oe_runmake instrument-binary
+}
+
+python () {
+    # Only add coverage analysis if the client is newer than 2.2.x
+    if not mender_is_2_2_or_older(d):
+        bb.warn("Building the client with coverage instrumentation enabled")
+        # Disable do_configure inherited from go.bbclass. This function tries to
+        # symlink to the builddir from source i.e., ln -s ${S}/src ${B}/src.
+        d.setVarFlag('do_configure', 'noexec', "1")
```
kacf, line 35: I would perhaps argue that it's better to be surgical and remove only the part we don't want, and not the whole function. We don't know what it does, or will do in the future, besides making the symlink. Can you simply use `do_configure_append` and do the link removal and copying there?

tests/meta-mender-ci/recipes-testing/mender-coverage-client/mender-client_%.bbappend
```diff
@@ -0,0 +1,67 @@
...
+        # Copy the source over to the build directory instead of symlinking
+        # so that the source can be dirtied freely
+        # addtask do_copy_src_to_build before do_compile after do_configure
+        bb.build.addtask('do_copy_src_to_build', 'do_compile', 'do_configure', d)
+        # Coverage instrument the client before compiling it
+        d.prependVarFlag('do_compile', 'prefuncs', "do_instrument_client ")
```
kacf, line 41: Why not do `addtask` here too?

oleorhagen: Because I dont want a separate stamp for it, it is directly related to re-compilation

tests/meta-mender-ci/recipes-testing/mender-coverage-client/mender-client_%.bbappend
oleorhagen: Fails, with a dirty source dir when applying the patch

tests/meta-mender-ci/recipes-testing/mender-coverage-client/mender-client_%.bbappend
oleorhagen: Smrt!
Way better than this approach. Lets do create the inverse function!

tests/meta-mender-ci/recipes-testing/mender-coverage-client/mender-client_%.bbappend
oleorhagen: To make sure it is run everytime the recipe is executed

oleorhagen:
@kacf this is now ready for review :)

tests/meta-mender-ci/recipes-testing/mender-coverage-client/mender-client_%.bbappend
```diff
@@ -0,0 +1,60 @@
...
+do_instrument_client[doc] = "Modifies the Mender client source to enable coverage analysis"
+do_instrument_client () {
+    cd ${B}/src/${GO_IMPORT}
+    bbwarn "Building the Mender client with coverage analysis"
+    oe_runmake instrument-binary
+}
+
+do_configure_prepend () {
+    # Remove all the src present in build if it is not a symbolic link to ${S}
+    if [ -d ${B}src ]; then
```
kacf, line 15: Good check! :+1:

oleorhagen:
> Nice! This looks quite clean now!
> 
> I noticed that coverage upload is missing. Do you have any idea how to upload it to the Mender client coverage results? Is it even possible, or do we have track it separately?

@lluiscampos has said that he will take it over at this point, and handle the coverage upload, if I am not mistaken?

lluiscampos:
> Nice! This looks quite clean now!
> 
> I noticed that coverage upload is missing. Do you have any idea how to upload it to the Mender client coverage results? Is it even possible, or do we have track it separately?

@oleorhagen and me agreed in that I will take over this part. I think I can work on this this afternoon. I suggest we merge the PR and I take it from there in the same JIRA ticket.

kacf:
> has said that he will take it over at this point, and handle the coverage upload

:+1:

PR #953, opened 2020-03-23
raspberrypi: add state script to update boot firmware

mirzak:
Raspberry Pi boards have a set of boot firmware files that are
located on the vfat boot part, and these files are normally not updated
when you perform an update of the root filesystem.

Occasionally there will be changes to the Raspberry Pi software stack
that requires that these files are update. Typically there is something
in the Linux kernel that requires something that is in the boot
firmware. This means that to update the Linux kernel you must also
update the boot firmware files.

Above is really sub-optimal design of the Raspberry Pi boards but a
limitation that we must live with.

Note that the DTB files are also part of "boot firmware" and included in
the state script update.

The provided state-script can be enabled by adding the following to e.g
local.conf:

    INHERIT += "rpi-update-firmware"

The ArtifactInstall_Leave_50 script can be overriden to customize what
files to update, e.g only DTB files. By default all files on the boot
part will be updated including config.txt and cmdline.txt.

Changelog: Add state-script that can be enabled to update RPi boot firmware files

### DIFF
--- a/meta-mender-raspberrypi/classes/rpi-update-firmware.bbclass
+++ b/meta-mender-raspberrypi/classes/rpi-update-firmware.bbclass
@@ -1,0 +1,16 @@
+rpi_install_firmware_to_rootfs() {
+    install -d ${IMAGE_ROOTFS}/boot/firmware/overlays
+
+    cp ${DEPLOY_DIR_IMAGE}/bcm2835-bootfiles/* ${IMAGE_ROOTFS}/boot/firmware/
+
+    # To exclude files such as bcm2710-rpi-3-b-1-4.19.88+git0+988cc7beac-r0-raspberrypi3-20200323173633.dtb
+    # as only the link names are actually valid and searched for on the device.
+    find ${DEPLOY_DIR_IMAGE}/ -type l \( -iname "*.dtb" \) -exec cp {} ${IMAGE_ROOTFS}/boot/firmware/ \;
+    find ${DEPLOY_DIR_IMAGE}/ -type l \( -iname "*.dtbo" \) -exec cp {} ${IMAGE_ROOTFS}/boot/firmware/overlays/ \;
+
+    cp ${DEPLOY_DIR_IMAGE}/u-boot.bin ${IMAGE_ROOTFS}/boot/firmware/${SDIMG_KERNELIMAGE}
+    cp ${DEPLOY_DIR_IMAGE}/boot.scr ${IMAGE_ROOTFS}/boot/firmware/
+}
+ROOTFS_POSTPROCESS_COMMAND += "rpi_install_firmware_to_rootfs; "
+
+IMAGE_INSTALL_append = " update-firmware-state-script"
--- a/meta-mender-raspberrypi/recipes-mender/update-firmware-state-script/files/ArtifactInstall_Leave_50.in
+++ b/meta-mender-raspberrypi/recipes-mender/update-firmware-state-script/files/ArtifactInstall_Leave_50.in
@@ -1,0 +1,38 @@
+#!/bin/sh
+
+set -e
+
+function trap_exit() {
+  umount --lazy /mnt/inactive_part
+}
+trap trap_exit EXIT
+
+MENDER_ROOTFS_PART_A="@@MENDER_ROOTFS_PART_A@@"
+MENDER_ROOTFS_PART_B="@@MENDER_ROOTFS_PART_B@@"
+
+if mount | grep ${MENDER_ROOTFS_PART_A}; then
+  inactive_part="${MENDER_ROOTFS_PART_B}"
+else
+  inactive_part="${MENDER_ROOTFS_PART_A}"
+fi
+
+mkdir -p /mnt/inactive_part
+mount -o ro ${inactive_part} /mnt/inactive_part
+
+# These are dangerous operations and if they fail (partial copy) it might
+# render the device unusable.
+
+# Copy 'core' firmware files first
+find /mnt/inactive_part/boot/firmware/ -maxdepth 1 -type f | xargs -I {} cp -v {} /uboot/
+
+# Synchronize before trying to copy the rest of the files
+sync
+
+# Copy overlays
+find /mnt/inactive_part/boot/firmware/overlays/ -maxdepth 1 -type f | xargs -I {} cp -v {} /uboot/overlays/
+
+# Synchronize to ensure all files are written before leaving the ArtifactInstall state
+sync
+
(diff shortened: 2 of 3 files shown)


guillaumekh:
I might be wrong but i think uImage is missing

mirzak:
Intentionally left it out as it is actually not used the boot firmware nor bootloader. It is loaded from the rootfs `/boot/uImage` and this is actually already part of the rootfs updates.

guillaumekh:
This is great, thanks @mirzak 

A few ideas I think I'll try to implement on top:
- some conditional logic to only run the hazardous operations if required. Not sure yet of how I'll proceed. I'm thinking of dumping ${KERNEL_VERSION} to a .kernelversion textfile, and comparing it to `uname -r` on state script evaluation
- bundling the files in a tar.gz file to save some space

meta-mender-raspberrypi/recipes-mender/update-firmware-state-script/files/ArtifactInstall_Leave_50
```diff
@@ -0,0 +1,36 @@
...
+
+# Copy 'core' firmware files first
+find /mnt/inactive_part/boot/firmware/ -maxdepth 1 -type f | xargs -I {} cp -v {} /uboot/
+
+# Synchronize before trying to copy the rest of the files
+sync
+
+# Copy overlays
+find /mnt/inactive_part/boot/firmware/overlays/ -maxdepth 1 -type f | xargs -I {} cp -v {} /uboot/overlays/
+
```
kacf: I would consider adding a `sync` here too, to make sure the client does not leave the ArtifactInstall state without having everything in physical storage.

meta-mender-raspberrypi/recipes-mender/update-firmware-state-script/files/ArtifactInstall_Leave_50
```diff
@@ -0,0 +1,36 @@
...
+trap trap_exit EXIT
+
+echo "$(basename "$0") was called!" >&2
+
+# Check if mmcbkl0p2 is active part
+mount | grep /dev/mmcblk0p2
+if [ $? -eq 0 ]; then
+  inactive_part="/dev/mmcblk0p3"
+else
+  inactive_part="/dev/mmcblk0p2"
```
kacf: This will fit the vast majority, but I would still consider a small `source ${WORKDIR}/variables.sh` and corresponding:

```bash
do_compile() {
    cat > ${WORKDIR}/variables.sh <<EOF
MENDER_ROOTFS_PART_A=${MENDER_ROOTFS_PART_A}
MENDER_ROOTFS_PART_B=${MENDER_ROOTFS_PART_B}
EOF
}
```

mirzak: Makes sense, but complicates things a bit. 

As this is something that you can enable/disable it can not be bound to the `update-firmware-state-script` recipe as the script will run on the "old" image where this might not be present.

So we would need to do one of the following:

1. Add a separate recipe that always inställes a sourcable file for Raspberry Pi boards
2. Above, but make it generic that is put it in meta-mender-core, could be part of the mender-client recipe, some update modules could benefit from this?
3. Parse `/var/lib/mender/mender.conf` to extract the necessary information

Do you have a prefererad method?

meta-mender-raspberrypi/recipes-mender/update-firmware-state-script/files/ArtifactInstall_Leave_50
mirzak: Ack

mirzak:
> Will this be documented somewhere?

Was planning on writing a [tutorial](https://hub.mender.io/c/tutorials/mender/21) for this as it is very board specific. Can follow-up and update the README in this repository as well

mirzak:
> I think the message Changelog: Add state-script that can be enabled to update RPi boot firmware files should be a bit more clear that this has no rollback, even if the rest of the artifact is rolled back.

Update it to `Changelog: Commit`

guillaumekh:
Any chance of this being backported to Warrior branch on merge ?

meta-mender-raspberrypi/recipes-mender/update-firmware-state-script/files/ArtifactInstall_Leave_50
kacf: Right, I see what you mean. A separate recipe feels very heavyweight for this. Although I like method 3, unfortunately the JSON makes it a bit tricky if `jq` isn't installed. I have a 4th suggestion then:

1. Start your script with:
   ```bash
   #!/bin/sh

   @PUT_MENDER_ROOTFS_PART_A_HERE@
   @PUT_MENDER_ROOTFS_PART_B_HERE@
   ```
2. Use a sed script to replace those during the build, instead of during execution.

kacf:
> Any chance of this being backported to Warrior branch on merge ?

Given that this is completely optional, and needs to be turned on, I think that should be no problem. Unless you can see incompatibilities, @mirzak?

meta-mender-raspberrypi/recipes-mender/update-firmware-state-script/files/ArtifactInstall_Leave_50
mirzak: > I have a 4th suggestion then:

mirzak:
> Given that this is completely optional, and needs to be turned on, I think that should be no problem. Unless you can see incompatibilities, @mirzak?

I do not see any problems with backporting.

mirzak:
PR has been updated based on feedback and should be ready for a final review

kacf:
Failure is for sure unrelated, this layer is not used in integration tests.

kacf:
Running a simple build test of meta-mender-raspberrypi, since this doesn't run by default.

mirzak:
Will you backport this to warrior as port of the bulk or do you want me to do it? Either way I would like to test it once set up the PR

kacf:
Done here: #957.

michaek:
I realize this is an old PR, but I had a question about the class it provides. It is a common case that we might want to only update a setting in config.txt and not replace all of the device tree. Is there reason to think that a modified version of rpi-update-firmware.bbclass that added only config.txt to the image would work properly with the state script to update config.txt in the firmware? Thanks!

michaek:
If only I had read the original PR description more carefully. It says clearly that the preferred way to filter the update is to override the state script!

PR #954, opened 2020-03-27
grub-efi: Update precompiled binaries with dynamic root support.

drewmoseley:
Changelog: None


drewmoseley:
These binaries were built from the warrior branch of vexpress-qemu since I already had a build for that branch.

kacf:
I'm ok with this, but be aware of [this commit](https://github.com/mendersoftware/meta-mender/commit/da4e6575a22700714167c4f94a63dd212c5ba543#diff-99dd8cee4fd64b4925552dc888a821b3R23), headed for zeus, which will break this again very soon. After this, mender-convert and the precompiled version of grub-efi will download from the same source, so the place this really has to be fixed is [here](https://github.com/mendersoftware/mender-convert-integration-scripts/blob/master/grub-efi/docker-create-grub-efi-binaries#L7).

Should I create a ticket for this, or do you want to handle that right away?

PR #955, opened 2020-03-30
u-boot: Make sure auto-patcher handles multiline Kconfig entries.

kacf:
This would lead to the configure step getting stuck in a config screen
which would repeatedly ask for input over and over, consuming all
memory.

Changelog: Fix infinite loop which would eat all available memory
on the build host, when using U-Boot v2020.01 or later together with
the auto-patcher.

MEN-3265


kacf:
Thanks!

PR #956, opened 2020-03-31
MEN-3264: Restore lsb_release tool for testing purposes.

kacf:
Changelog: None


kacf:
This is just a reimport of a script that used to exist as a bbappend file.

PR #958, opened 2020-04-02
Update to new grub-mender-grubenv SHA.

kacf:
Changelog: grubenv: Handle debug command prompt when running as EFI
app.


kacf:
Passed all relevant tests in https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/133532752.

PR #959, opened 2020-04-02
Zeus

kacf:


kacf:
@mirzak: I'm particularly interested in your opinion on the Raspberry Pi patch. I have not tested all of those, but it felt natural that the patch should go everywhere.

mirzak:
Looks good, it was supposed to be merged in to one patch for U-Boot v2020.01.

PR #960, opened 2020-04-02
mender: Add sanity check to ensure partuuid with X86.

drewmoseley:
Changelog: Title

### DIFF
--- a/meta-mender-core/classes/mender-setup.bbclass
+++ b/meta-mender-core/classes/mender-setup.bbclass
@@ -330,6 +330,9 @@ mender_sanity_handler[eventmask] = "bb.event.ParseCompleted"
 python mender_sanity_handler() {
     if bb.utils.contains('MENDER_FEATURES_ENABLE', 'mender-partuuid', True, False, d) and d.getVar('MENDER_STORAGE_DEVICE', True) != "":
         bb.warn("MENDER_STORAGE_DEVICE is ignored when mender-partuuid is enabled. Clear MENDER_STORAGE_DEVICE to remove this warning.")
+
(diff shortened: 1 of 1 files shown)


meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -330,6 +330,9 @@ mender_sanity_handler[eventmask] = "bb.event.ParseCompleted"
 python mender_sanity_handler() {
     if bb.utils.contains('MENDER_FEATURES_ENABLE', 'mender-partuuid', True, False, d) and d.getVar('MENDER_STORAGE_DEVICE', True) != "":
         bb.warn("MENDER_STORAGE_DEVICE is ignored when mender-partuuid is enabled. Clear MENDER_STORAGE_DEVICE to remove this warning.")
+
+    if bb.utils.contains('MENDER_FEATURES_ENABLE', 'mender-partuuid', True, False, d) and not bb.utils.contains_any('TUNE_ARCH', 'x86 x86_64', True, False, d):
+        bb.fatal("mender-partuuid is not supported on the %s architecture." % d.getVar("TUNE_ARCH"))
```
mirzak, line 335: I think the trigger here should be if `uboot-mender` is enabled, because this would work in the U-boot/GRUB case which?

kacf: Yes, that is my gut feeling too, although I haven't tested it. I would restrict this to trigger only if `mender-partuuid` and `mender-uboot` are enabled together. The architecture is of less importance.

drewmoseley: Fixed.

drewmoseley: Weird.  My branch is updated but the PR doesn't show the update.

drewmoseley:
I'm not sure why this PR is busted. I'll create a new one.

PR #961, opened 2020-04-03
raspberrypi: update Mender integration patch path

mirzak:
Missed that you renamed the file in #959


kacf:
Thanks, missed that one!

PR #963, opened 2020-04-03
mender-client-migrate-configuration: Rename mender to mender-client in DEPENDS.

mischief901:
It looks like one of these was missed in commit e65ca3a44414ce9d08180ddadfaaef14f9efbfa3.

This should fix it. Thanks.


kacf:
Oh, and please also target the master branch. zeus-test is just a temporary branch which will be deleted.

kacf:
Merged in #958.

PR #972, opened 2020-04-20
raspberrypi: busybox compatibility for boot firmware state script

guillaumekh:
The ArtifactInstall_Leave_50 state script fails on busybox systems because umount does not understand the long `--lazy` flag :

> stderr collected while running script /var/lib/mender/scripts/ArtifactInstall_Leave_50 [umount: unrecognized option '--lazy'\nBusyBox v1.27.2 (2018-12-03 18:22:35 UTC) multi-call binary.\n\nUsage: umount [OPTIONS] FILESYSTEM|DIRECTORY\n]" module=executor

The short `-l` flag is useable with both util-linux and busybox versions of umount.

### DIFF
--- a/meta-mender-raspberrypi/recipes-mender/update-firmware-state-script/files/ArtifactInstall_Leave_50.in
+++ b/meta-mender-raspberrypi/recipes-mender/update-firmware-state-script/files/ArtifactInstall_Leave_50.in
@@ -3,7 +3,7 @@
 set -e
 
 function trap_exit() {
-  umount --lazy /mnt/inactive_part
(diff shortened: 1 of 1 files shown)


guillaumekh:
cc @mirzak

guillaumekh:
Botched this a little but there it is again with the requested changes.

guillaumekh:
Thanks @kacf 

This will need to be backported to warrior branch too.

kacf:
That should be no problem. Cherry-picks to warrior are handled monthly, so it should be picked up in 10+ days or so.

PR #974, opened 2020-04-20
[tests] Create coverage dir before calling scp

lluiscampos:
Otherwise all remote reports get copied to the same local file
named "coverage", overriding it every time.


lluiscampos:
To be tested together with https://github.com/mendersoftware/mender-qa/pull/393

PR #975, opened 2020-04-23
More zeus patches

kacf:


kacf:
It was quite lucky that I discovered the missing `CONFIG_SYS_REDUND_ENVIRONMENT` switch. It just happened to be an undefined reference to it in the Raspberry Pi build. But it silently affects all boards, and essentially removes the redundant environment, increasing the risk of bricking in a way that can't be fixed by an update (it's in the bootloader).

I'm a bit surprised that the mismatch between the bootloader and the user space tools (which should still be using a redundant environment) didn't show up anywhere. I didn't look into it though, and we should be safe now with the test for the define.

meta-mender-raspberrypi/recipes-bsp/u-boot/patches/0001-configs-rpi-enable-mender-requirements.patch
```diff
@@ -55,11 +58,14 @@ index 4e8204ef88..4f4029e006 100644
...
  CONFIG_DM_KEYBOARD=y
-@@ -43,3 +41,4 @@ CONFIG_SYS_WHITE_ON_BLACK=y
+@@ -43,3 +41,7 @@ CONFIG_SYS_WHITE_ON_BLACK=y
  CONFIG_CONSOLE_SCROLL_LINES=10
  CONFIG_PHYS_TO_BUS=y
  CONFIG_OF_LIBFDT_OVERLAY=y
 +CONFIG_ENV_IS_IN_MMC=y
++CONFIG_ENV_OFFSET=0x400000
++CONFIG_ENV_OFFSET_REDUND=0x800000
++CONFIG_SYS_REDUNDAND_ENVIRONMENT=y
```
mirzak, line 68: Would it be worth it to provide these in [u-boot-mender-common.inc](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/recipes-bsp/u-boot/u-boot-mender-common.inc) in a post configure step based on `MENDER_PARTITION_ALIGNMENT`. This would be similar to what we have been doing up until now.

Or do you see any big drawbacks with attempting above?

meta-mender-core/recipes-bsp/u-boot/patches/0002-Generic-boot-code-for-Mender.patch
```diff
@@ -58,6 +56,10 @@ index 0000000000..69a526805d
 +# error CONFIG_BOOTCOUNT_ENV is required for Mender to work. Make sure that: 1) All the instructions at docs.mender.io/devices/integrating-with-u-boot have been followed. 2) All required layers are included in bblayers.conf, including any board specific layers such as meta-mender-<board>
 +#endif
 +
++#ifndef CONFIG_SYS_REDUNDAND_ENVIRONMENT
++# error CONFIG_SYS_REDUNDAND_ENVIRONMENT is required for Mender to work. Make sure that: 1) All the instructions at docs.mender.io/devices/integrating-with-u-boot have been followed. 2) All required layers are included in bblayers.conf, including any board specific layers such as meta-mender-<board>
++#endif
```
mirzak, line 61: This would break the build on many U-Boot forks out there I am wondering if we can avoid this, but I do not see easy way of doing this unless we can detect U-Boot version in pre-processing.

Otherwise we probably need to keep two versions of this patch pre/post 2020.01

kacf: Yes, I have the same concern. However, `CONFIG_SYS_REDUNDAND_ENVIRONMENT` was used even prior to them moving it into `Kconfig`, but then it was defined based on whether `CONFIG_ENV_OFFSET_REDUND` was defined. So I'm thinking one of two scenarios is going to happen:

1. The U-Boot build already defines it based on other conditions. If so this will just silently work.
2. The U-Boot build doesn't define it (at least not before getting here), which means the user has to do it. However, it has no ill effects as it either doesn't influence the build, or means the same thing as it does in v2020.01 and onwards.

Does that make sense?

It would be possible to conditionally detect this by looking inside the `Kconfig` files in the build, but given that defining `CONFIG_SYS_REDUNDAND_ENVIRONMENT` should have no ill effects, I'm leaning towards avoiding that additional complexity.

meta-mender-raspberrypi/recipes-bsp/u-boot/patches/0001-configs-rpi-enable-mender-requirements.patch
kacf: Yes, it annoys me a bit that something that was automatic before, won't be anymore. But patching this is much more intrusive than before, since we can no longer introduce a new file, we have to patch it inline. This is part of the auto-patcher, so my thinking was that maybe we should keep the non-auto-patching path clean?

What do you think?

mirzak: I am torn.  Agree with you, annoying and we should keep the non-auto-patching path clean. 

But this will also break all the existing board integrations using non-auto-patching. I guess the middle ground is providing more explicit information in the error message as I would expect many people to hit this when they upgrade from warrior -> zeus.

meta-mender-core/recipes-bsp/u-boot/patches/0002-Generic-boot-code-for-Mender.patch
mirzak: > CONFIG_SYS_REDUNDAND_ENVIRONMENT was used even prior to them moving it into Kconfig

Ah, then it is not as a big issue as I thought and agree that we should avoid additional complexity.

meta-mender-raspberrypi/recipes-bsp/u-boot/patches/0001-configs-rpi-enable-mender-requirements.patch
kacf: Yes, I'm also torn. 🙄 But I think I agree that it's better to keep the non-auto-patching clean. I guess the good news is that moving it into Kconfig increases the chance that the auto-patcher will work.

But good point about the message, I'll try to make it more explicit, and mention exactly what people need to do to fix it.

kacf: Looks like this needs a bit more work. Not defining the parameters actually crash the *build machine*! I have filed it under [MEN-3476](https://tracker.mender.io/browse/MEN-3476), but will return to it soon.

kacf: @mirzak: Take a look now! I have added links to our Troubleshooting section in the error messages in the patch, and this goes together with https://github.com/mendersoftware/mender-docs/pull/952.

mirzak: Very nice information on what needs to be done to remedy the problems. Thank you for improving this and this patch set looks good to me

wezzynl:
Should this be on the `dunfell` branch too or am I overlooking something? Seems that branch still has the incorrect values? I've set `MENDER_PARTITION_ALIGNMENT = "4194304"` as that seems to be required for building against Rapberry Pi 3B+ but I cannot get Yocto to build without manually adjusting `0001-configs-rpi-enable-mender-requirements.patch` so it patches with the below values:

```
CONFIG_ENV_OFFSET=0x400000
CONFIG_ENV_OFFSET_REDUND=0x800000
```

mirzak:
>  I've set MENDER_PARTITION_ALIGNMENT = "4194304" as that seems to be required for building against Rapberry Pi 3B+

Actually you should not set this anymore for Raspberry Pi. It is not really required and just there for a historical reason.

So on dunfell you should remove the `MENDER_PARTITION_ALIGNMENT = "4194304"` and use the defaults, then the it will build just fine.

PR #979, opened 2020-05-08
MEN-3477: Start downloading grub-efi from grub-mender-grubenv URLs.

kacf:
The URLs include the revision SHA of grub-mender-grubenv as well as
the version of GRUB, which means we can get a stable checksum instead
of a constantly changing one.

Also standardize on some things that are shared between the
grub-efi-mender-precompiled and grub-mender-grubenv recipes, such as
URIs, revisions and the GRUB module list. We also check that the GRUB
module list defined in Yocto is the same as the one used to compile
the binaries.

Changelog: None

### DIFF
--- a/meta-mender-core/classes/grub-mender-grubenv.bbclass
+++ b/meta-mender-core/classes/grub-mender-grubenv.bbclass
@@ -1,0 +1,7 @@
+GRUB_MENDER_GRUBENV_REV = "6ffc280e64dd7b6bbc52f7747609f11fbdd1a057"
+GRUB_MENDER_GRUBENV_SRC_URI ?= "git://github.com/mendersoftware/grub-mender-grubenv;protocol=https;branch=master;rev=${GRUB_MENDER_GRUBENV_REV}"
+
+GRUB_BUILDIN = "boot linux ext2 fat serial part_msdos part_gpt normal \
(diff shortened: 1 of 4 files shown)


meta-mender-core/classes/grub-mender-grubenv.bbclass
```diff
@@ -0,0 +1,7 @@
+GRUB_MENDER_GRUBENV_REV = "6ffc280e64dd7b6bbc52f7747609f11fbdd1a057"
```
lluiscampos, line 1: You need to update this checksum, right? The [last build](https://gitlab.com/Northern.tech/Mender/grub-mender-grubenv/-/jobs/546363759) has `aa7e8f8c76c6aca6dca1820aaa42dc2cbf9762a1`.

lluiscampos:
A bit late on my review... Sorry for that.

meta-mender-core/classes/grub-mender-grubenv.bbclass
kacf: I already fixed that with a direct push! :-) This feature branch is not tested in Gitlab (yet), and there was nothing new to review besides the checksum, so I skipped a PR for this simple change.

PR #985, opened 2020-05-15
Add OpenSSL as dependency for Git versions of the Mender client.

kacf:
This actually adds an unnecessary dependency when using Git SHAs for
older versions of Mender, but this should be harmless, and it won't
affect production recipes.

Changelog: Title


kacf:
To be tested with https://github.com/mendersoftware/mender/pull/527.

kacf:
@estenberg, @lluiscampos: FYI. This shouldn't be a difficult dependency to satisfy, since OpenSSL is installed on virtually all Linux devices.

Eventually this should go into release recipes too, but this will be at release time.

lluiscampos:
> Eventually this should go into release recipes too, but this will be at release time.

And into the deb packages dependencies.

PR #986, opened 2020-05-19
zeus: Several fixes for UBI and auto-patching

kacf:
```
commit 7ed12f57d3add2530df9db8ca74847e033b57c69
Date:   Tue May 19 17:24:04 2020

    MEN-3514: Explicitly add some UBI/Flash options that we previously assumed.
    
    This would work previously because the combination of defines
    implicitly enabled what we needed. But from U-Boot v2020.01 onwards we
    need to add the options explicitly because they have been moved to
    Kconfig.
    
    Changelog: None
    
commit 63a1dc7fe49685376e9dbbdf5f956be3531d36e6
Date:   Tue May 19 15:58:26 2020

    MEN-3514: Make Kconfig option adder smart enough to detect satisfied deps.
    
    This is needed because previously, if we encountered a dependency for
    a config option, which had multiple alternatives, we would blindly
    return the first alternative. However, this returned the wrong
    alternative for UBI based boards (CONFIG_ENV_IS_IN_EEPROM). Instead we
    need to make it aware that one of the later alternatives is already
    enabled (CONFIG_ENV_IS_IN_UBI).
    
    Fix by assembling the list of already set options beforehand, and use
    that to satisfy dependencies while adding new options.
    
    Changelog: Fix U-Boot auto-patcher sometimes adding two conflicting
    options to the defconfig file.
    
```


meta-mender-core/recipes-bsp/u-boot/files/add_kconfig_option_with_depends.py
```diff
@@ -23,20 +23,30 @@ def term_parse(term_string):
     algebra = boolean.BooleanAlgebra()
     return algebra.parse(term_string)
 
-def term_reduce_to_first_alternative(term):
+def term_reduce_to_one_alternative(term, removal_terms):
```
lluiscampos, line 26: It is the first time I am looking at this piece of code, but the logic below seems correct. What I don't understand is why we named it `removal_terms` here but is `already_added` by the caller. I think the latter or something similar is better.

kacf: The thinking was that the `term` set of functions is more related to logical algebra, not strictly related to Kconfig, so I switched the terminology to fit "independently" into that function. But maybe that just makes it more confusing. I'm ok with changing it if you think `already_added` is clearer.

lluiscampos: I see your point. It is not important, let's leave it as it is.

PR #987, opened 2020-05-25
MEN-3580: Catch SIGCHLD errors in mender-artifact snapshot test.

kacf:
We do this by running both with and without terminal.

Changelog: None

### DIFF
--- a/tests/acceptance/test_snapshot.py
+++ b/tests/acceptance/test_snapshot.py
@@ -149,7 +149,9 @@ class TestSnapshot:
 
     @pytest.mark.min_mender_version("2.2.0")
     @pytest.mark.only_with_image("uefiimg", "sdimg", "biosimg", "gptimg")
-    def test_snapshot_using_mender_artifact(self, bitbake_variables, connection):
+    def test_snapshot_using_mender_artifact(
+        self, bitbake_path, bitbake_variables, connection
+    ):
         try:
             (active, passive) = determine_active_passive_part(
                 bitbake_variables, connection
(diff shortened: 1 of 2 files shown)


kacf:
I will first do a test run where I verify that this test case fails (specifically for Mender 2.4.0b1-build1), then afterwards I will make sure it passes for newer revisions.

tests/acceptance/test_snapshot.py
```diff
@@ -149,7 +149,12 @@ def test_snapshot_avoid_deadlock(self, connection):
 
     @pytest.mark.min_mender_version("2.2.0")
     @pytest.mark.only_with_image("uefiimg", "sdimg", "biosimg", "gptimg")
-    def test_snapshot_using_mender_artifact(self, bitbake_variables, connection):
+    # Make sure we run both with and without terminal. Many signal bugs lurk in
+    # different corners of the console code.
+    @pytest.mark.parametrize("terminal", ["", "screen -D -m -L"])
```
alfrunes, line 154: Nice!

tests/acceptance/test_snapshot.py
```diff
@@ -159,11 +164,20 @@ def test_snapshot_using_mender_artifact(self, bitbake_variables, connection):
...
+                        terminal,
+                        common_args,
+                        connection.user,
+                        connection.host,
+                        connection.port,
+                    ),
+                    shell=True,
+                )
+            finally:
+                subprocess.call("cat screenlog.0 ; rm -f screenlog.0", shell=True)
```
alfrunes, line 180: Can we be sure that `screenlog.0` remains constant?

kacf: You're right, there's an argument, I'll change to that!

kacf:
Updated!

kacf:
Confirmed failing for the old code in https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/566395930, and passing in the latest.

lluiscampos:
Great work @kacf!

PR #990, opened 2020-05-26
Warrior branch update

lluiscampos:


lluiscampos:
Testing together with poky/warrior-next @ [Pipeline 149583033](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/149583033).

lluiscampos:
> Testing together with poky/warrior-next @ [Pipeline 149583033](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/149583033).

I forgot to enable all client build/test jobs... New build @ [Pipeline 149678106](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/149678106)

eigendude:
It seems like, after the data partition resize changes, two services are now responsible for resizing the data partition:

* `mender-grow-data.service` (calls `/usr/bin/mender-client-resize-data-part`)
* `mender-systemd-growfs-data.service` (calls `/lib/systemd/systemd-growfs /data`)

As expected, the first service only runs once, and logs this on subsequent runs:

```
Jan 01 00:00:14 sundstrom mender-client-resize-data-part[1075]: Disk has already been resized.
```

However, on subsequent runs, the second service always seems to run. In the log:

```
Jan 01 00:00:26 sundstrom systemd[1]: Starting Grow File System on /data...
Jan 01 00:00:26 sundstrom systemd[1]: Started Grow File System on /data.
```

And the following appears at boot:

```
[   26.729861] EXT4-fs (mmcblk0p4): resizing filesystem from 2600960 to 2600960 blocks
```

Should the second service be removed?

eigendude:
I see, `mender-grow-data.service` resizes the partition, and `mender-systemd-growfs-data.service` resizes the filesystem. Maybe a feature add can be wrapping the filesystem service so that it doesn't resize at every boot?

kacf:
I think it doesn't matter. The reason that we implemented detection of an already resized partition for the first service is that parted is "dumb", and will always write the partition table, even if it writes the same as was already there. This is a risky write, as there is no recovery if you have a power loss. So it's important that is is only done once.

The second service however, uses the online resizer of the kernel, which is much smarter, and uses the ext4 journal and all that. Now, I haven't checked specifically under the hood whether the operation is a no-op, but I strongly suspect that after printing that message, the kernel realizes there is nothing to do and just returns. And even if it doesn't, it will do something which is both safe and minimal.

eigendude:
Totally makes sense. Awesome

PR #991, opened 2020-05-26
Fix RNG on zeus, plus clean up warnings

kacf:
```
commit 84456b8c556c4a28cfeca990c83289bbf23bf25a
Date:   Tue May 26 11:30:37 2020

    Fix broken RNG in vexpress-qemu sessions.
    
    The symptom of this is that both Mender key generation and SSH key
    generation hang on reading from /dev/random.
    
    It is unknown what causes the RNG to be broken. It is driven by rngd,
    which doesn't start correctly. It eventually starts successfully if
    restarted, but not until the rest of the boot has completed, and it
    may take several attempts even then. This is typical output, followed
    by the eventual successful start:
    
    ```
    May 26 09:25:35 vexpress-qemu systemd[1]: Started Hardware RNG Entropy Gatherer Daemon.
    May 26 09:25:35 vexpress-qemu rngd[1479]: Initializing available sources
    May 26 09:25:35 vexpress-qemu rngd[1479]: Failed to init entropy source hwrng
    May 26 09:25:35 vexpress-qemu rngd[1479]: JITTER rng fails with code 2
    May 26 09:25:35 vexpress-qemu rngd[1479]: Failed to init entropy source jitter
    May 26 09:25:35 vexpress-qemu rngd[1479]: can't open any entropy source
    May 26 09:25:35 vexpress-qemu rngd[1479]: Maybe RNG device modules are not loaded
    May 26 09:25:35 vexpress-qemu systemd[1]: rngd.service: Main process exited, code=exited, status=1/FAILURE
    May 26 09:25:35 vexpress-qemu systemd[1]: rngd.service: Failed with result 'exit-code'.
    May 26 09:25:40 vexpress-qemu systemd[1]: rngd.service: Service RestartSec=5s expired, scheduling restart.
    May 26 09:25:40 vexpress-qemu systemd[1]: rngd.service: Scheduled restart job, restart counter is at 6.
    May 26 09:25:51 vexpress-qemu systemd[1]: Stopped Hardware RNG Entropy Gatherer Daemon.
    May 26 09:25:51 vexpress-qemu systemd[1]: Started Hardware RNG Entropy Gatherer Daemon.
    May 26 09:25:51 vexpress-qemu rngd[1498]: Initializing available sources
    May 26 09:25:52 vexpress-qemu rngd[1498]: Failed to init entropy source hwrng
    May 26 09:26:30 vexpress-qemu rngd[1498]: Initializing AES buffer
    May 26 09:26:30 vexpress-qemu rngd[1498]: Enabling JITTER rng support
    May 26 09:26:30 vexpress-qemu rngd[1498]: Initializing entropy source jitter
    ```
    
    The `JITTER rng fails with code 2` error appears to be because the
    Jitter RNG driver returns ECOARSETIME, indicating that the timer is
    not accurate enough for RNG purposes. It's possible that this is just
    an effect of the emulated CPU being busy with booting and slowing down
    time measurements. As mentioned it eventually succeeds once the system
    is calm.
    
    This probably has existed for quite some time, and previously this was
    fixed in 6d705f1309bca by reverting to OpenSSL 1.0, which presumably
    requires less entropy and therefore doesn't starve the RNG pool.
    However, OpenSSL 1.0 is not available anymore, and we need rngd to
    work to generate new entropy.
    
    The fix is to instruct systemd to automatically restart rngd when it
    fails.
    
    Changelog: None
    
commit 23ff892e7c167ed5396e3d9a7950739abcf9c178
Date:   Tue May 26 11:28:14 2020

    vexpress-qemu: Fix warnings for unknown settings while building kernel.
    
    The removed entries were just taken straight from the log output,
    which says that the options no longer exist in the kernel
    configuration.
    
    Changelog: None
    
```


meta-mender-qemu/recipes-support/rng-tools/files/restart-rngd.diff
```diff
@@ -0,0 +1,11 @@
+--- a/rngd.service	2020-05-26 10:42:14.575073761 +0200
++++ b/rngd.service	2020-05-26 10:42:34.110746120 +0200
+@@ -8,6 +8,8 @@
+ [Service]
+ EnvironmentFile=-@SYSCONFDIR@/default/rng-tools
+ ExecStart=@SBINDIR@/rngd -f $EXTRA_ARGS
++Restart=on-failure
++RestartSec=5
```
lluiscampos, line 8: I didn't know systemd allowed this kind of restart. Nice finding.

kacf:
> Should we report upstream also?

The problem is that I don't know exactly what to report., This is specific to vexpress-qemu, which is a machine type that we created, and I have not seen it anywhere else. If I knew the underlying cause, perhaps we could, but without it, I guess this is Our Problem™.

kacf:
Thanks! This led to the first ever green build of zeus: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/149652930

PR #992, opened 2020-05-26
Add Raspberry Pi 4 as a build-tested platform.

kacf:
Changelog: None


lluiscampos:
Good. Are we adding it also to `mender-qa`, then?

lluiscampos:
> Good. Are we adding it also to `mender-qa`, then?

Ouch. See https://github.com/mendersoftware/mender-qa/pull/400

PR #996, opened 2020-06-02
[acceptance tests] Do not create empty "coverage" dir

lluiscampos:
It the target did not generate coverage reports, skip creating locally a
"coverage" dir so that the CI does not expect files there.


lluiscampos:
Stopping the build. To be tested/merged together with https://github.com/mendersoftware/mender/pull/531.

PR #1001, opened 2020-06-05
Update the mender-client license checksum

merlin-northern:
ChangeLog:none


kacf:
Already fixed in #1000.

PR #1002, opened 2020-06-05
zeus: MEN-3513: Work around unstable checksum when using 64bit ext4 feature.

kacf:
```
commit 4b40efb9b7d6eee40857ecd9e7d8c91923d4f1e8
Date:   Fri Jun 5 11:39:47 2020

    MEN-3513: Test that R/O rootfs checksum is correct also at runtime.
    
    Changelog: None
    
commit 401c6d8b4562af7ae79fd53d0ad778b0b797d895
Date:   Fri Jun 5 11:26:07 2020

    MEN-3513: Work around unstable checksum when using 64bit ext4 feature.
    
    See the ticket for details. The underlying checksum on disk doesn't
    actually change, but it changes in memory, and it doesn't seem to be
    possible to read the correct checksum out of the block device. Since
    we need this for mender-binary-delta to work, just disable the 64bit
    feature.
    
    Changelog: Disable `64bit` ext4 filesystem feature.
    
```


tests/acceptance/helpers.py
```diff
@@ -0,0 +1,143 @@
...
+
+import os
+import re
+import subprocess
+import requests
+
+from common import *
+
+
+class Helpers:
```
lluiscampos, line 24: This seems to be the exact same code, right?

tests/acceptance/commercial/test_delta_update_module.py
```diff
@@ -51,3 +52,77 @@ def test_build_and_run_module(
...
+                % bitbake_variables["LAYERDIR_MENDER"]
+            ],
+        )
+
+        image = latest_build_artifact(
+            prepared_test_build["build_dir"], "core-image*.mender"
+        )
+
+        try:
+            os.symlink(image, "temp-artifact.mender")
```
lluiscampos, line 97: Why do we do this, can't we install `image`? It has no implications, but after the update we will be doing reading back the original image (instead of the symlink).

lluiscampos: Okey, I guess is because of this comment in `install_update`:
```
# Note: `image` needs to be in current directory.
```
A cleaner solution can be to do this symlink inside the helper method.

kacf: Good idea, I'll try that.

tests/acceptance/helpers.py
kacf: Yep, no change at all!

tests/acceptance/commercial/test_delta_update_module.py
kacf: Since I didn't want to modify the code that I moved in the same commit, and I had already made a merge with master in between this PR and the review, I decided to merge this and make a followup PR instead: https://github.com/mendersoftware/meta-mender/pull/1004

PR #1006, opened 2020-06-12
Upgrade master to zeus

kacf:


kacf:
Stopping mender-qa build. Zeus has already been tested, and there are no new commits in master that are not already in zeus.

PR #1007, opened 2020-06-15
add MENDER_DATA_PART_MOUNT_LOCATION variable

coreycothrum:
Gauging interests in this. If this is a change we'd like to pursue, it'd be best to replace "/data" references in other files w/ the new variable.

I think it'd be cleaner to have this defined as a variable, instead of being hardcoded all over the place. I think it is most advantageous for user recipes that install something into "/data". 

Obviously it is pretty unlikely that "/data" would ever change locations, and hardcoding is likely to never be an issue.

Signed-off-by: Corey Cothrum <contact@coreycothrum.com>


kacf:
Fully agree, better to have this as a variable. As you say, changing the data location may be difficult for a variety of reasons, but you gotta start somewhere, and this is definitely cleaner.

coreycothrum:
> Fully agree, better to have this as a variable. As you say, changing the data location may be difficult for a variety of reasons, but you gotta start somewhere, and this is definitely cleaner.

So how far do we need/want to take it before merging this PR? Just meta-mender-core? Just the classes? Classes and recipes? Etc.

kacf:
If we change it, we should change it everywhere in meta-mender. The only possible exception would be in the tests, since they will always run with `/data`, but even here I would consider changing it unless it's difficult.

PR #1008, opened 2020-06-15
uboot_auto_configure: build U-Boot the same way Yocto does

twoerner:
The uboot_auto_configure.sh script uses the cross compiler to build
cmd/version.o from the U-Boot sources so it can generate a list of files to
patch. However, this will fail if the target board (and therefore the target
toolchain) uses floating point flags (e.g. armv5tehf-vfp) since U-Boot is not
meant to be built with floating-point.

The Yocto recipes invoke the cross toolchain when building U-Boot in a way
so that any floating-point build flags are not passed to the build. Tweak
the way the cross toolchain is invoked by following the way Yocto invokes
it.

Signed-off-by: Trevor Woerner <twoerner@gmail.com>

### DIFF
--- a/meta-mender-core/recipes-bsp/u-boot/files/uboot_auto_configure.sh
+++ b/meta-mender-core/recipes-bsp/u-boot/files/uboot_auto_configure.sh
@@ -4,7 +4,7 @@ set -e
 
 BUILD_AR="${BUILD_AR:-ar}"
 BUILD_CC="${BUILD_CC:-gcc}"
-CC=${CC:-${CROSS_COMPILE}gcc}
+CC="${CROSS_COMPILE}gcc"
 MAKE="${MAKE:-make}"
 MAKEFLAGS="${MAKEFLAGS:-}"
 MAYBE_UBI=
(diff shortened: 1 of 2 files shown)


twoerner:
@kacf 

> However, can you add a `Changelog: Title` tag to your commit message to get the changelog generated correctly?

Okay, done.

PS could this please be applied to warrior too (which is where I need it) or should I push a new commit?

kacf:
There was an (unrelated) error in test running, I am resubmitting it in #1012.

twoerner:
ping?
ready to apply? and to the warrior branch too? ;-)

kacf:
It is ready, but I ran into some unrelated problems in our test framework, which affects all U-Boot builds. Don't worry, I will include this, I just have to untangle that mess first. Stay tuned.

twoerner:
Could I get this cherry-picked onto warrior as well?
Or would a separate pull request be better?

kacf:
There is a cherry-picking round planned for warrior as soon as the 2.3.0 Mender client is out (being tested as we speak). The patch is already in zeus.

twoerner:
Perfect! Thanks for the update

PR #1009, opened 2020-06-16
Move default alignment for Raspberry Pi to layer settings.

kacf:
This is because the the patch now requires this information, unlike
previously.

Changelog: Fix error message about `CONFIG_ENV_OFFSET` being wrong,
such as:
```
ERROR: u-boot-fw-utils-mender-auto-provided-1.0-r0 do_configure: U-Boot configuration rpi_4_config has setting:
CONFIG_ENV_OFFSET=0x400000
CONFIG_ENV_OFFSET_REDUND=0x800000
but Mender expects:
CONFIG_ENV_OFFSET=0x800000
Please fix U-Boot’s configuration file.
```


mirzak:
Can you elaborate on why this is needed? Why does it not work in `local.conf`? 

Would really like us to avoid adding stuff to `layer.conf`

drewmoseley:
Yeah, that is required if we want to be officially labeled as "Yocto compatible".

kacf:
This is related to [a discussion we had a while ago](https://github.com/mendersoftware/meta-mender/pull/975#discussion_r414354490): The `MENDER_PARTITION_ALIGNMENT` has to match the single and the double value of `CONFIG_ENV_OFFSET` and `CONFIG_ENV_OFFSET_REDUND`, respectively. When not using auto-patching, the values will no longer be auto-fixed for you, the main reason being that this is much more intrusive than it used to be. So without this patch, the Raspberry Pi build is broken, because the our patch doesn't match what is in `MENDER_PARTITION_ALIGNMENT`.

Originally I thought that we documented that you should set `MENDER_PARTITION_ALIGNMENT = "4194304"` for Raspberry Pi, but looking at it now, I think I am wrong. And in that case I guess people have just been using the default, which is `8388608`. So maybe a better patch would be to adjust the Raspberry Pi patch.

kacf:
@mirzak, @drewmoseley: So how about this patch instead?

PR #1013, opened 2020-06-19
Dunfell fixes

kacf:


kacf:
Thanks @mirzak, I added [MEN-3703](https://tracker.mender.io/browse/MEN-3703) to track this, and will handle it as a followup.

kacf:
I had to make a couple of extra build fixes. Pretty obvious/trivial stuff, so no need to re-review unless you feel like it. Also, the patch I added has been submitted upstream and reviewed separately by Stefano Babic on the swupdate mailing list.

mirzak:
Unrelated to current changes, but I think we should also drop the 2020.01 recipe we have here (that should be the same in dunfell upstream?),

https://github.com/mendersoftware/meta-mender/tree/master/meta-mender-core/recipes-bsp/u-boot

kacf:
Indeed, thanks for reminding me, I'll make a note of that too!

PR #1020, opened 2020-06-25
Add MENDER_EXTRA_PARTS_SIZES_MB variable

kacf:
Add MENDER_EXTRA_PARTS_SIZES_MB variable to store partition size(s) of
EXTRA_PARTS.

These values are used to correct solution of
mender_calculate_rootfs_size_kb()

wks will be auto-populated with the --fixed-size option

Backwards compatibility should be preserved, i.e. omitting a value for
this new variable should behave as before

Changelog: Title

Signed-off-by: Corey Cothrum <contact@coreycothrum.com>


oleorhagen:
Related: https://github.com/mendersoftware/mender-docs/pull/1031

PR #1030, opened 2020-07-08
More Dunfell fixes

kacf:
See individual commits.


kacf:
Never mind, this won't work without a valid client recipe. This is now blocked until we release a new client.

PR #1037, opened 2020-07-23
[RFC] Zeus yocto check layer changes

drewmoseley:

### Description

Changes needed to allow this layer to pass the yocto-check-layers script which is required for the Yocto-compatible designation.


drewmoseley:
@dwalkes FYI.  I don't know Matt's nick so I can't mention him here.
Note that this is for Zeus.  I'm still testing and will also replicate this work on the master branch.

dwalkes:
Thanks @drewmoseley for sharing these.  I took a glance through and don't have specific suggestions but I've linked from the issue at https://github.com/madisongh/test-distro/issues/5 requesting feedback.

meta-mender-raspberrypi/recipes-bsp/u-boot/patches/0001-configs-rpi-enable-mender-requirements.patch
```diff
@@ -42,11 +42,11 @@ index 75c6c9c447..493a99a26d 100644
  CONFIG_OF_LIBFDT_OVERLAY=y
  CONFIG_EFI_LOADER=y
 +CONFIG_ENV_IS_IN_MMC=y
-+CONFIG_ENV_OFFSET=0x800000
-+CONFIG_ENV_OFFSET_REDUND=0x1000000
++CONFIG_ENV_OFFSET=0x400000
++CONFIG_ENV_OFFSET_REDUND=0x800000
```
kacf, line 46: This doesn't look correct to me, this was [just fixed the opposite way](https://github.com/mendersoftware/meta-mender/commit/3d366e21cec4b6e15faa4febcedc430cdd9eb546) because zeus would not work otherwise. What's the reason for this patch?

meta-mender-core/conf/layer.conf
```diff
@@ -15,7 +15,6 @@ BBFILE_COLLECTIONS += "mender"
 BBFILE_PATTERN_mender = "^${LAYERDIR}/"
 BBFILE_PRIORITY_mender = "6"
 
-INHERIT += "mender-setup"
```
kacf, line 18: This has an unintended side effect, namely that the `MENDER_FEATURES_ENABLE` and `MENDER_FEATURES_DISABLE` variables stop working. Those are supposed to function independently of `mender-full` & friends, exactly so that you can include only the things you need instead of relying on one of the predefined classes. There is also a lot of error checking in `mender-setup` which doesn't work correctly if this is removed and someone is selecting features one by one.

I think instead we should consider splitting this into "things that change the build", and "things that enable the layer to work". The latter would not influence the build, but would enable the use of `MENDER_FEATURES_ENABLE` and `MENDER_FEATURES_DISABLE`.

Out of curiosity, what exactly did you use to find out how the build changes after adding `meta-mender`? Is this something we can test for?

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -1,3 +1,5 @@
+OVERRIDES =. "mender-enabled:"
```
kacf, line 1: We are using `_mender-client-install` and other similar overrides in *a lot* of places already. Under what conditions exactly does it not work?

Assuming that that can be solved, I think we should not add the `mender-enabled` override, because everything should already fit into one of the variables added in `MENDER_FEATURES_ENABLE` and `MENDER_FEATURES_DISABLE`. Or else we should add more `MENDER_FEATURES` to cover the new cases, so that they can be enabled and disabled selectively.

meta-mender-core/recipes-bsp/grub/grub_%.bbappend
```diff
@@ -1,86 +1,9 @@
...
-addtask do_deploy after do_setcorelocation
+FILESEXTRAPATHS_prepend_mender-enabled := "${THISDIR}/files:"
+
+# Using this syntax allows the primary implementation code to be
+# conditionally included only when Mender is enabled. This allows
+# these layers to be included in a configuration without changing
+# the build unless specifically enabled.
+GRUB_MENDER_IMPL=""
+GRUB_MENDER_IMPL_mender-enabled="mender-grub-impl"
+inherit ${GRUB_MENDER_IMPL}
```
kacf, line 9: Clever!

meta-mender-qemu/recipes-bsp/u-boot/patches/0001-vexpress-qemu-specific-fixes-for-Flash.patch
```diff
@@ -1,23 +1,41 @@
...
++CONFIG_ENV_SIZE=0x20000
+ CONFIG_NR_DRAM_BANKS=2
+ CONFIG_ENV_SECT_SIZE=0x40000
+ CONFIG_DISTRO_DEFAULTS=y
+@@ -24,6 +24,8 @@ CONFIG_ENV_IS_IN_FLASH=y
+ CONFIG_ENV_ADDR=0x47F80000
+ CONFIG_MTD=y
+ CONFIG_MTD_NOR_FLASH=y
++CONFIG_MTDIDS_DEFAULT="nor2=40000000.flash"
++CONFIG_MTDPARTS_DEFAULT="mtdparts=40000000.flash:1m(u-boot)ro,-(ubi)"
```
kacf, line 34: The specific things that you are adding in this patch are the exact things we would like not to be added, so that the auto-patcher will do it for us. It's how we test the auto-patcher. The existing things in this patch are here because they are too difficult to the patcher to catch, or specific to our setup.

meta-mender-raspberrypi-demo/conf/layer.conf
```diff
@@ -13,6 +13,4 @@ BBFILE_PATTERN_mender-raspberrypi-demo = "^${LAYERDIR}/"
 
 LAYERDEPENDS_mender-raspberrypi-demo = "mender mender-demo mender-raspberrypi"
 
-INHERIT += "rpi-demo-wifi"
```
kacf, line 16: This is not re-added anywhere. Is it ok to just remove it?

meta-mender-qemu/recipes-kernel/linux/linux-yocto_%.bbappend
```diff
@@ -20,4 +20,4 @@ COMPATIBLE_MACHINE_vexpress-qemu-flash = "vexpress-qemu-flash"
 # at the time of writing. Rather than dig into this we will just remove this
 # patch for now. This can probably be removed later when upstream has fixed this
 # problem. (July 2018)
-KERNEL_FEATURES_remove_qemuall = "features/kernel-sample/kernel-sample.scc"
+KERNEL_FEATURES_remove_vexpress-qemu-flash = "features/kernel-sample/kernel-sample.scc"
```
kacf, line 23: I suspect this is needed for `vexpress-qemu` as well, but not 100% sure.

tests/meta-mender-beaglebone-yocto-ci/recipes-mender/mender-qa/mender-qa.bbappend
```diff
@@ -1,13 +1,15 @@
-PACKAGES += "rssh-tunnel"
-FILESEXTRAPATHS_prepend := "${THISDIR}/files/beagleboneblack/:${THISDIR}/files/:"
+PACKAGES_append_mender-client-enabled = " rssh-tunnel"
+FILESEXTRAPATHS_prepend_mender-client-enabled := "${THISDIR}/files/beagleboneblack/:${THISDIR}/files/:"
```
kacf, line 2: No one else will use these. Is it really necessary to add all these overrides for test layers? Do we have to qualify as Yocto compatible on a repository level, or on a layer level?

meta-mender-raspberrypi/recipes-bsp/u-boot/patches/0001-configs-rpi-enable-mender-requirements.patch
drewmoseley: At this point I am only doing build time tests so it's very possible this is wrong.  I'll resync my sources and see if this is no longer needed.

meta-mender-core/conf/layer.conf
drewmoseley: In poky/scripts there is a yocto-check-layer.sh script that does all the checks.

I'll see about creating a separate class that can be automatically included but it's kinda frowned upon to make such changes in layer.conf.

drewmoseley: Once we get this cleaned up and merged, I would definitely recommend adding CI testing for this.

meta-mender-core/classes/mender-setup.bbclass
drewmoseley: My only concern is, is it possible for mender-client-install _not_ to be set?

tests/meta-mender-beaglebone-yocto-ci/recipes-mender/mender-qa/mender-qa.bbappend
drewmoseley: I debated that and it's probably ok not to handle these layers.  I was even thinking of adding a mender-testing-enabled feature.  In my case, I try to keep all the layers in my normal config and simply enable/disable things in local.conf.  I try to avoid having multiple build directories and even started using a single build directory for multiple MACHINEs.  I can see some value in having the testing layers enabled in our debug builds but simply disabled in local.conf unless needed.

Your call though as I don't get into these layers much currently.

meta-mender-qemu/recipes-bsp/u-boot/patches/0001-vexpress-qemu-specific-fixes-for-Flash.patch
drewmoseley: I'm not sure what you are saying here. Should I not need to add these?  The current Zeus branch build of vexpress-qemu-flash fails with:
```
NOTE: Setscene tasks completed
ERROR: u-boot-fw-utils-1_2019.07-r0 do_configure: U-Boot configuration vexpress_ca9x4_config has setting:
CONFIG_MTDIDS_DEFAULT=""
but Mender expects:
CONFIG_MTDIDS_DEFAULT="nor2=40000000.flash"
Please fix U-Boot's configuration file.
ERROR: u-boot-fw-utils-1_2019.07-r0 do_configure: Execution of '/work/dmoseley/yocto-build/Mender/yocto/zeus/build/tmp/work/vexpress_qemu_flash-poky-linux-gnueabi/u-boot-fw-utils/1_2019.07-r0/temp/run.do_configure.54691' failed with exit code 1:
CONFIG_MTDIDS_DEFAULT=""

ERROR: Logfile of failure stored in: /work/dmoseley/yocto-build/Mender/yocto/zeus/build/tmp/work/vexpress_qemu_flash-poky-linux-gnueabi/u-boot-fw-utils/1_2019.07-r0/temp/log.do_configure.54691
ERROR: Task (/work/dmoseley/community/yocto/zeus/src/poky/meta/recipes-bsp/u-boot/u-boot-fw-utils_2019.07.bb:do_configure) failed with exit code '1'
```

drewmoseley: NM.  It turns out that [this line](https://github.com/mendersoftware/meta-mender/blob/64baa4815e03b89a100976c6e171733a579f1eab/meta-mender-raspberrypi/recipes-bsp/u-boot/u-boot-raspberrypi.inc#L7) was disabling autoconfig for _all_ targets rather than just RPI.  I've got it fixed locally and will push a new version soon.

meta-mender-core/classes/mender-setup.bbclass
kacf: Honestly, probably not. Would it be a problem though?

The intention to be able to disable it is there, and the idea was that this could be used if we added some product besides Mender that could be used independently (SWUpdate talking directly to Mender Server, CFEngine, etc.). Ultimately we may end up never using it for that purpose, but I still think it's better to use it than to introduce yet another one.

tests/meta-mender-beaglebone-yocto-ci/recipes-mender/mender-qa/mender-qa.bbappend
kacf: That's a good point, I guess that's a good way to experience first hand whether the layer is making intrusive changes or not! I'm ok with leaving it as is, except possibly changing the name of the override, as discussed elsewhere.

meta-mender-core/classes/mender-setup.bbclass
drewmoseley: OK.  I'll clean that up.

tests/meta-mender-beaglebone-yocto-ci/recipes-mender/mender-qa/mender-qa.bbappend
drewmoseley: Should I create a mender-testing-enabled override specifically for these settings?

meta-mender-core/classes/mender-setup.bbclass
drewmoseley: Now I remember why I had to add that.  Since the Mender feature overrides are processed _later_ by anonymous python, they are usable for things such as the _impl bbclass selection.  I needed something that was always in the OVERRIDES list.

tests/meta-mender-beaglebone-yocto-ci/recipes-mender/mender-qa/mender-qa.bbappend
drewmoseley: Looks like I tried to do that but just missed it in a couple of places.

meta-mender-core/conf/layer.conf
kacf: I took a look at `mender-setup.bbclass`. For the most part, it is already supposed to have no effect on the build, other than providing default values for variables, and setting things up so that you can enable them. Same goes for all of the `mender-setup*` files.

I did spot some exceptions though, like:
```
IMAGE_INSTALL_append = " mender-client"
```

This should be changed to:
```
IMAGE_INSTALL_append_mender-client-install = " mender-client"
```

So I think we need to make a choice regarding `mender-setup.bbclass` and the other `mender-setup*` files. I'm proposing one of these:

1. Restrict `mender-setup` to do `MENDER_FEATURES` evaluation, and to set default values for `MENDER_` variables and `mender_` functions that are not expected to be used outside of Mender. This is probably not a huge change, but will require fixing the odd exception here and there, like the one I mentioned above. Downside is that many variables are still evaluated, but they are unlikely to have any direct effect.

2. Restrict `mender-setup` to do `MENDER_FEATURES` evaluation only. This means we would do something like:
   ```
   mender-setup.bbclass:
   ---------------------

   # Code for MENDER_FEATURES evaluation. At minimum it is these three:

   # ------------------------------------------------------------
   DISTRO_FEATURES_BACKFILL_append = " ${MENDER_FEATURES_ENABLE}"
   DISTRO_FEATURES_BACKFILL_CONSIDERED_append = " ${MENDER_FEATURES_DISABLE}"

   python() {
       # Add all possible Mender features here. This list is here to have an
       # ...

   def mender_feature_is_enabled(feature, if_true, if_false, d):
       # ...
   # ------------------------------------------------------------

   # And then, to get the variables, we could do:
   MENDER_SETUP_VARIABLES_CLASS = "${@mender_feature_is_enabled('mender-client-install', 'mender-setup-variables', '', d)}"
   inherit ${MENDER_SETUP_VARIABLES_CLASS}
   ```
   `mender-setup-variables.bbclass` would then contain most of what is in `mender-setup.bbclass` now. I guess this is not a huge amount of work either, mostly just shuffling things around.

3. Don't include `mender-setup` at all, thereby breaking `MENDER_FEATURES` evaluation, and require manual inclusion of `mender-setup` in every build. This has a definite downside, namely that it cannot be added to zeus, as this would break existing builds that select features one by one.

I suppose that number 3 is the cleanest of the three, but it is also the most intrusive. I think I would vote for number 2, as it seems relatively straightforward to implement, is still fairly clean (only `MENDER_FEATURES` are evaluated OOTB), and IMHO provides the best compromise between staying compatible, being clean and providing an easy path to enable features.

tests/meta-mender-beaglebone-yocto-ci/recipes-mender/mender-qa/mender-qa.bbappend
kacf: I think yes. But this requires a change in the build scripts too, to [add the feature when the testing layers are being used](https://github.com/mendersoftware/mender-qa/blob/2399567768f42186ab3cdfb28f442fe82ab06e2c/scripts/jenkins-yoctobuild-build.sh#L554).

meta-mender-core/conf/layer.conf
drewmoseley: Is #3 viable for master? As you say it is the cleanest and I don't think an unrealistic expectation.

meta-mender-core/classes/mender-setup.bbclass
drewmoseley: I seem to recall having some issues with python in those variables not resolving correctly, but I'll give it a shot.

meta-mender-raspberrypi-demo/conf/layer.conf
drewmoseley: This is similar to the question of automatic inclusion of mender-setup.  Just adding the layer should not explicitly add anything so it's generally up to the user to add the settings to local.conf.  Not sure how to handle this.

Question: is anyone using this layer?  Can it just be deprecated?

meta-mender-core/conf/layer.conf
kacf: Yes, it is viable for master. Note that we released dunfell yesterday, so strictly speaking it is not viable, but if we act quickly, I suppose we can sneak it into the next tag. It's already slightly broken because of the meta-raspberrypi issue.

meta-mender-core/classes/mender-setup.bbclass
kacf: I tested this, and it works.

meta-mender-raspberrypi-demo/conf/layer.conf
kacf: We use it in our Yocto Raspberry Pi image, but now our onboarding is based on the image from mender-convert, so we no longer depend on it as much. However, there have been several contributions to this layer, so I assume people are using it when they test Raspberry Pi with Yocto.

mirzak: I think we should remove it (to me this layer has always been a bit out of place). 90 % of this layer is generic actually and we could merge these parts in meta-mender-demo, to provide an easy way to configure WiFi.

The Raspberry Pi specific parts are to me a workaround to a bug at some point in meta-raspberrypi and should not be necessary. `wifi` is a default distro feature and as long as you build an image that is derived from `core-image-base` you should have everything in place for wifi to work, otherwise there is a bug in meta-raspberrypi

meta-mender-core/conf/layer.conf
drewmoseley: Actually, looking further, we need at least the mender_feature_is_enabled function automatically added to avoid parse failures.  I _think_ we have to go with #2 because of that.  I'll experiment and post a confirmation later today.

drewmoseley:
@kacf I pushed a number of new commits to address the comment threads above.  For now I have simply added new commits to make it easier to review only what has changed since the last time.  I'll rearrange, squash, etc before we merge this.

I'm still testing but with this latest set of changes, all the layers in meta-mender pass the yocto-check-layer script.

meta-mender-raspberrypi-demo/conf/layer.conf
drewmoseley: Yeah, I just took a look through and this does not need to be separate from meta-mender-demo.  I'll tack that onto the end of this PR.

meta-mender-qemu/recipes-kernel/linux/linux-yocto_%.bbappend
drewmoseley: How do I test this?

drewmoseley:
Nope.  Still not working.  DISTRO_FEATURES are killing me.

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -211,11 +211,16 @@ IMAGE_CLASSES += "mender-part-images mender-ubimg mender-artifactimg mender-data
...
+def mender_feature_is_enabled(feature, if_true, if_false, d):
+    in_enable = bb.utils.contains('MENDER_FEATURES_ENABLE', feature, True, False, d)
+    in_disable = bb.utils.contains('MENDER_FEATURES_DISABLE', feature, True, False, d)
+
+    if in_enable and not in_disable:
+        return if_true
+    else:
+        return if_false
+
+inherit ${@mender_feature_is_enabled('mender-client-install', 'mender-setup-distro-features', '', d)}
```
kacf, line 223: I don't think you need this complicated workaround. Check out my proposed fix in https://github.com/drewmoseley/meta-mender/pull/2.

meta-mender-core/classes/mender-setup.bbclass
drewmoseley: Yeah, I think it might have been when I was trying to use the _mender-* overrides that it didn't work.  It does seem to be working in my latest branch.

meta-mender-raspberrypi/recipes-bsp/u-boot/u-boot-raspberrypi.inc
```diff
@@ -4,7 +4,7 @@ BOOTENV_SIZE_rpi ?= "0x4000"
 
 # U-Boot auto-configuration doesn't work properly for Raspberry Pi, due to its
 # use of boot.scr and cmdline.txt.
-MENDER_UBOOT_AUTO_CONFIGURE ?= "0"
+MENDER_UBOOT_AUTO_CONFIGURE_rpi ?= "0"
```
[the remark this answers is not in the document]
drewmoseley: I think I'll go with the easy way.  As you say this is not something that can be changed for RPI.

meta-mender-core/classes/mender-setup.bbclass
drewmoseley: Testing that now.

drewmoseley:
OK, even with https://github.com/drewmoseley/meta-mender/pull/2 this does not work.  The MENDER_FEATURES_DISABLE values are not actually disabled.  I am afraid we have a circular dependency that we cannot resolve. The combination of OVERRIDES, MENDER_FEATURES_(EN|DIS)ABLE, and DISTRO_FEATURES just is resistant to proper ordering.  Any time I clean up one of them, the others are broken.

I'm starting to think we have no choice but to explicitly remove mender-setup.bbclass from layer.conf and require that users explicitly add it when they want to build with Mender

kacf:
Finally getting back to this!

Can you give me the `local.conf` and `bblayers.conf` for a failing build? I have tried lots of combinations of enabling/disabling features, removing and adding `mender-full` class, and all combinations work as expected for me, including those used in `inherit` statements. I'm not able to reproduce the problems you're having.

drewmoseley:
I've reworked this a bit and have another version.  I'm still testing and hopefully will have something to post soon.

drewmoseley:
@kacf  I updated this with my latest attempt.  I think it's cleaner and more obvious what and why.
I still need more testing though.  All the layers pass yocto-check-layers with the qemu machines but I know we have some things that will fail with other MACHINE types.  That's the next check on my list.

kacf (changes requested):
Thanks for sticking with this! Code wise this looks good to me now, just make sure to fix those "ci/gitlab" errors.

FYI, I started a [test build](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/181174176) using the required [mender-qa fix](https://github.com/mendersoftware/mender-qa/pull/416). I also used 2.4.0 for the testing instead of the Mender master components, since they have some unrelated checksum errors.

kacf:
I see some errors [building the client](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/699958581):

```
install: cannot stat '/builds/Northern.tech/Mender/build-vexpress-qemu-flash/tmp/work/armv5-poky-linux-gnueabi/mender-client/2.3.0-r0/mender-client-data-dir.service': No such file or directory
```

 Does anything ring a bell?

lluiscampos:
> I see some errors [building the client](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/699958581):
> 
> ```
> install: cannot stat '/builds/Northern.tech/Mender/build-vexpress-qemu-flash/tmp/work/armv5-poky-linux-gnueabi/mender-client/2.3.0-r0/mender-client-data-dir.service': No such file or directory
> ```
> 
> Does anything ring a bell?

Not really. I've been looking into this a bit but can't see anything.

It is very strange that only fails this way for `vexpress_qemu_flash` and `rasperrypi_3/4`. (The other jobs failed later on in testing for a different reason).

kacf:
I also cannot spot where the problems come from. @drewmoseley, I think you just have to try those configurations for yourself. I didn't look at all the jobs, but at the very least, these things seem to fail:

* Raspberry Pi 4 build
* x86 qemu build when `PREFERRED_VERSION_pn-mender-client = "2.2.1"`

lluiscampos:
>     * x86 qemu build when `PREFERRED_VERSION_pn-mender-client = "2.2.1"`

@kacf This one I run locally, using this PR as my `meta-mender` and `zeus` branch for `poky` and `meta-openembedded`. It built just fine for me...

kacf:
Hmm, ok. If Drew doesn't find anything tonight, I can try to run the test tomorrow.

drewmoseley:
Sorry I didn't have time to do much with this yet.  I cannot reproduce any of the mentioned issues in my local build but my setup is different from what is used in the CI setup so that's not necessarily indicative of anything specific.

kacf:
I think I fixed the failure in `test_preferred_versions`, and pushed a new commit. I will try another test round.

kacf:
Regarding this error:

```
install: cannot stat '/builds/Northern.tech/Mender/build-raspberrypi4/tmp/work/cortexa7t2hf-neon-vfpv4-poky-linux-gnueabi/mender-client/2.3.0-r0/mender-client-data-dir.service': No such file or directory
```

in Raspberry Pi 4 and a few others, I have made some progress, but haven't understood the whole picture yet. It happens when `EXTERNALSRC_pn-mender-client` is enabled. With this I'm able to reproduce it locally, and what happens is that `SRC_URI` somehow ends up empty. At first I thought it was the same issue as earlier, some evaluation issue because class selectors have been added. But this doesn't seem to be the case, because even the original code has class selectors in this area, and there it works fine. It needs more investigation, but I'll have to continue tomorrow due to other duties.

drewmoseley:
@kacf is this far enough along that I should move it over to dunfell/master now?

kacf:
Yes, I believe you can. I think the last remaining issue is not significantly affected by which branch it is, and I'll try to work on that separately.

Make sure you include the extra commit that I pushed here a few days ago.

kacf:
Hopefully I've fixed the last issue now, check out the "Make mender-features and OVERRIDES handling always run." commit. It should still keep the layer clean unless Mender features are enabled, but you might want to double check it @drewmoseley, just to be sure.

drewmoseley:
Dang.  The commit https://github.com/mendersoftware/meta-mender/pull/1037/commits/60819d86dbffd63c3255ad8587bd6c515fe3821c breaks the check-layer:

```
      basehash changed from 61ca84603566c2fd33e165b4db0e6331900c2c44d2c4005bf9ea2c38391d4c48 to bf55cf9426712cb59c50d3258b483ec5c3ba1bb50d8664e1b6154fc78e3beb0e
      List of dependencies for variable PACKAGECONFIG changed from '{'PACKAGECONFIG[grub]', 'PACKAGECONFIG[modules]', 'PACKAGECONFIG[mender-client-install]', 'PACKAGECONFIG[split-mender-config]', 'PACKAGECONFIG[u-boot]'}' to '{'PACKAGECONFIG[grub]', 'mender_feature_is_enabled', 'PACKAGECONFIG[modules]', 'PACKAGECONFIG[mender-client-install]', 'PACKAGECONFIG[split-mender-config]', 'PACKAGECONFIG[u-boot]'}'
      changed items: {'mender_feature_is_enabled'}
      Dependency on variable mender_feature_is_enabled was added
      Dependency on variable MENDER_FEATURES_DISABLE was added
      Dependency on variable MENDER_FEATURES_ENABLE was added
      Variable PACKAGECONFIG value changed:
      @@ -1,4 +1,4 @@

```

drewmoseley:
But maybe that doesn't matter.  Didn't we decide we were going to enable modules by default?  Can we just remove that from the meta-mender-demo layer?

drewmoseley:
Actually it looks like checking DISTRO_FEATURES directly instead of using mender_feature_is_enabled works.  And that's how it is implemented in meta-mender-core so I think maybe we are good to go with that changed.  I'll push an update to test.

drewmoseley:
OK.  I pushed a commit to move the module PACKAGECONFIG into meta-mender-core.  That gets us past the check-layer failure and with any luck will get us a clean build here.  I'll start cherry-picking to dunfell/master.

kacf:
Yes, that we can do, but since it changes the default installation configuration, I was not planning to add that to any branch before yocto-3.3 (post-dunfell). But let's just try to get this into master first, and we can worry about the details for dunfell later. Will you resubmit this to master?

drewmoseley:
Yes, I have a master branch version of this in progress now.  I'm fixing a few build issues and will hopefully be able to post it today.

kacf:
Closing this as we are not planning to port this to zeus. See #1110 for dunfell, however.

PR #1039, opened 2020-07-27
Fix broken `set-time-epoch` feature in dunfell.

kacf:
On dunfell it is apparently not enough anymore to just specify
`set-time-epoch` to force systemd to use 1st Jan 1970 as epoch. It is
apparently possible to to use any epoch you like, and the default is
the build time, so we need to specifically set it to zero. See the
comment above the change for context on why the epoch is needed in the
first place.

Changelog: Fix fsck running on every boot in dunfell.

### Description

Thank you!

### DIFF
--- a/meta-mender-core/recipes-core/systemd/systemd_%.bbappend
+++ b/meta-mender-core/recipes-core/systemd/systemd_%.bbappend
@@ -4,4 +4,5 @@
 # in the future and an fsck will be done.  Setting this to 0 results
 # in an epoch of January 1, 1970 which is detected as an invalid time
 # and the fsck will be skipped.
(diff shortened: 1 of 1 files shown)


mirzak:
The change looks good to me.

Below is mostly an FYI,

I do not know if you had the opportunity to dig a bit deeper in to this, but I would expect it to be `SOURCE_DATE_EPOCH = 0` by default if you are using `poky` as distribution. This relates to reproducible builds which should be enabled by default on dunfell:

From dunfell changelog:
```
Enabled reproducible builds by default in poky.conf
```

We want this change to apply even if you are not using `poky` distribution so we should still proceed.

kacf:
I'm not sure I understood everything correctly from their handling of `SOURCE_DATE_EPOCH` (there is *a lot* of logic surrounding it), but my impression is that the reproducible build they are talking about actually means reproducible by using the latest timestamp from the source files. Stable, but not zero.

PR #1047, opened 2020-07-28
Adapt to name change in meta-raspberrypi + merge with master

kacf:
To reviewers, it is https://github.com/mendersoftware/meta-mender/commit/ce88ec9b92afb9ff6311f459e98bf189b3e6e2b1 which is the new commit here.


kacf:
@drewmoseley, @mirzak: I'm not going to make a new tag for dunfell, since meta-raspberrypi managed to break dunfell on the same day that we released it. But just FYI, if users on Mender Hub are getting failures that "bcm2835-bootfiles" is not found, they need to check out the dunfell **branch**, if they are getting failures that "bootfiles" is not found, they need to check out the dunfell-v2020.07 **tag**.

PR #1052, opened 2020-08-04
MEN-3730 OpenSSL: set SECLEVEL=2 in /etc/ssl/openssl.cnf

merlin-northern:
_Note_: goes with [mender/pull/581](https://github.com/mendersoftware/mender/pull/581)

ChangeLog:none

### DIFF
--- a/meta-mender-qemu/recipes-connectivity/openssl/openssl_%.bbappend
+++ b/meta-mender-qemu/recipes-connectivity/openssl/openssl_%.bbappend
@@ -1,0 +1,9 @@
+
+do_install_append () {
+	# see MEN-3730, this is to align with the current default settings on raspbian
+	cat >> "${D}${sysconfdir}/ssl/openssl.cnf" <<EOF
+[system_default_sect]
+MinProtocol = TLSv1.2
+CipherString = DEFAULT@SECLEVEL=2
+EOF
+}


meta-mender-qemu/recipes-mender/mender-client/mender-client-test-files.inc
```diff
@@ -7,6 +7,14 @@ do_compile_append_mender-image_qemuall() {
         cat > ${WORKDIR}/rootfs-image-v2.conf <<EOF
 MENDER_ROOTFS_PART_A="${MENDER_ROOTFS_PART_A}"
 MENDER_ROOTFS_PART_B="${MENDER_ROOTFS_PART_B}"
+EOF
+    fi
+
+    if [ -d "${WORKDIR}/etc/ssl" ]; then
+        cat > "${WORKDIR}/etc/ssl/openssl.cnf" <<EOF
```
merlin-northern, line 14: oh, crap. @lluiscampos you are right, it should be '>>' of course :)

meta-mender-qemu/recipes-mender/mender-client/mender-client-test-files.inc
```diff
@@ -7,6 +7,14 @@ do_compile_append_mender-image_qemuall() {
         cat > ${WORKDIR}/rootfs-image-v2.conf <<EOF
 MENDER_ROOTFS_PART_A="${MENDER_ROOTFS_PART_A}"
 MENDER_ROOTFS_PART_B="${MENDER_ROOTFS_PART_B}"
+EOF
+    fi
+
+    if [ -d "${WORKDIR}/etc/ssl" ]; then
```
alfrunes, line 13: I don't think we should hard-code this anywhere, but rather make the client implementation resilient to openssl parameters if possible.

meta-mender-qemu/recipes-mender/mender-client/mender-client-test-files.inc
lluiscampos, line 14: To append to the file, we need `>>`

meta-mender-qemu/recipes-mender/mender-client/mender-client-test-files.inc
lluiscampos: @alfrunes I had a discussion with @merlin-northern and we think that making qemu match raspbian is the least we could do (i.e. have SECLEVEL=2 in both places).

If we want to test for multiple configurations, we need to separate this change to a separate layer, and then add it conditionally via fixture for the acceptance tests. Is it worth it? (If so I can take over or mentor @merlin-northern on how to do this).

lluiscampos:
Also, once we have this in good shape, you need to launch a build together with https://github.com/mendersoftware/mender/pull/581

meta-mender-qemu/recipes-mender/mender-client/mender-client-test-files.inc
merlin-northern: > I don't think we should hard-code this anywhere, but rather make the client implementation resilient to openssl parameters if possible.

client code is one thing (we can do it, check and support other levels, but this is only a discussion about tests! since all will work, if you conform with your settings in your `openssl.cnf`, if you change them, you knew what you were doing and you have to align the keys lengths for instance). the thing here is: to make qemu image reflect what we build and ship with raspbian. and by default we ship `seclevel=2`.

merlin-northern:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/174599230

meta-mender-qemu/recipes-connectivity/openssl/openssl_%.bbappend
```diff
@@ -0,0 +1,17 @@
...
+		if [ "${level_set}" == "" ]; then
+				# see MEN-3730, this is to align with the current default settings on raspbian
+				cat >> "${cnf}" <<EOF
+[system_default_sect]
+MinProtocol = TLSv1.2
+CipherString = DEFAULT@SECLEVEL=2
+EOF
+		fi
+	fi
+}
```
lluiscampos, line 17: We know exactly the contents of `openssl.cnf` from upstream, so we don't need to check if the file exists nor if SECLEVEL is already set. And actually, we want to make sure that this `do_install_append` breaks if the file ever gets moved, for example (as it is now I would silently pass and do nothing). That is the best practice in yocto.

Therefore I suggest a simple:

```suggestion
do_install_append () {
	# see MEN-3730, this is to align with the default settings on Raspbian
	cat >> "${D}${sysconfdir}/ssl/openssl.cnf" <<EOF
[system_default_sect]
MinProtocol = TLSv1.2
CipherString = DEFAULT@SECLEVEL=2
EOF
}
```

merlin-northern:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/175142481

PR #1064, opened 2020-08-20
Install standard Mender Update Modules by default.

kacf:
The default is changed in order to have a better chance to fix
problems on devices, when the rootfs-image setup us not working, for
example by using the script module.

Changelog: Install standard Mender Update Modules by default. These
are:
* deb
* directory
* docker
* rpm
* script
* single-file
To disable the automatic installation of these Update Modules, add
`PACKAGECONFIG_remove = "modules"` to your `mender_%.bbappend` file,
or set `PACKAGECONFIG` explicitly.

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client.inc
@@ -92,6 +92,8 @@ INSANE_SKIP_${PN}-ptest = "ldflags textrel"
 
 GO_IMPORT = "github.com/mendersoftware/mender"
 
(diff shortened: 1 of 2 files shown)


kacf:
@estenberg, @drewmoseley, @merlin-northern: As discussed in client team meeting.

merlin-northern (approved):
I am all for it!

lluiscampos (approved):
I don't think makes much sense to install `deb` and `rpm` modules. However from `meta-mender` point of view, makes sense to just install "whatever modules are supported OOTB", so this PR is okey. Maybe something to consider when adding new modules.

PR #1057, opened 2020-08-07
Wip zeus

drewmoseley:
# External Contributor Checklist

🚨 Please review the [guidelines for contributing](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md) to this repository.

- [ ] Make sure that all commits have a [`Changelog`](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#changelog-tags) tag. If nothing should be added to the Changelog, add a `Changelog: None` tag. If there is a change, add `Changelog: Commit`, or `Changelog: Title`, depending on what should be included in the changelog.

- [ ] Make sure that all commits are signed with [`git --signoff`](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work). Also note that the signoff author must match the author of the commit.

### Description

Various fixes from my recent projects.
These are not high priority but wanted to get them submitted on the branches I use for review.

PR #1066, opened 2020-08-25
[zeus] vexpress: Remove nonexistent kernel config options that issue warnings.

lluiscampos:
Commit missing from #1057


lluiscampos:
@drewmoseley While reviewing zeus/warrior I saw that this commit was missing from your original PR. Was it intentional?

lluiscampos:
Build @ [Pipeline 181731743](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/181731743)

kacf (approved):

lluiscampos:
Interesting enough, the acceptance tests fail and timeout after 8h for vexpress jobs (I retried the run, the failure is consistent). I will be looking more into this later.

lluiscampos:
> Not intentional but I'm not sure I saw the warnings on Zeus.

Not that I have seen the warnings neither, but the patch was ported to warrior so that was surprising to me.

> Interesting enough, the acceptance tests fail and timeout after 8h for vexpress jobs (I retried the run, the failure is consistent). I will be looking more into this later.

I honestly don't know that the problem is. The error is consistent. These two vepxress jobs fail in test `test_inventory.py::TestInventory::test_rootfs_type`, throwing the following error at boot:

```
Mounting [0;1;39m/data[0m...
[   21.260019] EXT4-fs (mmcblk0p4): Filesystem with huge files cannot be mounted RDWR without CONFIG_LBDAF
[[0;1;31mFAILED[0m] Failed to mount [0;1;39m/data[0m.
```

and making then the framework to hang until CI timeout (8 hours).

These errors, together with the fact that I don't have a good reason to port the patch to `zeus`, makes me conclude that it better to reject this PR.

PR #1067, opened 2020-08-25
Remove CI beaglebone/raspberrypi3 conf files

lluiscampos:
These seem to only change the MACHINE name. Should not have any effect
on the build.


lluiscampos:
Build for `beaglebone` and `raspberrypi3/4` @ [Pipeline 181777634](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/181777634)

lluiscampos:
> Looks ok to me but I wonder if any of the CI stuff will need to change.

The build for beablegoneblack/rasperrypi passed and there is no mention of these files in `mender-qa` repo (the corresponding yocto layers are used, but not explicitly the files). So I think it is okey to remove.

I'll wait for final word from @kacf

kacf (approved):
I think this was used back when we did HW tests. For sure this is completely out of date now, so :+1: to removing them!

PR #1069, opened 2020-08-27
MEN-3851: Warn when detecting U-Boot version without script '=' support. 

kacf:
This is needed on zeus and older, which do not use libubootenv.

Changelog: Title


lluiscampos (approved):

lluiscampos:
@kacf The pipeline failed on the license checks; we should build for 2.4.0 release. However I don't think we need to verify this patch, as nothing will actually change in our builds nor we have an explicit test for the warning. So I am okey with merging.

kacf:
Unrelated error, merging.

PR #1077, opened 2020-09-07
Fail build upon missing demo WiFi password and/or SSID

oleorhagen:
Another possibility is to `Warn` I guess?

Or maybe only fail if one is set, and not the other?

I like this approach though.


kacf:
Hmm, not sure. All three combinations are valid combinations: Nothing set (Ethernet), only SSID set (open network), both set (closed network). As much as I'd like to warn, I think we cannot avoid false warnings if we do.

PR #1083, opened 2020-09-10
Master yocto check layer changes

drewmoseley:

### Description

yocto-check-layer changes for master-next


lluiscampos:
Build at [Pipeline 188779910](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/188779910).

lluiscampos (changes requested):
It looks in pretty good shape to me, both this and #1084.

Two of your commits need amend:
```
Commit 2a2fa9b37b1e72e022040a1f1d745ae51cc2154c doesn't have a changelog tag! Make a changelog entry for your commit (https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#changelog-tags).
Commit d91d72da18fe576b90f0aa9fec3a01424dfb6a2e has less than three words in its changelog tag! Typo? (https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#changelog-tags).
```

For the second one I think the problem is the trailing `:` after `Title`.

lluiscampos:
@kacf I launched a build with this PR against mendersoftware/yocto:master. `master` in our fork is very old and we are getting:
```
ERROR: Layer mender is not compatible with the core layer which only supports these series: zeus (layer is compatible with dunfell)
```

So how should we actually build this PR?

drewmoseley:
Yeah, I was a bit unclear on the branching here. I did the master work against master-next and used upstream master for all the other branches.  Maybe the dunfell fixes in #1084 are more appropriate for the master branch.

lluiscampos:
> Yeah, I was a bit unclear on the branching here. I did the master work against master-next and used upstream master for all the other branches. Maybe the dunfell fixes in #1084 are more appropriate for the master branch.

I think master to use master is correct; is only that in our CI we use mendersoftware/yocto fork and that one has a very old master. I'll check this with @kacf on Monday :)

kacf:
Just FYI, this is a followup from https://github.com/mendersoftware/meta-mender/pull/1037.

drewmoseley:
Fixed the broken/missing changelogs.

madisongh:
These changes are a step in the right direction for making this layer more usable. Is there a roadmap for further improvements you're planning along these lines?  The use of global inherits is quite intrusive, as is overloading both DISTRO_FEATURES (worse, DISTRO_FEATURE_BACKFILL) and OVERRIDES for every MENDER_FEATURES setting globally.

There's also the MENDER_MACHINE export and BB_HASHBASE_WHITELIST change, which also affects the hash for every single build task, making it impossible to share sstate between Mender-enabled on non-Mender builds for package for which using Mender is irrelevant.

It would be really nice to see the mender-core layer make more progress towards minimizing its footprint/blast radius on users' builds, restricting any overrides or modifications to just those tasks that need to change, and only changing them when required.

kacf:
> These changes are a step in the right direction for making this layer more usable. Is there a roadmap for further improvements you're planning along these lines? The use of global inherits is quite intrusive, as is overloading both DISTRO_FEATURES (worse, DISTRO_FEATURE_BACKFILL) and OVERRIDES for every MENDER_FEATURES setting globally.

None of the Mender features are on by default, so all of those should be untouched unless you turn something on.

The dependency on `DISTRO_FEATURES` is not very strong, we already require that all of our features are enabled by using `MENDER_FEATURES_ENABLE`. So this dependency could potentially be removed. I guess the question is mostly about who will pick it up?

OTOH, the `OVERRIDES` are used extensively. And many of the options do indeed influence the build in a global fashion, such as `mender-uboot` or `mender-grub`. Don't you think this is appropriate for an override?

> There's also the MENDER_MACHINE export and BB_HASHBASE_WHITELIST change, which also affects the hash for every single build task, making it impossible to share sstate between Mender-enabled on non-Mender builds for package for which using Mender is irrelevant.

`MENDER_MACHINE` is only used for QEMU and the tests, so we could probably limit its exposure. I'm not sure why the whitelisting is there.

madisongh:
> > These changes are a step in the right direction for making this layer more usable. Is there a roadmap for further improvements you're planning along these lines? The use of global inherits is quite intrusive, as is overloading both DISTRO_FEATURES (worse, DISTRO_FEATURE_BACKFILL) and OVERRIDES for every MENDER_FEATURES setting globally.
> 
> None of the Mender features are on by default, so all of those should be untouched unless you turn something on.

It does make it more difficult to mix Mender-enabled and non-Mender-enabled platforms in the same distro, and thus more difficult to gradually add Mender support to an already-existing distro (presuming, of course, that there are recipes/bbappends checking for Mender features in DISTRO_FEATURES).  There's also quite a bit of settings and code that get included in every single recipe with the use of global inherits, cluttering `bitbake -e` output even for recipes for which Mender enablement is irrelevant.

> The dependency on `DISTRO_FEATURES` is not very strong, we already require that all of our features are enabled by using `MENDER_FEATURES_ENABLE`. So this dependency could potentially be removed. I guess the question is mostly about who will pick it up?
> 
> OTOH, the `OVERRIDES` are used extensively. And many of the options do indeed influence the build in a global fashion, such as `mender-uboot` or `mender-grub`. Don't you think this is appropriate for an override?

I don't think every single Mender feature requires its own override.  For instance, `mender-uboot` would only be relevant to machines that use U-Boot as their bootloader, while `mender-grub` would only be relevant to machines that use grub. Why can't just plain `mender` be used in that case?  Why should Mender features be applicable to anything except Mender components themselves, bootloader- or kernel-related recipes (possibly), and image types (or wic configuration files)?  Some system configuration files might also be necessary, but why couldn't those be provided as additional packages?

> > There's also the MENDER_MACHINE export and BB_HASHBASE_WHITELIST change, which also affects the hash for every single build task, making it impossible to share sstate between Mender-enabled on non-Mender builds for package for which using Mender is irrelevant.
> 
> `MENDER_MACHINE` is only used for QEMU and the tests, so we could probably limit its exposure. I'm not sure why the whitelisting is there.

I think it was added to prevent the global per-machine export from triggering rebuilds of non-machine-specific packages, but removing the global export and moving the definition of MENDER_MACHINE to just the recipes that need it would be a better solution.

lluiscampos (approved):
At this point, this PR looks good to me.

@kacf Should we just merge it into master-next and keep the changes there until we are ready to support the next yocto branch? We probably will need some bug fixing then, but there is no way to catch the errors right now via CI.

lluiscampos:
> @kacf Should we just merge it into master-next and keep the changes there until we are ready to support the next yocto branch? We probably will need some bug fixing then, but there is no way to catch the errors right now via CI.

And we should also handle https://github.com/mendersoftware/mender-qa/pull/416

kacf:
> Should we just merge it into master-next and keep the changes there until we are ready to support the next yocto branch? We probably will need some bug fixing then, but there is no way to catch the errors right now via CI.

No, not yet. I'm mainly handling #1084, which is the one which will actually go to meta-mender/master. And it still has problems, though I have not been able to work them out yet. Once that one is merged we will have a clearer view of what can go into master-next.

lluiscampos:
After merging #1097 we can now rework this PR. @kacf if you want, I can take over.

kacf:
Please do!

When you do, try to minimize the changes that need to go into `master-next`. This is not really a maintained branch, it's just a storage location for things we need to pick up when we move to the next Yocto branch. Only keep things there can cannot be added to `master` due to incompatibility with dunfell.

(I want to pick up the discussion from above as well, but I have to do that next week)

lluiscampos:
All commits from this PR are now in master via #1097, #1103 or in master-next via #1107. I am closing the PR to mark that no more changes are to be pushed here.

@kacf feel free to continue the discussion even if closed.

kacf:
> It does make it more difficult to mix Mender-enabled and non-Mender-enabled platforms in the same distro, and thus more difficult to gradually add Mender support to an already-existing distro

@madisongh: I'm not sure I understand. The overrides are supposed to make it *easier* to do exactly this; introduce features one by one as you need them. How do you think it should be done?

> I don't think every single Mender feature requires its own override. For instance, `mender-uboot` would only be relevant to machines that use U-Boot as their bootloader, while `mender-grub` would only be relevant to machines that use grub. Why can't just plain `mender` be used in that case?

It is possible that some of the features do not need to be overrides, but the `mender-uboot` and `mender-grub` overrides are perfect examples of ones that need to be: There is not just one Mender configuration. You choose which one to use, and you can turn on either of the two, or you can turn both of them off, and do something custom, while still using the rest of the Mender features. While it would be possible to do this purely as a global feature variable, without overrides, I think the result would be even uglier than what we have now (`bb.utils.contains()` calls all over the place).

> Why should Mender features be applicable to anything except Mender components themselves, bootloader- or kernel-related recipes (possibly), and image types (or wic configuration files)?

They aren't, but as soon as there is one more than one package that needs a change because of a feature, then you need something global. An alternative that we use in the Mender client is `PACKAGECONFIG`, but this does not propagate outside the recipe, which is what we need.

> Some system configuration files might also be necessary, but why couldn't those be provided as additional packages?

This is what we try to do, but mostly in addition to the overrides. I guess we have not really seen the overrides as being very intrusive, given that all of them are prefixed with "mender-".

madisongh:
>>It does make it more difficult to mix Mender-enabled and non-Mender-enabled platforms in the same distro, and thus more difficult to gradually add Mender support to an already-existing distro

>@madisongh: I'm not sure I understand. The overrides are supposed to make it easier to do exactly this; introduce features one by one as you need them. How do you think it should be done?

That comment about making things more difficult was mainly about adding the feature settings to DISTRO_FEATURES/DISTRO_FEATURES_BACKFILL and the use of global inherits.

> It is possible that some of the features do not need to be overrides, but the mender-uboot and mender-grub overrides are perfect examples of ones that need to be: There is not just one Mender configuration. You choose which one to use, and you can turn on either of the two, or you can turn both of them off, and do something custom, while still using the rest of the Mender features.

The selection of bootloader is a machine configuration setting, and is not a feature of Mender per se, right?  What if I have two machines in my distro, one that uses grub and one that uses uboot.  Is there an easy way to express the "mender-grub" enablement just for the grub machine while enabling "mender-uboot" just for the uboot machine (which might be arm/aarch64 and could be affected by the mender-grub override as well)?

> ...as soon as there is one more than one package that needs a change because of a feature, then you need something global.

Or you need to identify exactly which recipes/packages need those changes and supply specific bbappends for them.  I have about 2500 recipes in my builds (and that doesn't include the BBCLASSEXTENDed variants), but the number of recipes that might be affected by Mender features I could count on the fingers of one hand.

kacf:
> The selection of bootloader is a machine configuration setting, and is not a feature of Mender per se, right?

Actually, you're hitting the hammer on the nail exactly here: Yes, it *is* a feature of Mender. What you are selecting is not which bootloader to use, but which bootloader to integrate with. This is an important difference, and they can even be different. For example, on ARM we provide booting through UEFI/GRUB, but since ARM typically doesn't come with a UEFI layer, we leverage the UEFI provider support in U-Boot to do it, and the boot becomes U-Boot->UEFI->GRUB. In this case your machine configuration uses U-Boot, while Mender uses `mender-grub` (and also pulls in GRUB, of course).

But you're right that it does relate strongly to machine configuration, and I would think it natural to add something like `MENDER_FEATURES_ENABLE_append = " mender-uboot"` to my `machine/<MACHINE>.conf`.

> Or you need to identify exactly which recipes/packages need those changes and supply specific bbappends for them. I have about 2500 recipes in my builds (and that doesn't include the BBCLASSEXTENDed variants), but the number of recipes that might be affected by Mender features I could count on the fingers of one hand.

We don't know what the changes are until the user has selected his `MENDER_FEATURES` combination. And the only way I know to propagate that to *multiple* bbappends, is to use global overrides or variables.

Is the primary objection you have that `DISTRO_FEATURES` and/or `OVERRIDES` are being used, or that the features are enabled in a global fashion?

kacf:
Thanks for the input, I agree very much with the general intention, but the devil's in the details!

To get started, I can think of at least a few low hanging fruits that would be possible to do:

1. Move Mender's overrides from `OVERRIDES` to `DISTRO_OVERRIDES`. This has already been contributed [here](https://github.com/mendersoftware/meta-mender/pull/1144), and is in the process of being tested. This at least namespaces it a bit better, and allows `forcevariable` to work correctly.

2. Move all Mender features from `DISTRO_FEATURES` to `MENDER_FEATURES`. They don't actually need to be in `DISTRO_FEATURES`, this was just a convenient list to use at the time that it was implemented.

> I think a good first goal would be to enable users to selectively enable Mender support on a subset of the machines supported in their distro, with no interference with the rest of the builds.

Doesn't it do the trick to simply add `inherit mender-full` (or something more selective with `MENDER_FEATURES_ENABLE`) to one of the `machine/<MACHINE>.conf` and omitting it from the other? At least this is the intention.

I know the sstate cache may be busted for this scenario, but let's focus on correctness first, then we can worry about performance.

madisongh:
> Thanks for the input, I agree very much with the general intention, but the devil's in the details!

Definitely.

>Doesn't it do the trick to simply add inherit mender-full (or something more selective with MENDER_FEATURES_ENABLE) to one of the machine/<MACHINE>.conf and omitting it from the other? 

Machine configurations shouldn't be mucking around with distro settings (which represent policy to be applied across all builds in the distro), but that might work about as well as having a separate distro.

PR #1090, opened 2020-09-17
Remove unsupported versions after Mender 2.5 release

lluiscampos:
Changelog: Remove mender-client 2.2.1 recipe


lluiscampos:
@kacf For `mender-artifact` we have both 3.3.0 and 3.3.1, should we have removed already 3.3.0?

kacf:
Since Mender 2.4 and 2.5 both use mender-artifact 3.4, I think we can remove both of the 3.3 versions now.

I usually keep patch versions around unless there is some specific reason to remove them (like a vulnerability). They are supposed to be fully compatible with each other, so it gives the user more opportunities to choose if the latest one has a regression. Not a strict policy though, just seemed sensible and cheap.

lluiscampos:
> Since Mender 2.4 and 2.5 both use mender-artifact 3.4, I think we can remove both of the 3.3 versions now.

Good point, I will remove them.

> I usually keep patch versions around unless there is some specific reason to remove them (like a vulnerability). They are supposed to be fully compatible with each other, so it gives the user more opportunities to choose if the latest one has a regression. Not a strict policy though, just seemed sensible and cheap.

It makes sense. I will use this policy from now on.

lluiscampos:
ping @kacf

kacf (approved):

PR #1093, opened 2020-09-24
MEN-3076: update test_build to the latest mender-artifact

tranchitella:
Changelog: none

Signed-off-by: Fabio Tranchitella <fabio.tranchitella@northern.tech>

### DIFF
--- a/tests/acceptance/test_build.py
+++ b/tests/acceptance/test_build.py
@@ -948,6 +948,11 @@ deployed-test-dir9/*;renamed-deployed-test-dir9/ \
         # Then remove it, not to mess up the expected test output
         del other.provides["rootfs_image_checksum"]
 
+        # MEN-3076: Mender-Artifacts writes software version by default
+        # older versions did not, thus we remove the key before asserting the content
+        if "rootfs-image.version" in other.provides.keys():
+            del other.provides["rootfs-image.version"]
+
         assert dependsprovides.__dict__ == other.__dict__
 
     @pytest.mark.only_with_image("sdimg", "uefiimg", "gptimg", "biosimg")


tests/acceptance/test_build.py
```diff
@@ -948,6 +948,9 @@ def test_build_artifact_depends_and_provides(
         # Then remove it, not to mess up the expected test output
         del other.provides["rootfs_image_checksum"]
 
+        # MEN-3076: Mender-Artifacts writes software version by default
+        dependsprovides.provides.update({'rootfs-image.version': 'mender-image-master'})
```
kacf, line 952: Meta-mender is used by several different Mender versions, this will break with older versions. I propose to delete the key instead, as we do for `rootfs_image_checksum`. This is tested in so many other places that I don't think it's critical to test it here.

kacf (changes requested):

kacf:
Btw, I have no clue what's going on with the `id_rsa` failure. @lluiscampos, @oleorhagen: Any ideas of the top off your head?

Note that you can leave Raspberry Pi platform off. It's off by default in mender-qa because they keep breaking the layer upstream, so we only handle it during releases.

tests/acceptance/test_build.py
tranchitella: Ack, I wasn't aware of this back-dependency on older versions, I've updated the test.

kacf (approved):

kacf:
@lluiscampos: Is meta-mender considered when the integration-test-runner chooses which branch to run? I suspect it's not. If it causes problems, we could merge this particular change to `master`, since it should be compatible with any branch, and does not really harm anything by being introduced early.

lluiscampos:
> Btw, I have no clue what's going on with the `id_rsa` failure. @lluiscampos, @oleorhagen: Any ideas of the top off your head?

I haven't seen this before neither. I am looking a bit more into it.

> @lluiscampos: Is meta-mender considered when the integration-test-runner chooses which branch to run? I suspect it's not.

It cannot, as meta-mender is not tracked by the `release_tool`.

> If it causes problems, we could merge this particular change to `master`, since it should be compatible with any branch, and does not really harm anything by being introduced early.

I guess we need to make an exception with the feature branches approach wrt meta-mender repo. In this case is harmless to merge to master, but also in general meta-mender should be able to build different versions of Mender, so most of the cases this will be fine. Only when introducing a new test for a new feature can be a problem, but then we could just wait on the PR stage...

lluiscampos:
> > Btw, I have no clue what's going on with the `id_rsa` failure. @lluiscampos, @oleorhagen: Any ideas of the top off your head?
> 
> I haven't seen this before neither. I am looking a bit more into it.

Ah, yeah. This Pipeline was launched with `TEST_BEAGLEBONEBLACK=true`. The regular build actually succeed, but then the "test" failed building the image.

@tranchitella We don't test on real hardware from CI, so for future reference don't set them to true. Basically as Kristian said: leave everything by default when launching build for client PRs

kacf:
Yes, good catch, @lluiscampos! Make sure the defaults from mender-qa are used, not everything on.

PR #1094, opened 2020-09-24
MEN-3970: Make libubootenv support key<space>value syntax in scripts

oleorhagen:
This is required in order to support the old script syntax present in some older
version of U-Boot.

Changelog: Title


oleorhagen:
Note: Not tested yet!

oleorhagen:
Decided to not add this to `libubootenv` Instead the client will probe the bootloader for the correct separator to use.

PR #1099, opened 2020-09-29
Warrior update v2020-09

lluiscampos:

PR #1098, opened 2020-09-29
Zeus update v2020-09

lluiscampos:

PR #1101, opened 2020-09-30
Inherit go-ptest in Mender golang recipes

lluiscampos:
The ptest part of go class was separated to go-ptest class some time
before zeus. See poky commit 70a862a.

https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?id=70a862a367beea2c92ec0fb1bcaba41b942c78ac

By the looks of it, since then we have not been running unit tests on
our mender-client and mender-artifact recipes builds.


kacf (approved):

lluiscampos:
Cherry-picked to zeus (via #1098) and warrior (via #1099).

PR #1104, opened 2020-09-30
MEN-3648

oleorhagen:
* Cleanup

* Disable the version check on release binaries


meta-mender-commercial/recipes-mender/mender-binary-delta/mender-binary-delta.inc
```diff
@@ -30,6 +30,7 @@ do_version_check() {
         bbfatal "String '${PN} ${PV}' not found in binary. Is it the correct version? Check with --version. Possible candidates: $(strings ${WORKDIR}/mender-binary-delta | grep '${PN} [a-f0-9]')"
     fi
 }
+do_version_check[noexec] = "${@'1' if d.getVar('DEV_MODE') else '0'}"
```
kacf, line 33: Really nice solution! But I would consider changing the name of the variable to `MENDER_DEV_MODE` just to reduce the likelihood of clashing with something in poky.

kacf (changes requested):

kacf (approved):
Perfect!

PR #1097, opened 2020-09-28
Dunfell yocto check layer changes + rebase + fix

lluiscampos:
See #1084

PR #1110, opened 2020-10-08
Dunfell update

kacf:
As previously agreed, this brings the "Yocto check layer" changes (#1097 and #1037) into dunfell, which is a quite big change, but we opted to put it in dunfell since the next release is very far away. We do not plan to port it further to zeus and older though.


lluiscampos (approved):
The commit list looks sensible to me.

PR #1113, opened 2020-10-09
test: Remove test_redundant_uboot_env

oleorhagen:
This test fails on the new client because it now writes to the U-Boot
environment a probe, and hence dirtying the env, which will lead to this test
always failing.

Besides, this is really testing U-Boot functionality which is out of our source
tree. Therefore, removing this test.

Changelog: None


oleorhagen:
This is in relation to: https://github.com/mendersoftware/mender/pull/616

kacf (approved):

PR #1114, opened 2020-10-09
mender-binary-delta: do_version_check() fixes

kacf:
```
commit f08e4c8819da6446fc3ec90802f9b57b61bed251
Date:   Mon Oct 12 13:02:01 2020

    mender-binary-delta: Fix broken do_version_check.
    
    For unknown reasons, the "noexec" flag is not actually possible to
    turn off, not even with "0" nor "". The only way to disable it is to
    remove it, which kind of defeats the purpose of having it. Luckily
    it's easy to work around, so just do the variable check in code
    instead.
    
    The reason this has passed in our CI is that it is not covered by the
    sstate cache, so if it had a hit then this problem would not reveal
    itself.
    
    Changelog: None
    
commit 14113b2f3d8a4a2688ea9879f565caca2dade015
Date:   Fri Oct 9 14:09:47 2020

    mender-binary-delta: Fix do_version_check having side effects.
    
    Changelog: None
    
```

### DIFF
--- a/meta-mender-commercial/recipes-mender/mender-binary-delta/mender-binary-delta.inc
+++ b/meta-mender-commercial/recipes-mender/mender-binary-delta/mender-binary-delta.inc
@@ -24,11 +24,12 @@ FILES_${PN} = " \
 INSANE_SKIP_${PN} = "already-stripped"
 
 do_version_check() {
-    if ! strings ${WORKDIR}/mender-binary-delta | fgrep -q "${PN} ${PV}"; then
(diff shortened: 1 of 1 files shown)


mirzak:
It worked because of [this](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-commercial/recipes-mender/mender-binary-delta/mender-binary-delta.inc#L27) I suppose.

lluiscampos (approved):

kacf:
> It worked because of [this](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-commercial/recipes-mender/mender-binary-delta/mender-binary-delta.inc#L27) I suppose.

Well spotted. Then this fix is actually wrong.. I'll rework this and try to figure out why it didn't work for me.

oleorhagen (approved):
:+1:

PR #1122, opened 2020-10-15
Add dependabot for pip, Docker, and Git Submodules

lluiscampos:


.github/dependabot.yml
```diff
@@ -0,0 +1,21 @@
...
+updates:
+  - commit-message:
+      prefix: "Changelog:All"
+    directory: /tests/acceptance/
+    package-ecosystem: pip
+    schedule:
+      interval: weekly
+  - commit-message:
+      prefix: "Changelog:All"
+    directory: /meta-mender-qemu/docker/
```
lluiscampos, line 11: @mzedel I am unsure about this one, the Dockerfile(s) are at `/meta-mender-qemu/docker/*/Dockerfile`

mzedel: I don't know it for a fact, but I'm pretty sure these would have to be added as individual entries in the dependabot file (so one with _qemux86-64_ and one with _vexpress-qemu_ as the directory)

mzedel (changes requested):
left a comment for the `docker` updates as I think that would fail, the only other nitpick is the trailing slash for the directory, but this should not cause problems...

lluiscampos:
> left a comment for the `docker` updates as I think that would fail, the only other nitpick is the trailing slash for the directory, but this should not cause problems...

Thanks. It should be fixed now :+1:

mzedel (approved):

PR #1136, opened 2020-10-17
[dependabot] Extend submodule update to maintained branches

lluiscampos:
Trigger submodule update from zeus on, including future yocto branches.
Note that warrior is out as it uses Python 2 and, as per today, can not
be further updated.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,4 +24,32 @@ updates:
     package-ecosystem: gitsubmodule
     schedule:
       interval: weekly
+  - commit-message:
+      prefix: "Changelog:All"
+    directory: /
+    package-ecosystem: gitsubmodule
+    schedule:
+      interval: weekly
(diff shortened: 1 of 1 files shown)


lluiscampos:
@mzedel As I understand [the feature](https://dependabot.com/docs/config-file/#target_branch), duplicating the job is the only way. There is however a recent [feature request](https://github.com/dependabot/dependabot-core/issues/2511) to simplify this.

lluiscampos:
> Hello smile_cat I created a pipeline for you here: [Pipeline-203901067](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/203901067)
> Build Configuration Matrix

Stopped the build.

mzedel (approved):
Thanks for looking into ways to avoid the duplication! Let's hope the request gets a response as I guess the list might only grow with every new yocto branch...

PR #1142, opened 2020-10-19
mender-client: fix a wrong PACKAGECONFIG

liuming50:
libubootenv-bin should not be set explicitly since it's one of a
RPROVIDER of u-boot-fw-utils, or else the end users want to use
u-boot-fw-utils would get build errors.

Fixed by changing to u-boot-fw-utils.

Signed-off-by: Ming Liu <liu.ming50@gmail.com>

### Description

Thank you!

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client.inc
@@ -100,7 +100,7 @@ PACKAGECONFIG_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-uboot', 
 PACKAGECONFIG_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-grub', ' grub', '', d)}"
 
 PACKAGECONFIG[mender-client-install] = ",,,mender-artifact-info ca-certificates"
-PACKAGECONFIG[u-boot] = ",,,libubootenv-bin"
+PACKAGECONFIG[u-boot] = ",,,u-boot-fw-utils"
 PACKAGECONFIG[grub] = ",,,grub-editenv grub-mender-grubenv"
 # The docker module depends on bash, and of course on docker. However, docker is
 # a very large requirement, which we will not mandate. Bash however, we require,


lluiscampos:
Hi @liuming50 and thanks for your contribution.

The change looks good to me.

However you are missing a Changelog tag in your commit (ref "External Contributor Checklist" above). For this PR I suggest something like `Changelog: mender-client: u-boot config depends on u-boot-fw-utils instead of libubootenv-bin`

liuming50:
@lluiscampos thanks for the review, I have pushed a V2 added Changelog item as you suggested, please kindly help me review it.

lluiscampos:
Thanks! Running now all our tests against this PR.

lluiscampos (approved):

kacf:
I'm not sure about this one. I think we should depend on libubootenv. The reasoning is a bit tricky, but here goes:

libubootenv has one major difference from u-boot-fw-utils, which is that it does not come with a default environment. This is *very good*, because providing a default environment opens up for a broken Mender setup. Mender requires that the environment be written by the bootloader when the board first boots, so having a default environment actually masks this issue, and makes it both harder to understand why things are not working, and easier to screw up the setup by mistake. For an example of how this can go wrong, see [this thread](https://hub.mender.io/t/mender-setup-failure-orphans-deployed-sensors-need-remote-recover-solution/2167) on Mender Hub (search for the word "dunfell" also to see my reasoning for why we should always use libubootenv).

My next question is then: Why do you need u-boot-fw-utils at all? libubootenv is supposed to provide everything that u-boot-fw-utils.

PS! I know that there is [a bug](https://tracker.mender.io/browse/MEN-3970) regarding libubootenv, but this has been fixed separately in the client, and will be released very soon.

liuming50:
@kacf Sorry for the delayed feedback, I missed the mail notification of your comment.

First of all, I agree with you about the libubootenv's advantages, actually this fix would not lead to any functional changes in dunfell and master, since libubootenv is the preferred provider/rprovider of u-boot-fw-utils by default.

But theoretically the end users should be able to choose their own PREFERRED_PROVIDER_u-boot-fw-utils/PREFERRED_RPROVIDER_u-boot-fw-utils in their machine config (a user case could be, if they are building a old uboot and there is not a u-boot-initial-env which is being required by libubootenv, see https://patchwork.openembedded.org/patch/175708/), and this line "PACKAGECONFIG[u-boot] = ",,,libubootenv-bin"" makes that impossible by explicitly depending on a libubootenv-bin.

kacf:
> a user case for not choosing libubootenv could be, if they are building a old uboot and there is not a u-boot-initial-env which is being required by libubootenv

This is exactly the case we are protecting against. `u-boot-initial-env` must not be included in the image, but must be written by the bootloader. We should actually add `RCONFLICTS_libubootenv = "u-boot-initial-env"` to the `libubootenv.bbappend` file in meta-mender.

Unless there is some hardware that doesn't work with libubootenv, then I can't really see a convincing case for why we should allow u-boot-fw-utils, given that we know it causes problems.

If you really need to use u-boot-fw-utils then I propose this workaround for your specific case: Create a `u-boot-fw-utils_%.bbappend` file, and add:

```
PROVIDES_append = " libubootenv-bin"
RPROVIDES_${PN}_append = " libubootenv-bin"
```

liuming50:
> > a user case for not choosing libubootenv could be, if they are building a old uboot and there is not a u-boot-initial-env which is being required by libubootenv
> 
> This is exactly the case we are protecting against. `u-boot-initial-env` must not be included in the image, but must be written by the bootloader. We should actually add `RCONFLICTS_libubootenv = "u-boot-initial-env"` to the `libubootenv.bbappend` file in meta-mender.
> 
> Unless there is some hardware that doesn't work with libubootenv, then I can't really see a convincing case for why we should allow u-boot-fw-utils, given that we know it causes problems.
> 
> If you really need to use u-boot-fw-utils then I propose this workaround for your specific case: Create a `u-boot-fw-utils_%.bbappend` file, and add:
> 
> ```
> PROVIDES_append = " libubootenv-bin"
> RPROVIDES_${PN}_append = " libubootenv-bin"
> ```

@kacf  Thanks for the suggestion, I actually have a workaround, just simply put "PACKAGECONFIG[u-boot] = ",,,u-boot-fw-utils"" in bbappend would override the original setting in bb file,  this is not a block issue for me, 

OK, I would not insist we should merge this, though I do think it should be the BSP developer's responsibility to choose their preferred provider/rproviders for a specific recipe and risk on their own, that's the reason why these PREFERRED_PROVIDER/PREFERRED_RPROVIDER variables were introduced.

kacf:
> OK, I would not insist we should merge this, though I do think it should be the BSP developer's responsibility to choose their preferred provider/rproviders for a specific recipe and risk on their own, that's the reason why these PREFERRED_PROVIDER/PREFERRED_RPROVIDER variables were introduced.

Ok, thanks for understanding. If I didn't know that u-boot-fw-utils was problematic, I would agree it's good to provide a choice. It's just in this specific case that I disagree.

PR #1145, opened 2020-10-21
MEN-2208: test_update.py to mender-image-tests

lluiscampos:
The actual fixes are:

* [tests/acceptance] MEN-2208: New cli parameter --qemu-wrapper
Defaulting to the previously hard-coded path, this parameter allows for
the user to specify a different wrapper to launch QEMU.

* [tests/acceptance] MEN-2208: New cli parameter --ssh-priv-key
Defaulting to empty, this parameter lets the user specify an SSH private
key to log in into the device under test. Take into account the key on
regular Fabric calls and system's ssh/scp calls.
    
* [tests/acceptance] MEN-2208: Fixtures amends for convert images
 * Split qemu_finalizer clean-ups in different try/except blocks to do as
many clean-ups as possible but never fail.
 * Modify fixtures build_image_fn and prepared_test_build_base to work in
mender-convert scenarios (i.e. conversion fixture).

The submodule sha is pointing to https://github.com/mendersoftware/mender-image-tests/pull/17 to test them together.


lluiscampos:
Tests for `test_accep_vexpress_qemu_flash` fail with:

```
sh: scp: command not found
```

I am investigating.

kacf:
I cannot spot the problem either. What I can say is that I believe vexpress-qemu-flash is the only image which uses core-image-minimal, which means that Dropbear ssh is used instead of OpenSSH. This could explain why it only affects this image.

lluiscampos:
> I cannot spot the problem either. What I can say is that I believe vexpress-qemu-flash is the only image which uses core-image-minimal, which means that Dropbear ssh is used instead of OpenSSH. This could explain why it only affects this image.

@kacf Thanks for looking into it.

This might have been the case in the past, but not any more. If you look at the [log in latest nightly](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/805228938), you can see that `openssh` is added to the image.

Anyway, this doesn't seem to be the culprit. Very strange error...

tests/acceptance/fixtures.py
```diff
@@ -301,15 +312,13 @@ def latest_part_image(conversion, mender_image):
...
+def successful_image_update_mender(request, latest_mender_image):
     """Provide a 'successful_image_update.mender' file in the current directory that
     contains the latest built update."""
 
-    latest_mender_image = latest_build_artifact(build_image_fn(), "core-image*.mender")
-
     shutil.copy(latest_mender_image, "successful_image_update.mender")
 
-    print("Copying 'successful_image_update.mender' to '%s'" % latest_mender_image)
+    print("Copying '%s' to 'successful_image_update.mender'" % latest_mender_image)
```
lluiscampos, line 321: @kacf @merlin-northern This is what is making the vexpress-tests fail.

With the previous code, build_image_fn() returned a yocto build dir with `ssh-server-openssh` (see [here](https://github.com/mendersoftware/meta-mender/blob/d9e6cece4eba175d13606097ea6417528e434606/tests/acceptance/fixtures.py#L372)), while with my change is just returns the original `.mender` that was build.

Then, I assume that for core-image-full-cmdline (i.e. all configurations except vespress-flash) is already included, so this statement does not take any effect.

Finally, why does it fail in `test_signed_updates` and not before? Because the QEMU image being used for tests has openssh server (and scp) so all previous tests using scp go fine, until we arrive to `test_network_based_image_update`, which passes but installs a image witn no scp, and next test to execute is `test_signed_updates`.

Now, I only have to fix (it is not enough with just revert, as this change actually fixed something else in mender-convert). New push is coming up.

merlin-northern: wow. well done. I was just about to comment that if getting scp in is a problem, then we can reimplement it with i/o redirection, since we have ssh.

tests/acceptance/common.py
```diff
@@ -52,11 +52,7 @@ def _start_qemu(qenv=None, conn=None):
     if qenv:
         env.update(qenv)
 
-    proc = subprocess.Popen(
-        ["../../meta-mender-qemu/scripts/mender-qemu"],
-        env=env,
-        start_new_session=True,
-    )
+    proc = subprocess.Popen([qemu_wrapper], env=env, start_new_session=True,)
```
merlin-northern, line 55: wait, so what if `qemu_wrapper` is None here?

tests/acceptance/fixtures.py
```diff
@@ -118,7 +118,7 @@ def board_cleanup():
     request.addfinalizer(board_cleanup)
 
 
-def setup_qemu(request, build_dir, conn):
+def setup_qemu(request, qemu_wrapper, build_dir, conn):
```
merlin-northern, line 124: minor: I wonder why you added that as a second and not last argument?

tests/acceptance/common.py
```diff
@@ -36,7 +36,7 @@
 configuration = {}
 
 
-def _start_qemu(qenv=None, conn=None):
+def _start_qemu(qenv=None, conn=None, qemu_wrapper=None):
```
merlin-northern, line 39: could we have at least a bit of comment here on what qemu_wrapper is?

merlin-northern (changes requested):

tests/acceptance/common.py
```diff
@@ -222,13 +222,16 @@ def determine_active_passive_part(bitbake_variables, conn):
         )
 
 
-def get_ssh_common_args():
-    return "-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
+def get_ssh_common_args(conn):
+    args = "-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
+    if "key_filename" in conn.connect_kwargs.keys():
+        args += " -i %s" % conn.connect_kwargs["key_filename"]
```
merlin-northern, line 228: I would add a space here, just in case, i.e.: `" -i %s"` -> `" -i %s "`

merlin-northern: also maybe some quotes around the %s? as in: what if the `conn.connect_kwargs["key_filename"]` contains spaces?

tests/acceptance/common.py
```diff
@@ -222,13 +222,16 @@ def determine_active_passive_part(bitbake_variables, conn):
         )
 
 
-def get_ssh_common_args():
-    return "-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
+def get_ssh_common_args(conn):
+    args = "-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
+    if "key_filename" in conn.connect_kwargs.keys():
```
merlin-northern, line 227: for backwards compatibility maybe lets do `conn=None` and here add an `if conn:` ? (just a suggestion)

tests/acceptance/common.py
```diff
@@ -222,13 +222,16 @@ def determine_active_passive_part(bitbake_variables, conn):
...
+        args += " -i %s" % conn.connect_kwargs["key_filename"]
+    return args
 
 
 # Yocto build SSH is lacking SFTP, let's override and use regular SCP instead.
 def put_no_sftp(file, conn, remote="."):
-    cmd = "scp -C %s" % get_ssh_common_args()
+    cmd = "scp -C %s" % get_ssh_common_args(conn)
```
merlin-northern, line 234: I really wonder why we enable compression here, and everywhere within the acceptance tests. I would expect to disable the compression. is that on purpose?

tests/acceptance/conftest.py
```diff
@@ -36,6 +36,12 @@ def pytest_addoption(parser):
         default="root",
         help="user to log into remote hosts with (default is root)",
     )
+    parser.addoption(
+        "--ssh-priv-key",
+        action="store",
+        default="",
+        help="Path to SSH private key, when required for login",
```
merlin-northern, line 43: minor: suggested rewording: "Path to _an_ SSH private key .."

tests/acceptance/fixtures.py
```diff
@@ -157,25 +157,31 @@ def setup_qemu(request, qemu_wrapper, build_dir, conn):
     # cases more predictable.
     def qemu_finalizer():
         def qemu_finalizer_impl(conn):
+            # Try clearing FW variables
```
merlin-northern, line 160: FW stands for?

tests/acceptance/common.py
lluiscampos: Good question. I will modify the method signature to make all arguments mandatory.

tests/acceptance/fixtures.py
lluiscampos: Because it is the most important, IMO. Just personal taste, I guess :)

tests/acceptance/common.py
lluiscampos: Okey.

tests/acceptance/common.py
lluiscampos: Why the space? We know exactly how and where is this used. Similarly for the quotes, it is only used in one place and we know the path.

Generally speaking, when I write test code I try not to be too defensive and let things fail. A failure might hint someone that is doing something fishie.

tests/acceptance/common.py
lluiscampos: I think I'll leave it as it is.

I would say that the normal case should be having a key, and that Yocto builds are the weird ones (it is created specially for testing that allows root acess with no password). So IOW, I expect this to be the regular path for other future images from mender-convert.

tests/acceptance/common.py
lluiscampos: I don't understand this comment, I think you meant to comment somewhere else...

tests/acceptance/conftest.py
lluiscampos: Ok

tests/acceptance/fixtures.py
lluiscampos: firmware. I'll rewrite.

tests/acceptance/common.py
lluiscampos: Now I understand your comment :)

Good question! I don't know myself and it has been like this for very long time, so I don't dare to touch.

lluiscampos: I guess that for the QEMU testing you are correct, but the full framework was designed to also run tests in real devices. So if you have a device in Oslo and the test running in a GCP VM, then the transfer time would be longer that the compression, decompression, right?

merlin-northern: the question is how often we have a device in Oslo and test is running in GCP.

merlin-northern (approved):

tests/acceptance/common.py
lluiscampos: 0% in Oslo and 100% on GCP :wink:  This is very valid point, thanks for catching it. I will follow-up independently.

lluiscampos:
Last force-push consist on updating submdoule to latest master (after https://github.com/mendersoftware/mender-image-tests/pull/17) and a rebase on latest master.

Previous build can be found at [Pipeline 206460709](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/206460709).

PR #1147, opened 2020-10-22
MEN-4014: Make DBus support optional in Mender client with `PACKAGECONFIG`.

kacf:
It defaults to on, but can be turned off with:

```
PACKAGECONFIG_remove = "dbus"
```

Changelog: Commit


lluiscampos (approved):
It looks good, the logic makes sense to me. Nice test also.

lluiscampos (approved):

kacf:
Finally, that was two difficult story points!

PR #1148, opened 2020-10-23
mender-client: Update LICENSE to include OpenSSL

lluiscampos:
For 2.4.0 recipe, 2.4.x builds, and master builds.

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client_git.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client_git.inc
@@ -84,10 +84,15 @@ def mender_license(branch):
                    "md5": "3e7426c258f60f9876ae3746f54c49d4",
                    "license": "Apache-2.0 & BSD-2-Clause & BSD-3-Clause & ISC & MIT & OLDAP-2.8",
         }
+    elif branch == "2.4.x":
+        return {
+                   "md5": "5649992d13a6fda40abfac7730a62b07",
+                   "license": "Apache-2.0 & BSD-2-Clause & BSD-3-Clause & ISC & MIT & OLDAP-2.8 & OpenSSL",
+        }
     else:
         return {
                    "md5": "5649992d13a6fda40abfac7730a62b07",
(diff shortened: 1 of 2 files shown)


lluiscampos:
Stopping the integration pipeline.

@kacf also, do you think this deserves a Changelog tag?

meta-mender-core/recipes-mender/mender-client/mender-client_git.inc
```diff
@@ -84,10 +84,15 @@ def mender_license(branch):
...
+    elif branch == "2.4.x":
+        return {
+                   "md5": "5649992d13a6fda40abfac7730a62b07",
+                   "license": "Apache-2.0 & BSD-2-Clause & BSD-3-Clause & ISC & MIT & OLDAP-2.8 & OpenSSL License",
+        }
     else:
         return {
                    "md5": "5649992d13a6fda40abfac7730a62b07",
-                   "license": "Apache-2.0 & BSD-2-Clause & BSD-3-Clause & ISC & MIT & OLDAP-2.8",
+                   "license": "Apache-2.0 & BSD-2-Clause & BSD-3-Clause & ISC & MIT & OLDAP-2.8 & OpenSSL License",
```
kacf, line 95: I believe it's just `OpenSSL`. It maps directly to [this file](https://github.com/mendersoftware/poky/blob/master/meta/files/common-licenses/OpenSSL).

kacf (changes requested):

kacf:
> Stopping the integration pipeline.

You probably should let it run actually, at least the `build_client` step. It's possible the field has some ripple effects, since Yocto is quite strict about licenses.

meta-mender-core/recipes-mender/mender-client/mender-client_git.inc
lluiscampos: Thanks, I will change it. I thought about it, but followed the upstream reference [here](https://www.openssl.org/source/license-openssl-ssleay.txt). (Plus I had no idea that poky had this)

kacf:
> do you think this deserves a Changelog tag?

Yes, I think so. I think it will affect virtually no one, but it's still good to be open about it I think.

kacf (approved):

PR #1150, opened 2020-10-26
[dependabot] Fix parse error in dependabot file

lluiscampos:
By explicitly marking the first entry with target_branch master, so that
dependabot does not detect the rest as duplicates of this one.


mzedel (approved):
🤞 let's hope this really was the parsing issue...

PR #1156, opened 2020-10-28
[rocko] Fix lockfile error due to using system lock directory.

lluiscampos:
Cherry-picking 7879f6ac3f24e94c3ab0ec227f71e5d85e8a372f to rocko. See [community request at Mender Hub](https://hub.mender.io/t/do-mender-uboot-auto-configure-fails-due-to-multiple-users-running-the-build/2722).


lluiscampos:
@kacf I would like to merge this one without testing, as it is hard for our CI to run `rocko` nowadays and the bugfix seems to be very low risk. What do you think?

kacf:
The `-l` option was introduced in U-Boot v2016.11. Rocko is using v2017.09, so it should be fine. However, users of rocko are far more likely to be on a U-Boot fork from a version which is before v2016.11.

I think I would advise against merging it actually, since it is easier to fix the issue by removing a lock file, than it is to patch U-Boot to support the option. But I don't have a strong opinion, so consider it advise only.

PR #1157, opened 2020-10-28
test: bump min mender version

oleorhagen:
The fix for the sudo bug is on the released Mender-Artifact 3.4.x branch, but is
not yet released, meaning that it is not in the 3.4.0 tag in Mender-Artifact:

https://github.com/mendersoftware/mender-artifact/commits/3.4.0

Therefore, this will only be in master and be released with Mender release 2.6.0

### DIFF
--- a/tests/acceptance/test_snapshot.py
+++ b/tests/acceptance/test_snapshot.py
@@ -196,7 +196,7 @@ class TestSnapshot:
             except:
                 pass
 
-    @pytest.mark.min_mender_version("2.2.0")
+    @pytest.mark.min_mender_version("2.5.0")
     @pytest.mark.only_with_image("uefiimg", "sdimg", "biosimg", "gptimg")
     # Make sure we run both with and without terminal. Many signal bugs lurk in
     # different corners of the console code.


oleorhagen:
I do _think_ this is right

oleorhagen:
Client version it is.

lluiscampos (changes requested):
> I do _think_ this is right. or should it be `2.5.0` client version?

If any, it should be 2.5.0 client version, yes.

But I think it is about time to introduce `min_mender_artifact_version` fixture. If you look at the code for `min_mender_version`, you will see that it is quite straight forward to have a similar one for mender-artifact, as we can always find out the version of the tool in the yocto build (it won't work for mender-convert tests, but that we don't need to worry about now).


(related)
I also realized that [here](https://github.com/mendersoftware/meta-mender/blob/master/tests/acceptance/fixtures.py#L465) we query for `mender` instead of `mender-client`. This works just because in [here](https://github.com/mendersoftware/mender-qa/blob/master/scripts/jenkins-yoctobuild-build.sh#L143) we set both, but we should correct the fixture anyway. -> Let me follow-up on this independently, as I suspect needs a fix in `mender-convert` too.

oleorhagen:
The recipe renaming has had some funny reprecussions

oleorhagen:
> (related)
> I also realized that [here](https://github.com/mendersoftware/meta-mender/blob/master/tests/acceptance/fixtures.py#L465) we query for `mender` instead of `mender-client`. This works just because in [here](https://github.com/mendersoftware/mender-qa/blob/master/scripts/jenkins-yoctobuild-build.sh#L143) we set both, but we should correct the fixture anyway. -> Let me follow-up on this independently, as I suspect needs a fix in `mender-convert` too.

I did add it here: https://github.com/mendersoftware/meta-mender/pull/1158/commits/aa406d8a9102a176d4151b428ca3c6651a01fdcb

Should I remove it?

lluiscampos:
> > (related)
> > I also realized that [here](https://github.com/mendersoftware/meta-mender/blob/master/tests/acceptance/fixtures.py#L465) we query for `mender` instead of `mender-client`. This works just because in [here](https://github.com/mendersoftware/mender-qa/blob/master/scripts/jenkins-yoctobuild-build.sh#L143) we set both, but we should correct the fixture anyway. -> Let me follow-up on this independently, as I suspect needs a fix in `mender-convert` too.
> 
> I did add it here: [aa406d8](https://github.com/mendersoftware/meta-mender/commit/aa406d8a9102a176d4151b428ca3c6651a01fdcb)
> 
> Should I remove it?

No, is okey. Thanks.

lluiscampos (approved):

PR #1159, opened 2020-10-30
[tests/acceptance] Re-introduce min_mender_version fixture

lluiscampos:
For backwards compatibility in mender-convert stable release branch.


lluiscampos:
Reverting the original PR instead. See #1160.

PR #1161, opened 2020-10-30
Add a feature for the inventory-geo-script in the Mender-client

oleorhagen:
This adds a feature: 'mender-inventory-geo-script' which installs the
inventory-geo script as an inventory script when the Mender-client is installed.

By default, this is a sub-feature of the following features to go along with the
current setup which installs the script by default:

* mender-full
* mender-full-ubi
* mender-full-bios

Changelog: None

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client.inc
@@ -93,7 +93,7 @@ INSANE_SKIP_${PN}-ptest = "ldflags textrel"
 
 GO_IMPORT = "github.com/mendersoftware/mender"
 
-PACKAGECONFIG ?= "${@mender_feature_is_enabled("mender-client-install", "modules", "", d)}"
+PACKAGECONFIG ?= "${@mender_feature_is_enabled("mender-client-install", "modules inventory-network-scripts", "", d)}"
 
 PACKAGECONFIG_append = "${@bb.utils.contains('MENDER_FEATURES', 'mender-client-install', ' mender-client-install', '', d)}"
 PACKAGECONFIG_append = "${@bb.utils.contains('MENDER_FEATURES', 'mender-uboot', ' u-boot', '', d)}"
@@ -238,7 +238,8 @@ do_install() {
         systemd_unitdir=${systemd_unitdir} \
         install-bin \
         install-identity-scripts \
-        install-inventory-scripts \
+        install-inventory-local-scripts \
+        ${@bb.utils.contains('PACKAGECONFIG', 'inventory-network-scripts', 'install-inventory-network-scripts', '', d)} \
         install-systemd \
         ${@bb.utils.contains('PACKAGECONFIG', 'modules', 'install-modules', '', d)}
 
--- a/tests/acceptance/test_build.py
+++ b/tests/acceptance/test_build.py
@@ -31,6 +31,7 @@ from common import (
     get_local_conf_orig_path,
     make_tempdir,
     version_is_minimum,
+    reset_build_conf,
 )
 
 
@@ -1079,3 +1080,61 @@ deployed-test-dir9/*;renamed-deployed-test-dir9/ \
         assert re.search(test2_re, fstab, flags=re.MULTILINE) is not None
         assert re.search(test3_re, fstab, flags=re.MULTILINE) is not None
         assert re.search(test4_re, fstab, flags=re.MULTILINE) is not None
+
+    @pytest.mark.min_mender_version("2.5.0")
+    @pytest.mark.only_with_image("ext4")
+    def test_mender_inventory_network_scripts(self, prepared_test_build, bitbake_image):
+        """
+        Test the 'inventory-network-scripts' build feature configuration through
+        'PACKAGECONFIG' is working as expected.
+
+        This verifies that the 'inventory-network-scripts' option is a part
+        build, and also that the inventory scripts are not included when
+        removed.
+
+
+        The test only runs for sdimg, as the build image should not really matter here.
+        """
+
+        #
+        # Feature enabled
+        #
+        reset_build_conf(prepared_test_build["build_dir"])
+        build_image(
+            prepared_test_build["build_dir"],
+            prepared_test_build["bitbake_corebase"],
+            bitbake_image,
+            ['PACKAGECONFIG_append_pn-mender-client = " inventory-network-scripts"'],
+        )
+        rootfs = latest_build_artifact(
+            prepared_test_build["build_dir"], "core-image*.ext4"
+        )
+        assert len(rootfs) > 0, "rootfs not generated"
+
+        output = subprocess.check_output(
+            ["debugfs", "-R", "ls /usr/share/mender/inventory", rootfs]
+        )
+        assert (
+            b"mender-inventory-geo" in output,
+        ), "mender-inventory-network-scripts seems not to be a part of the image, like they should"
+
+        #
+        # Feature disabled
+        #
+        reset_build_conf(prepared_test_build["build_dir"])
+        build_image(
+            prepared_test_build["build_dir"],
+            prepared_test_build["bitbake_corebase"],
+            bitbake_image,
+            ['PACKAGECONFIG_remove_pn-mender-client = " inventory-network-scripts"'],
+        )
+        rootfs = latest_build_artifact(
+            prepared_test_build["build_dir"], "core-image*.ext4"
+        )
+        assert len(rootfs) > 0, "ext4 not generated"
+        output = subprocess.check_output(
+            ["debugfs", "-R", "ls /usr/share/mender/inventory", rootfs]
+        )
+        assert (
+            b"mender-inventory-geo" not in output,
+        ), "mender-inventory-network-scripts unexpectedly a part of the image"
--- a/meta-mender-core/recipes-mender/mender-client/mender-client_2.3.1.bb
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client_2.3.1.bb
@@ -26,3 +26,61 @@ LICENSE = "Apache-2.0 & BSD-2-Clause & BSD-3-Clause & ISC & MIT & OLDAP-2.8"
 
 DEPENDS += "xz"
 RDEPENDS_${PN} += "liblzma"
+
+do_install() {
+    oe_runmake \
+        -C ${B}/src/${GO_IMPORT} \
+        V=1 \
+        prefix=${D} \
+        bindir=${bindir} \
+        datadir=${datadir} \
+        sysconfdir=${sysconfdir} \
+        systemd_unitdir=${systemd_unitdir} \
+        install-bin \
+        install-identity-scripts \
+        install-inventory-scripts \
+        install-systemd \
+        ${@bb.utils.contains('PACKAGECONFIG', 'modules', 'install-modules', '', d)}
+
+    #install our prepared configuration
+    install -d ${D}/${sysconfdir}/mender
+    install -d ${D}/data/mender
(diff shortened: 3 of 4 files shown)


oleorhagen:
See also: https://github.com/mendersoftware/mender/pull/651

oleorhagen:
Like we discussed @kacf I think this might give me some ease of mind.

Not sure if the implementation is correct though.

meta-mender-core/recipes-mender/mender-client/mender-client.inc
```diff
@@ -239,6 +239,7 @@ do_install() {
         install-bin \
         install-identity-scripts \
         install-inventory-scripts \
+        ${@bb.utils.contains('MENDER_FEATURES_ENABLE', 'mender-inventory-geo-script', 'install-inventory-geo-script', '', d)} \
```
kacf, line 242: This doesn't need to be global. Instead, see how `modules` is used in `PACKAGECONFIG`, two lines below.

kacf (changes requested):

meta-mender-core/recipes-mender/mender-client/mender-client.inc
lluiscampos, line 242: As we are discussing at https://github.com/mendersoftware/mender/pull/651/files#r515787653 the target is likely to be called `install-network-inventory-scripts`; therefore the feature here should also align with "network" rather than the specific script (this allows for the future to have more scripts in this category).

lluiscampos (changes requested):
Nice initiative.

The feature needs to be added [here](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/classes/mender-maybe-setup.bbclass#L20). See also comment below wrt naming.

oleorhagen:
Note to self:
Remember to update the docs along with this (!)

lluiscampos (commented):
Some nit picks from my side:

* Rename the commit message with the new name of the feature
* Add a changelog tag. It would be super nice for the changelog to illustrate how to opt-out of the network scripts.
* Still missing from [mender-features](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/classes/mender-maybe-setup.bbclass#L20) as I pointed earlier.

Cherry on top: what about a test for this new feature?

lluiscampos (changes requested):
> Some nit picks from my side:
> 
>     * Rename the commit message with the new name of the feature
> 
>     * Add a changelog tag. It would be super nice for the changelog to illustrate how to opt-out of the network scripts.
> 
>     * Still missing from [mender-features](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/classes/mender-maybe-setup.bbclass#L20) as I pointed earlier.
> 
> 
> Cherry on top: what about a test for this new feature?

oleorhagen:
> Some nit picks from my side:
> 
>     * Rename the commit message with the new name of the feature
Great, thanks. Done locally.
> 
>     * Add a changelog tag. It would be super nice for the changelog to illustrate how to opt-out of the network scripts.
> 
This is what I now added locally, let's discuss it prior to pushing:
```
New Mender feature: 'mender-inventory-network-scripts' added. This
feature enables the installation of the third-party inventory-network scripts in the
Mender-client. The feature is enabled by default in all 'mender-full*' setups,
but must be added explicitly to 'MENDER_FEATURES_ENABLE' if you are building a
Mender setup yourself from the base features.
```
Feel free to add/remove change at will :)

>     * Still missing from [mender-features](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/classes/mender-maybe-setup.bbclass#L20) as I pointed earlier.
> 
Fixed (locally).
> 
> Cherry on top: what about a test for this new feature?

Yes, lets!

lluiscampos:
> This is what I now added locally, let's discuss it prior to pushing:
> 
> ```
> New Mender feature: 'mender-inventory-network-scripts' added. This
> feature enables the installation of the third-party inventory-network scripts in the
> Mender-client. The feature is enabled by default in all 'mender-full*' setups,
> but must be added explicitly to 'MENDER_FEATURES_ENABLE' if you are building a
> Mender setup yourself from the base features.
> ```

Looks good. Not very strong opinion, but I would point to how to opt-out from the default rather that how to opt in in case of building from base features (rare case, I think). Hence I suggest to have the last sentence something like:

```
The feature is enabled by default in all 'mender-full*' setups, and can be
disabled using MENDER_FEATURES_DISABLE_append = " mender-inventory-network-scripts".
```

(please verify the syntax)

What do you think?

oleorhagen:
> > This is what I now added locally, let's discuss it prior to pushing:
> > ```
> > New Mender feature: 'mender-inventory-network-scripts' added. This
> > feature enables the installation of the third-party inventory-network scripts in the
> > Mender-client. The feature is enabled by default in all 'mender-full*' setups,
> > but must be added explicitly to 'MENDER_FEATURES_ENABLE' if you are building a
> > Mender setup yourself from the base features.
> > ```
> 
> Looks good. Not very strong opinion, but I would point to how to opt-out from the default rather that how to opt in in case of building from base features (rare case, I think). Hence I suggest to have the last sentence something like:
> 
> ```
> The feature is enabled by default in all 'mender-full*' setups, and can be
> disabled using MENDER_FEATURES_DISABLE_append = " mender-inventory-network-scripts".
> ```
> 
> (please verify the syntax)
> 
> What do you think?

I like it. Rewording.

meta-mender-core/classes/mender-maybe-setup.bbclass
```diff
@@ -68,6 +68,9 @@ python() {
 
         # Enable the testing/* layers and functionality
         'mender-testing-enabled',
+
+        # Enable installation of the Mender-client network scripts
+        'mender-inventory-network-scripts'
```
kacf, line 73: I think you misunderstood, we should not use `MENDER_FEATURES` at all for this. `MENDER_FEATURES` are for features that affect the global build. So this, and all other references in class files, should be removed. See other comments for how to do it with `PACKAGECONFIG`.

meta-mender-core/recipes-mender/mender-client/mender-client.inc
```diff
@@ -98,6 +98,7 @@ PACKAGECONFIG ?= "${@mender_feature_is_enabled("mender-client-install", "modules
 PACKAGECONFIG_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-client-install', ' mender-client-install', '', d)}"
 PACKAGECONFIG_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-uboot', ' u-boot', '', d)}"
 PACKAGECONFIG_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-grub', ' grub', '', d)}"
+PACKAGECONFIG_append = "${@bb.utils.contains('DISTRO_FEATURES', 'mender-network-scripts', 'mender-inventory-network-scripts', '', d)}"
```
kacf, line 101: Instead, of doing this, locate the line 5 lines prior and do this:

```diff
-PACKAGECONFIG ?= "${@mender_feature_is_enabled("mender-client-install", "modules", "", d)}"
+PACKAGECONFIG ?= "${@mender_feature_is_enabled("mender-client-install", "modules inventory-network-scripts", "", d)}"
```

Note that I switched `mender-inventory-network-scripts` for `inventory-network-scripts`. There is no need to prefix here, since package configs are localized to the recipe, and we are already inside `mender-client`.

tests/acceptance/test_build.py
```diff
@@ -1079,3 +1079,60 @@ def test_extra_parts(self, latest_part_image, prepared_test_build, bitbake_image
...
+
+        #
+        # Explicitly disable the feature
+        #
+        reset_build_conf(prepared_test_build["build_dir"])
+        build_image(
+            prepared_test_build["build_dir"],
+            prepared_test_build["bitbake_corebase"],
+            bitbake_image,
+            ['MENDER_FEATURES_DISABLE_append = " mender-inventory-network-scripts"'],
```
kacf, line 1128: This should be changed to `PACKAGECONFIG_remove_pn-mender-client = "inventory-network-scripts"`.

meta-mender-core/recipes-mender/mender-client/mender-client.inc
```diff
@@ -238,7 +239,8 @@ do_install() {
         systemd_unitdir=${systemd_unitdir} \
         install-bin \
         install-identity-scripts \
-        install-inventory-scripts \
+        install-inventory-local-scripts \
+        ${@bb.utils.contains('PACKAGECONFIG', 'mender-inventory-network-scripts', 'install-inventory-network-scripts', '', d)} \
```
kacf, line 243: This should be renamed too.

tests/acceptance/test_build.py
```diff
@@ -1079,3 +1079,60 @@ def test_extra_parts(self, latest_part_image, prepared_test_build, bitbake_image
...
+        #
+        # Feature enabled by default
+        #
+        reset_build_conf(prepared_test_build["build_dir"])
+        build_image(
+            prepared_test_build["build_dir"],
+            prepared_test_build["bitbake_corebase"],
+            bitbake_image,
+            # The feature should be enabled by default in all Mender-client builds
+            # ['MENDER_FEATURES_DISABLE_append = " mender-inventory-network-scripts"'],
```
kacf, line 1108: `PACKAGECONFIG_append_pn-mender-client = "inventory-network-scripts"`

tests/acceptance/test_build.py
```diff
@@ -1079,3 +1079,60 @@ def test_extra_parts(self, latest_part_image, prepared_test_build, bitbake_image
         assert re.search(test2_re, fstab, flags=re.MULTILINE) is not None
         assert re.search(test3_re, fstab, flags=re.MULTILINE) is not None
         assert re.search(test4_re, fstab, flags=re.MULTILINE) is not None
+
+    @pytest.mark.min_mender_version("2.5.0")
+    @pytest.mark.only_with_image("sdimg")
+    def test_mender_inventory_network_scripts(
```
kacf, line 1085: Good test!

kacf (changes requested):

meta-mender-core/classes/mender-maybe-setup.bbclass
lluiscampos: I think this was my bad, then. I misunderstood too...

lluiscampos (commented):
It seems like I misunderstood and lead you to the wrong place, @oleorhagen I'll let Kristian do the review, then.

Just remember to adapt the `Changelog` in the upcoming rework.

oleorhagen:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/212231952

kacf (approved):

tests/acceptance/test_build.py
```diff
@@ -1079,3 +1080,60 @@ def test_extra_parts(self, latest_part_image, prepared_test_build, bitbake_image
...
+        )
+        sdimg = latest_build_artifact(
+            prepared_test_build["build_dir"], "core-image*.sdimg"
+        )
+        r = subprocess.run(
+            f"mender-artifact cat {sdimg}:/usr/share/mender/inventory/mender-inventory-geo"
+        )
+        assert (
+            r.status_code != 0
+        ), "mender-inventory-network scripts unexpectedly a part of the image"
```
lluiscampos, line 1139: Nitpick:
```suggestion
        ), "mender-inventory-network-scripts unexpectedly a part of the image"
```

Like above.

lluiscampos (approved):

oleorhagen:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/217383964

PR #1162, opened 2020-10-30
mender: Reestablish labels on the root filesystems.

drewmoseley:

### Description

Add labels to the rootfs partitions

### DIFF
--- a/meta-mender-core/classes/mender-part-images.bbclass
+++ b/meta-mender-core/classes/mender-part-images.bbclass
@@ -125,8 +125,8 @@ EOF
     fi
 
     cat >> "$wks" <<EOF
-part --source rawcopy --sourceparams="file=${IMGDEPLOYDIR}/${IMAGE_LINK_NAME}.${ARTIFACTIMG_FSTYPE}" --ondisk "$ondisk_dev" --align $alignment_kb --fixed-size ${MENDER_CALC_ROOTFS_SIZE}k $part_type_params
-part --source rawcopy --sourceparams="file=${IMGDEPLOYDIR}/${IMAGE_LINK_NAME}.${ARTIFACTIMG_FSTYPE}" --ondisk "$ondisk_dev" --align $alignment_kb --fixed-size ${MENDER_CALC_ROOTFS_SIZE}k $part_type_params
+part --source rawcopy --sourceparams="file=${IMGDEPLOYDIR}/${IMAGE_LINK_NAME}.${ARTIFACTIMG_FSTYPE}" --fstype=${ARTIFACTIMG_FSTYPE} --label primary --ondisk "$ondisk_dev" --align $alignment_kb --fixed-size ${MENDER_CALC_ROOTFS_SIZE}k $part_type_params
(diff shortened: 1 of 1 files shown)


meta-mender-core/classes/mender-part-images.bbclass
```diff
@@ -125,8 +125,8 @@ EOF
     fi
 
     cat >> "$wks" <<EOF
-part --source rawcopy --sourceparams="file=${IMGDEPLOYDIR}/${IMAGE_LINK_NAME}.${ARTIFACTIMG_FSTYPE}" --ondisk "$ondisk_dev" --align $alignment_kb --fixed-size ${MENDER_CALC_ROOTFS_SIZE}k $part_type_params
-part --source rawcopy --sourceparams="file=${IMGDEPLOYDIR}/${IMAGE_LINK_NAME}.${ARTIFACTIMG_FSTYPE}" --ondisk "$ondisk_dev" --align $alignment_kb --fixed-size ${MENDER_CALC_ROOTFS_SIZE}k $part_type_params
+part --source rawcopy --sourceparams="file=${IMGDEPLOYDIR}/${IMAGE_LINK_NAME}.${ARTIFACTIMG_FSTYPE}" --fstype=${ARTIFACTIMG_FSTYPE} --label primary --ondisk "$ondisk_dev" --align $alignment_kb --fixed-size ${MENDER_CALC_ROOTFS_SIZE}k $part_type_params
+part --source rawcopy --sourceparams="file=${IMGDEPLOYDIR}/${IMAGE_LINK_NAME}.${ARTIFACTIMG_FSTYPE}" --fstype=${ARTIFACTIMG_FSTYPE} --label secondary --ondisk "$ondisk_dev" --align $alignment_kb --fixed-size ${MENDER_CALC_ROOTFS_SIZE}k $part_type_params
```
kacf, line 129: Looks good, but should we do the same for the data partition below?

kacf (approved):

meta-mender-core/classes/mender-part-images.bbclass
drewmoseley: A volume label is already applied in the IMAGE_CMD_dataimg() function so it's not needed here.

lluiscampos:
@drewmoseley any any other reader for future reference,

This change modifies the checksum on the filesystem, making it not mach anymore the checksum from the Mender Artifact generated by yocto in the same build. It has been reverted in https://github.com/mendersoftware/meta-mender/pull/1219.

The reason for not caching this problem before is that the specific test for checksum verification was wrongly skip in modern yocto branches. See https://github.com/mendersoftware/mender-image-tests/pull/32 for more details on the test framework error.

clementperon:
@lluiscampos I'm interested to have label for the rootfs, if the checksum match is it acceptable to restore the label for rootfs?

kacf:
I suspect some wrong assumptions were made when this was first implemented. Labels cannot be used because they are tied to the filesystem. To be usable with Mender, they need to be tied to the partition entry. Consider this scenario:

```
 ┌──────────┬─────────┬──────────┐
 │Partition │ sda2    │ sda3     │
 │Label     │ primary │ secondary│
 └──────────┴─────────┴──────────┘
```

If it was updated, it might look like this:

```
 ┌──────────┬─────────┬──────────┐
 │Partition │ sda2    │ sda3     │
 │Label     │ primary │ primary  │
 └──────────┴─────────┴──────────┘
```

Because the label follows the filesystem, not the partition.

drewmoseley:
@kacf does that mean this patch needs to be reverted?

kacf:
> So using the PARTUUID instead of the UUID leaves to the same issue right?

Using PARTUUID instead of UUID solves the problem. Similarly, using `--partlabel` instead of `--label` would solve the problem for labels, but I haven't checked if this actually exists.

PR #1081, opened 2020-09-10
Add recipes for Mender 2.3.1 components to sumo

lluiscampos:
Changelog: Add mender 2.2.1 recipe
Changelog: Add mender-artifact 3.3.1 recipe
Changelog: Remove mender 2.2.0b1 recipe
Changelog: Remove mender-artifact 3.3.0b1 recipe

PR #1167, opened 2020-11-02
Efi provider

drewmoseley:

### Description

Fix the EFI_PROVIDER variable assignment.

### DIFF
--- a/meta-mender-core/classes/mender-setup-grub.inc
+++ b/meta-mender-core/classes/mender-setup-grub.inc
@@ -9,12 +9,7 @@ PREFERRED_RPROVIDER_grub-editenv ?= "grub"
 PREFERRED_PROVIDER_virtual/grub-bootconf ?= "grub-mender-grubenv"
 PREFERRED_RPROVIDER_virtual/grub-bootconf ?= "grub-mender-grubenv"
 
-# Set EFI_PROVIDER.  Not all MACHINE configs use it but notably
-# intel-corei7-64 does and without this we use the default of systemd-boot.
-EFI_PROVIDER ?= "${_MENDER_EFI_PROVIDER_DEFAULT}"
-_MENDER_EFI_PROVIDER_DEFAULT = ""
-_MENDER_EFI_PROVIDER_DEFAULT_mender-grub = "grub-efi"
-_MENDER_EFI_PROVIDER_DEFAULT_mender-grub_mender-bios = ""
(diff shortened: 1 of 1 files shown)


meta-mender-core/classes/mender-setup-grub.inc
```diff
@@ -9,12 +9,7 @@ PREFERRED_RPROVIDER_grub-editenv ?= "grub"
 PREFERRED_PROVIDER_virtual/grub-bootconf ?= "grub-mender-grubenv"
 PREFERRED_RPROVIDER_virtual/grub-bootconf ?= "grub-mender-grubenv"
 
-# Set EFI_PROVIDER.  Not all MACHINE configs use it but notably
-# intel-corei7-64 does and without this we use the default of systemd-boot.
-EFI_PROVIDER ?= "${_MENDER_EFI_PROVIDER_DEFAULT}"
-_MENDER_EFI_PROVIDER_DEFAULT = ""
-_MENDER_EFI_PROVIDER_DEFAULT_mender-grub = "grub-efi"
-_MENDER_EFI_PROVIDER_DEFAULT_mender-grub_mender-bios = ""
+EFI_PROVIDER ?= "${@mender_feature_is_enabled("mender-bios", "", "grub-efi", d)}"
```
kacf, line 12: This is not exactly the same as before. If neither `mender-grub` nor `mender-bios` is set, the provider will be set to `grub-efi`, before this patch it was set to empty. I don't think it will have any ill effects, but this violates the criterion we're trying to keep, that the layer does not introduce any changes unless any of our features are enabled. Maybe it should be reworked to check both feature flags?

drewmoseley: Hmm.  I see it being set with"?=" in live-vm.bbclass and set to other things in other files, including systemd-boot.  Ideally we would only _ever_ touch it if either `mender-grub` or `mender-bios` were set but I don't know how to do that without overrides which are apparently causing problems for users due to the order of processing.  Any suggestions?

kacf: I have a possible workaround, but it involves modifying the heart of how we populate the `DISTROOVERRIDES` variable. The good thing is that this may at least partially address some of the comments made in #1081.

I still need to experiment a bit more with it though, and my current test case for this is a bit contrived. Can you give me the exact circumstances where this happens for you, so I can experiment with it?

kacf: Check out #1175.

kacf:
Superseded by #1175.

PR #1175, opened 2020-11-10
Parsing fixes and stop using DISTRO_FEATURES.

kacf:
```
commit 6189f2d3bc4630de1a8b8163ed29187990b633f8
Date:   Tue Nov 10 09:32:21 2020

    Stop (ab)using `DISTRO_FEATURES` and use `MENDER_FEATURES` instead.
    
    Changelog: Features that are added or removed using
    `MENDER_FEATURES_ENABLE` or `MENDER_FEATURES_DISABLE` now end up in
    the `MENDER_FEATURES` variable instead of the `DISTRO_FEATURES`
    variable. Code of the kind:
    ```
    ${@bb.utils.contains('DISTRO_FEATURES', 'mender-uboot', 'true', 'false', d)}
    ```
    needs to change to look in `MENDER_FEATURES` instead.
    
commit 5466a47e8b05d093f5312d7e61bdfff16182909e
Date:   Tue Nov 10 09:26:01 2020

    Fix parsing order so `inherit` works with `MENDER_FEATURES`.
    
    It's two combined fixes that solve the problem: First, we need to move
    the inclusion of `mender-setup` so that it comes after the definition
    of all variables that add to `MENDER_FEATURES`. Second, we need to
    populate `DISTROOVERRIDES` in a variable assignment instead of in an
    anonymous Python function. The latter is executed at the end, whereas
    assignment is executed when it is needed (early).
    
    Changelog: Fix a parsing issue where `inherit` could not be used with
    variables that had been defined with overrides that depended on
    `MENDER_FEATURES_ENABLE`. One example would be:
    ```
    MYVAR_mender-grub = "grub-efi"
    inherit ${MYVAR}
    ```
    
```


kacf:
@drewmoseley: This is my proposed fix for the issue in #1167. With this I think your fix is not needed. It is the commit:
```
Fix parsing order so `inherit` works with `MENDER_FEATURES`.
```
which is important for you, the other one is just additional cleanup to improve upon #1083.

Could you ask your customer whether this solves the problem for them?

tests/acceptance/fixtures.py
```diff
@@ -579,21 +579,21 @@ def test_foo():
 
 
 @pytest.fixture(autouse=True)
-def only_with_distro_feature(request, bitbake_variables):
-    """Fixture that enables use of `only_with_distro_feature(feature1, feature2)` mark.
+def only_with_mender_feature(request, bitbake_variables):
+    """Fixture that enables use of `only_with_mender_feature(feature1, feature2)` mark.
```
lluiscampos, line 583: You probably know, but some tests need an update accordingly.

lluiscampos:
From a design point of view, this looks good!

tests/acceptance/fixtures.py
kacf: Lol, I went out of my way to make sure all the `DISTRO_FEATURES` variables were changed, and completely forgot this part!

lluiscampos: Hehe. I first thought that it was so obvious that I shouldn't comment on it... :smile:

drewmoseley (approved):
I like it.

drewmoseley:
@kacf  the reporting user indicated this fix worked for them.

kacf:
Excellent! I'll fix the issues Lluis pointed out ASAP.

tests/acceptance/fixtures.py
```diff
@@ -579,23 +579,30 @@ def test_foo():
...
```
lluiscampos, line 583: Reminder (because I see you just force-pushed)

We agreed on the call to leave `only_with_distro_feature` fixture as-is, to be used by old branches of meta-mender and mender-convert, and just create a new one `only_with_mender_feature` for branches of meta-mender that actually have changed this (master, dunfell)

lluiscampos: Disregard, I completely miss read your code

tests/acceptance/fixtures.py
```diff
@@ -579,23 +579,30 @@ def test_foo():
...
+        current += bitbake_variables.get("DISTRO_FEATURES", "").strip().split()
         if not all([feature in current for feature in features]):
             pytest.skip(
                 "no supported distro feature in {} "
                 "(supports {})".format(", ".join(current), ", ".join(features))
             )
+
+
+# Deprecated.
+def only_with_distro_feature(request, bitbake_variables):
```
lluiscampos, line 607: But this is not actually a fixture, how is it supposed to work for the other branches?

tests/acceptance/fixtures.py
```diff
@@ -579,23 +579,30 @@ def test_foo():
...
-       @pytest.mark.only_with_distro_feature('mender-uboot')
+       @pytest.mark.only_with_mender_feature('mender-uboot')
        def test_foo():
            # executes only if mender-uboot feature is enabled
            pass
 
     """
 
-    mark = request.node.get_closest_marker("only_with_distro_feature")
+    mark = request.node.get_closest_marker("only_with_mender_feature")
```
lluiscampos, line 593: This won't work for tests using `only_with_distro_feature` marker, will it?

kacf: Of course, this was a totally lame attempt, I must be tired! It should be fixed now!

lluiscampos (approved):

PR #1190, opened 2020-11-25
Switch to a more relaxed license checksum checker for the Git recipe.

kacf:
This doesn't actually cover all license we use, so it should not be
used for production recipes, but should be acceptable for the Git
recipes.

Changelog: None

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client_git.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client_git.inc
@@ -52,49 +52,24 @@ def mender_version_from_preferred_version(d, srcpv):
 SRC_URI = "git://github.com/mendersoftware/mender;protocol=https;branch=${MENDER_BRANCH}"
 
 # DO NOT change the checksum here without make sure that ALL licenses (including
-# dependencies) are included in the LICENSE variable below.
+# dependencies) are included in the LICENSE variable below. Note that for
+# releases, we must check the LIC_FILES_CHKSUM.sha256 file, not the LICENSE
+# file.
 def mender_license(branch):
-    if branch == "1.7.x":
+    import re
+    if branch.startswith("1."):
         return {
-                   "md5": "5632b9f17043c6f5f532501778595c78",
                    "license": "Apache-2.0 & BSD-2-Clause & BSD-3-Clause & ISC & MIT & OLDAP-2.8",
         }
-    elif branch.startswith("1."):
+    elif re.match(r"2\.[0-3]\.x", branch) is not None:
         return {
-                   "md5": "debbe5e440f2e65465e86b25fc7c9fcc",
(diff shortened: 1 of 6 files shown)


oleorhagen (approved):
Yes! 

Should it have a small comment to it though?

meta-mender-core/recipes-mender/mender-client/mender-client_git.inc
```diff
@@ -54,47 +54,20 @@ SRC_URI = "git://github.com/mendersoftware/mender;protocol=https;branch=${MENDER
...
-                   "md5": "5649992d13a6fda40abfac7730a62b07",
-                   "license": "Apache-2.0 & BSD-2-Clause & BSD-3-Clause & ISC & MIT & OLDAP-2.8 & OpenSSL",
-        }
     else:
         return {
-                   "md5": "76d17470649a8fdd1a679a9c56a35c15",
                    "license": "Apache-2.0 & BSD-2-Clause & BSD-3-Clause & ISC & MIT & OLDAP-2.8 & OpenSSL",
         }
-LIC_FILES_CHKSUM = "file://src/github.com/mendersoftware/mender/LIC_FILES_CHKSUM.sha256;md5=${@mender_license(d.getVar('MENDER_BRANCH'))['md5']}"
+LIC_FILES_CHKSUM = "file://src/github.com/mendersoftware/mender/LICENSE;md5=7fd64609fe1bce47db0e8f6e3cc6a11d"
```
lluiscampos, line 70: This will need an update once a year. Quite an improvement :tada:

lluiscampos (approved):
It would be nice to extend to `mender-convert` and `mender-shell`, but this one is the most important one, we can do the other ones later or separately.

meta-mender-core/recipes-mender/mender-client/mender-client_git.inc
kacf: Indeed!

kacf:
> Should it have a small comment to it though?

Anything in particular? The comment above the function is still correct, it's just the frequency of changes to the file that has gone down because it is pointed at `LICENSE` instead of `LIC_FILES_CHKSUM.sha256`.

kacf:
FYI, I added the same thing for mender-shell and mender-artifact.

oleorhagen:
> > Should it have a small comment to it though?
> 
> Anything in particular? The comment above the function is still correct, it's just the frequency of changes to the file that has gone down because it is pointed at `LICENSE` instead of `LIC_FILES_CHKSUM.sha256`.

No need to fuzz about this. My thinking was just, for outsiders to understand why we're using two different files for the checksum, but I guess it is understandable still :)

Just ignore me

kacf:
> No need to fuzz about this. My thinking was just, for outsiders to understand why we're using two different files for the checksum, but I guess it is understandable still :)

No, actually, that's a very good point, I didn't think about it! :-)
I will add it, because it is important when doing releases!

PR #1197, opened 2020-12-01
Dunfell image tests bump

oleorhagen:


oleorhagen:
Will be done at a later point, during the release

PR #1213, opened 2020-12-09
mender-shell: align recipe with other Mender go recipes

lluiscampos:
Namely:
* inherit go instead of go-mod
* inherit pkgconfig instead of manually specifying dependency
* fix do_compile function
* reorder declarations
* use term "add-on" in description

### DIFF
--- a/meta-mender-core/recipes-mender/mender-shell/mender-shell.bb
+++ b/meta-mender-core/recipes-mender/mender-shell/mender-shell.bb
@@ -1,4 +1,4 @@
-DESCRIPTION = "Mender program for remote terminal access."
+DESCRIPTION = "Mender add-on for remote terminal access."
 HOMEPAGE = "https://mender.io"
 
 SRC_URI = "git://github.com/mendersoftware/mender-shell.git;protocol=https;branch=master"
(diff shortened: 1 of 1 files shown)


kacf (approved):

meta-mender-core/recipes-mender/mender-shell/mender-shell.bb
```diff
@@ -14,16 +14,22 @@ PV = "0.1+git${SRCPV}"
...
+inherit go-ptest
+inherit pkgconfig
+inherit systemd
+
 GO_IMPORT = "github.com/mendersoftware/mender-shell"
 
 do_compile() {
-    oe_runmake V=1
+    oe_runmake \
+        -C ${B}/src/${GO_IMPORT} \
```
oleorhagen, line 31: Just curious:
Why is this needed, like it was working for me previously?

oleorhagen (approved):
LGTM

Just wondering, what was the actual issues:

* go-mod not appropriate?
* Is it decided for sure that it will be an add-on?

:up:

lluiscampos:
> LGTM
> 
> Just wondering, what was the actual issues:
> 
>     * go-mod not appropriate?

This is the actual issue, for zeus an older, that `go-mod` class does not exist. In any case, I think it makes sense to compile it the same way as mender-client or mender-artifact (which also support go modules, but we still compile them the old way).

> 
>     * Is it decided for sure that it will be an add-on?

Right. This I don't know what is the "official" decision. But we have been using the add-on terminology in the mender-docs, so I think it makes sense to not call it "program" here.

meta-mender-core/recipes-mender/mender-shell/mender-shell.bb
lluiscampos: I don't know exactly why, I haven't inspected the bitbake code this time. But with `inherint go-mod` this was not necessary, while with `inherit go` turned out necessary. Also note that it was already used for `do_install` below.

PR #1221, opened 2020-12-11
MEN-4242: mender-shell: generate and install mender-shell.conf

lluiscampos:
Changelog: mender-shell: generate and install mender-shell.conf with
required fields. `ServerURL` can be configured setting yocto variable
`MENDER_SERVER_URL`, same as used by mender-client recipe. If a
`mender-shell.conf` file is found in the `SRC_URI`, the contents will be
merged.

Note: includes test from https://github.com/mendersoftware/mender-image-tests/pull/33

In addition:
* mender-client: fix QA Issue: invalid PACKAGECONFIG: inventory-network-scripts
Changelog: Title

### DIFF
--- a/meta-mender-core/recipes-mender/mender-shell/mender-shell.bb
+++ b/meta-mender-core/recipes-mender/mender-shell/mender-shell.bb
@@ -17,7 +17,10 @@ LIC_FILES_CHKSUM = "file://src/${GO_IMPORT}/LICENSE;md5=7fd64609fe1bce47db0e8f6e
 DEPENDS_append = " glib-2.0"
 RDEPENDS_${PN} = "glib-2.0"
 
+MENDER_SERVER_URL ?= "https://docker.mender.io"
 SYSTEMD_AUTO_ENABLE ?= "enable"
+
+B = "${WORKDIR}/build"
 
 inherit go
 inherit go-ptest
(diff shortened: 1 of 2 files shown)


meta-mender-core/recipes-mender/mender-shell/mender-shell.bb
```diff
@@ -32,6 +35,28 @@ do_compile() {
         V=1
 }
 
+python do_prepare_mender_shell_conf() {
+    import json
+
+    # If a mender-shell.conf has been provided in SRC_URI, use this one instead
+    src_conf = os.path.join(d.getVar("WORKDIR"), "mender-shell.conf")
+    if os.path.exists(src_conf):
+        bb.debug(1, "mender-shell.conf already present in ${WORKDIR}, return.")
```
kacf, line 44: Returning is not what you do though. And the behavior in the client is to still add the key, so I reckon we should do the same here. So in other words, only the message here is wrong.

meta-mender-core/recipes-mender/mender-shell/mender-shell.bb
```diff
@@ -32,6 +35,28 @@ do_compile() {
...
+
+    mender_shell_conf = {}
```
kacf, line 46: Should this be populated with a `json.load` from the original file, if there is one?

Update: See my comment below as well.

meta-mender-core/recipes-mender/mender-shell/mender-shell.bb
```diff
@@ -43,6 +68,14 @@ do_install() {
         systemd_unitdir=${systemd_unitdir} \
         install-bin \
         install-systemd
+
+    # install configuration
+    mkdir -p  ${D}/${sysconfdir}/mender
+    if [ -f ${WORKDIR}/mender-shell.conf ]; then
+        install -m 0600 ${WORKDIR}/mender-shell.conf ${D}/${sysconfdir}/mender/mender-shell.conf
+    else
+        install -m 0600 ${B}/mender-shell.conf ${D}/${sysconfdir}/mender/mender-shell.conf
```
kacf, line 77: Oh, I see, you choose one or the other. I think it's better to merge though, do you agree?

lluiscampos: Yes, it is a better solution. For example the user might have other fields, but missing the server.

I am reworking, thanks!

kacf (approved):

PR #1235, opened 2020-12-16
Add mender-connect.service to SYSTEMD_SERVICE.

kacf:
Without this, SYSTEMD_AUTO_ENABLE has no effect.

Changelog: None


merlin-northern (approved):
with every PR from you I am learning something new.

meta-mender-core/recipes-mender/mender-shell/mender-shell.bb
```diff
@@ -20,6 +20,7 @@ RDEPENDS_${PN} = "glib-2.0"
 MENDER_SERVER_URL ?= "https://docker.mender.io"
 MENDER_SHELL_USER ??= "nobody"
 SYSTEMD_AUTO_ENABLE ?= "enable"
+SYSTEMD_SERVICE_${PN} = "mender-shell.service"
```
lluiscampos, line 23: Got it, thanks.

We should then remove these lines from below, right?
```
FILES_${PN}_append += "\
    ${systemd_unitdir}/system/mender-shell.service \
"
```

lluiscampos (approved):

meta-mender-core/recipes-mender/mender-shell/mender-shell.bb
kacf: I think that also needs to be there. `FILES` is required to include the file in the package. `SYSTEMD_SERVICE` is for marking this particular file as a service to be enabled.

kacf:
> with every PR from you I am learning something new.

I learned it myself about five minutes before this PR was posted. :-)

kacf:
Interesting. This has generated a lot of failures with messages like:

```
Error org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
```

I have no idea why, but this PR enables mender-shell, which will start pounding mender-client with requests. Maybe this somehow breaks it? Should try with https://github.com/mendersoftware/mender-shell/pull/26 to see if the timing fix there fixes it, since currently mender-shell is asking once every 5ms, instead of 5s.

kacf:
Started new pipeline with https://github.com/mendersoftware/mender-shell/pull/26 [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/230810028). If this passes, this must *not* be merged, because it will still fail with master.

meta-mender-core/recipes-mender/mender-shell/mender-shell.bb
lluiscampos: You are right (as usual). I got confused as I couldn't grep for this in `mender-client` equivalent recipe.

Not important, but if you have to rework anyway,

I suggest then that we move the `FILES_${PN}_append` up here, next to the `SYSTEMD_SERVICE_${PN}`, and possibly add any other missing files (the conf file is added at `do_install` but not explicitly included in `FILES_${PN}`) in a way that this recipe is closer to the [mender-client one](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/recipes-mender/mender-client/mender-client.inc#L30).

lluiscampos (approved):

kacf:
Didn't even compile. I guess we should wait until https://github.com/mendersoftware/mender-shell/pull/26 is merged.

merlin-northern:
> Didn't even compile. I guess we should wait until [mendersoftware/mender-shell#26](https://github.com/mendersoftware/mender-shell/pull/26) is merged.

it is coming! not my fault I got the toughest reviewer. talk some sense into him.

kacf:
Original pipeline, for reference: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/233142369

kacf:
@lluiscampos, @merlin-northern: I added two new commits here. Second look?

lluiscampos (approved):
Thanks for moving the `SYSTEMD_AUTO_ENABLE = "disable"` here from the actual test code. This has bugged me in the past.

kacf:
This already passed in https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/237561917, and the latest update is just to please the Black formatter. Merging.

PR #1236, opened 2020-12-17
mender-client.inc: support git build for 2.4.x and earlier

lluiscampos:
By conditionally expecting or not an install-dbus target in the
Makefile.

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client.inc
@@ -246,7 +246,6 @@ do_install() {
         install-bin \
         install-identity-scripts \
         install-systemd \
-        ${@bb.utils.contains('PACKAGECONFIG', 'dbus', 'install-dbus', '', d)} \
         ${@bb.utils.contains('PACKAGECONFIG', 'modules', 'install-modules', '', d)}
 
     # install inventory scripts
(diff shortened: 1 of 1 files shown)


lluiscampos:
Something like this is required for backports to `mender` repo, like https://github.com/mendersoftware/mender/pull/689 and https://github.com/mendersoftware/mender/pull/688.

@kacf My other idea was to split out mender-client.inc to something like mender-client-pre-2.5.inc and then conditionally include, but I think that this PR approach was cleaner and still maintainable as per today.

lluiscampos:
So this PR made the recipe build, but of course `TestBuild.test_build_nodbus` still failed. Even with https://github.com/mendersoftware/meta-mender/pull/1237 the test will still fail, as 2.4.x 2.3.x are considered "master". But this won't be a problem for the release, as any buid candidate 2.4.2-build1 will not run this test (will require mender 2.5.0).

@kacf Am I correct? If so, I suggest merging as-is and not add further fixes.

kacf:
I had to stare at this for quite some time. I don't really understand why it triggers on that test case. The reason I set it to `1.0.0` was that that removing something from `PACKAGECONFIG` should always be possible, even if it didn't exist there originally. But the results are clear, it fails only there. Weird...

In general I feel like these conditional features and the Git recipe is getting closer to a level which is unmaintainable. We have so many exceptions now, and it breaks so often. We should maybe start thinking about redesigning how it's all put together, maybe by having separate Git recipes for each release series or something like that. Do you agree?

Other than that I agree with your observations.

kacf (approved):

lluiscampos:
Thanks for checking.

> We should maybe start thinking about redesigning how it's all put together, maybe by having separate Git recipes for each release series or something like that. Do you agree?

I completely agree. This is not going to get simpler as we add add-ons to Mender...

PR #1237, opened 2020-12-18
[tests/acceptance] test_build_nodbus: set min mender 2.5.0

lluiscampos:


lluiscampos:
Stopped the build

kacf (approved):
As mentioned in #1236 I don't fully understand why this is needed, but you can't argue with the test results.

PR #1246, opened 2020-12-21
Amend shell syntax conditional dbus + revert skip test

lluiscampos:
Amend of #1236:

```
Fix shell script parse error when no dbus in PACKAGECONFIG

Error introduced at 4f8b02a, and it only shows when building with
preferred version master (grep succeed) and removing dbus from
PACKAGECONFIG.
```

And  and revert of #1237:
```
Revert "[tests/acceptance] test_build_nodbus: set min mender 2.5.0"

This reverts commit 76a2d441692d741701b7cd4ac60ae2d7db0a4c71.

This commit was not needed in the first place, the error that motivated
it was miss-understood. See previous commit fixing the shell syntax.
```

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client.inc
@@ -268,7 +268,7 @@ do_install() {
 
     # install dbus files
     if grep -q install-dbus ${B}/src/${GO_IMPORT}/Makefile; then
-        if ${@bb.utils.contains('PACKAGECONFIG', 'dbus', 'true', '', d)}; then
+        if ${@bb.utils.contains('PACKAGECONFIG', 'dbus', 'true', 'false', d)}; then
             oe_runmake \
                 -C ${B}/src/${GO_IMPORT} \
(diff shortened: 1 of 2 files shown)


lluiscampos:
See also discussion and @kacf comment here: https://github.com/mendersoftware/meta-mender/pull/1236#issuecomment-748084791

oleorhagen (approved):

meta-mender-core/recipes-mender/mender-client/mender-client.inc
```diff
@@ -268,7 +268,7 @@ do_install() {
 
     # install dbus files
     if grep -q install-dbus ${B}/src/${GO_IMPORT}/Makefile; then
-        if ${@bb.utils.contains('PACKAGECONFIG', 'dbus', 'true', '', d)}; then
+        if ${@bb.utils.contains('PACKAGECONFIG', 'dbus', 'true', 'false', d)}; then
```
kacf, line 271: Aha, there it was!

kacf (approved):

PR #1248, opened 2020-12-21
MEN-4273 (dunfell): Copy self-signed certificate to ca-certificates

lluiscampos:
In addition, I ported the minumum commits to fix mender-client build in `dunfell`.

Changelog: mender-client: The self-signed Mender server certificate, if
present, is copied to ca-certificates in addition to
`MENDER_CERT_LOCATION` to be trusted by other services running in the
device.

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client.inc
@@ -93,7 +93,8 @@ INSANE_SKIP_${PN}-ptest = "ldflags textrel"
 
 GO_IMPORT = "github.com/mendersoftware/mender"
 
-PACKAGECONFIG ?= "${@mender_feature_is_enabled("mender-client-install", "inventory-network-scripts", "", d)}"
+_MENDER_PACKAGECONFIG_DEFAULT = "${@mender_feature_is_enabled("mender-client-install", "dbus inventory-network-scripts", "", d)}"
+PACKAGECONFIG ?= "${_MENDER_PACKAGECONFIG_DEFAULT}"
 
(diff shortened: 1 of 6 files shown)


kacf (approved):

lluiscampos:
@kacf The last force-push introduces:
* Commit 1516c35. Otherwise `test_build_nodbus` fails with "cannot find -lgio-2.0" while running ptest (unit test have no support for removing dbus)
  * This also means that a user disabling this in his/her `conf/local.conf` will also run into this problem (as I did locally while debugging). We haven't seen it because ptest are disabled on CI. Should we follow independently?
* An **important** typo fix in conflict resolution at `mender-client.inc`: in `dunfell` update modules are only installed by default in demo layer. See https://github.com/mendersoftware/meta-mender/compare/4f1f51ec74f328de655c03fe053b5bc12916f933..b52b866ad8160671255bbcc917e8ac5bdf42b4a2#diff-b1b4a6c1b87e731a379af60c4413916faaecada2293291a27262ac7f79310395R100

kacf:
> Commit [1516c35](https://github.com/mendersoftware/meta-mender/commit/1516c356f831071b8708da5447c3793b79213fc1). Otherwise `test_build_nodbus` fails with "cannot find -lgio-2.0" while running ptest (unit test have no support for removing dbus)
> 
>     * This also means that a user disabling this in his/her `conf/local.conf` will also run into this problem (as I did locally while debugging). We haven't seen it because ptest are disabled on CI. Should we follow independently?

Right, I didn't consider that. Yes, please follow independently. The problem with disabling dbus is not a new problem anyway, so it should not block this PR.

PR #1251, opened 2020-12-28
Dependabot updates + Misc clean-ups

lluiscampos:
#1240, #1241, #1242, #1244 and:

* [tests/acceptance] Remove pytest-xdist from requirements
(makes #1243 N.A.)
* [mender-client.inc] Remove nonexisting file from FILES
* [meta-mender-ci] Remove override of PACKAGE_CLASSES

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client.inc
@@ -44,7 +44,6 @@ FILES_${PN} += "\
     ${datadir}/mender/modules/v3/rpm \
     ${datadir}/mender/modules/v3/script \
     ${datadir}/mender/modules/v3/single-file \
-    ${sysconfdir}/mender.conf \
     ${sysconfdir}/udev/mount.blacklist.d/mender \
     ${systemd_unitdir}/system/${MENDER_CLIENT}.service \
     /data/mender/device_type \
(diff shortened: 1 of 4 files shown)


meta-mender-core/recipes-mender/mender-client/mender-client.inc
```diff
@@ -44,7 +44,6 @@ FILES_${PN} += "\
     ${datadir}/mender/modules/v3/rpm \
     ${datadir}/mender/modules/v3/script \
     ${datadir}/mender/modules/v3/single-file \
-    ${sysconfdir}/mender.conf \
```
oleorhagen, line 47: Question:
Why not fix the path?

oleorhagen (approved):

meta-mender-core/recipes-mender/mender-client/mender-client.inc
lluiscampos: If it is not needed, why explicitly add it to `FILES`?

oleorhagen: Hmm, odd. I thought this was part of the QA check, and that we always install the `/etc/mender/mender.conf` (local) conf file. I was expecting this to fail if :thinking:

lluiscampos: You are right in that if a file is installed and nos listed in `FILES`, bitbake QA checks will fail.

But some paths are added to `FILES` by inherited classes and I suspect that the full `/etc/mender` dir is added this way.

PR #1258, opened 2021-01-05
[test_delta_update_module] Update skip reason in test

lluiscampos:


tests/acceptance/commercial/test_delta_update_module.py
```diff
@@ -173,8 +173,9 @@ def test_perform_update(
         """Perform a delta update.
 
         """
-        # TODO: MEN-4296: Enable after next mender-binary-delta release
-        pytest.skip("Needs mender-binary-delta release 1.1.1 or 1.2.0. See MEN-4296")
+        pytest.skip(
+            "Incompatible with vexpress-qemu, needs ARM with FPU support. See MEN-4297"
+        )
```
kacf, line 178: But it works on x86, doesn't it? That way we at least get some test coverage.

lluiscampos: It works, that is right and I considered originally. Using `not_for_machine` to skip it for ARM platforms only gives the wrong impression that the test is not applicable for these, which is not true.

I can do the `not_for_machine` approach, if you insist, but not before `mender-binary-delta` release. If so, I will then close this PR, leave it as is, and modify during release.

kacf: > It works, that is right and I considered originally. Using `not_for_machine` to skip it for ARM platforms only gives the wrong impression that the test is not applicable for these, which is not true.

Nothing a comment can't handle! :-)

> I can do the `not_for_machine` approach, if you insist, but not before `mender-binary-delta` release. If so, I will then close this PR, leave it as is, and modify during release.

:+1:

PR #1260, opened 2021-01-05
MEN-4294 [test_dbus] Uncomment second part of the test

lluiscampos:
To check if the race condition is still present in the software.


lluiscampos:
Ups. Already done at #1259

PR #1261, opened 2021-01-06
Dunfell update

kacf:


tests/acceptance/test_build.py
```diff
@@ -1091,6 +1096,43 @@ def test_extra_parts(
         assert re.search(test3_re, fstab, flags=re.MULTILINE) is not None
         assert re.search(test4_re, fstab, flags=re.MULTILINE) is not None
 
+    @pytest.mark.min_mender_version("2.5.0")
+    def test_build_nodbus(self, request, prepared_test_build, bitbake_path):
```
lluiscampos, line 1100: Thanks for swapping these. I think are the same tests that I had problems in zeus/warrior some time ago. I guess next time I update they'll get in sync again :)

lluiscampos (approved):
I browsed the commits. The list looks good :+1: And from my recent misc fixes I can't think about anything missing.

tests/acceptance/test_build.py
kacf: Yes, I figured it was better to keep it as cherry-pick-friendly as possible, even if it means an extra commit for dunfell which isn't in master.

Thanks for going through it!

PR #1265, opened 2021-01-11
New Mender binary delta releases

kacf:


kacf:
Wait a minute. I should remove all of the recipes that are incompatible with libubootenv, both here and on dunfell.

kacf:
There, fixed!

lluiscampos (approved):
I also looked at the pushed tags, the pipelines and the S3 buckets contents. All looks good :+1:

kacf:
Thanks!

PR #1274, opened 2021-01-19
Make mender-connect dependent on mender-client.

kacf:
Also make sure the demo layer only installs it if we are building a
recent enough client. `layer.conf` seems to have some restrictions on
inline Python, so move it to a recipe as a dependency instead.

Changelog: None


lluiscampos (approved):

lluiscampos:
There is an error in the build that you'll need to fix, but the design looks good and hence my approval.

lluiscampos (approved):

PR #1278, opened 2021-01-21
dunfell: Add recipes for Mender 2.4.2 and 2.5.1.

kacf:


lluiscampos (approved):

lluiscampos (approved):
The cert fix looks good to me :+1: 

A safer approach in 58a7412 could have been to leave `ServerCert` setting in `mender.conf`. That would have prevented this error.

kacf:
I believe a similar fix for the demo certificate is not needed in master, since a separate recipe is used there. But I launched [this job](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/245392867), building a production version, just to verify.

> A safer approach in [58a7412](https://github.com/mendersoftware/meta-mender/commit/58a741240c176c01b5b35d02249d023b9eb4352e) could have been to leave `ServerCert` setting in `mender.conf`. That would have prevented this error.

True. But we of course thought we had our bases covered when the tests passed. I'm not even sure what we could have done to reveal this kind of error, and there is a danger something like it could happen again, due to the duplicated `do_install`. If you have any suggestions...

lluiscampos:
> I believe a similar fix for the demo certificate is not needed in master, since a separate recipe is used there. But I launched [this job](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/245392867), building a production version, just to verify.
> 

Thanks :+1: 

> > A safer approach in [58a7412](https://github.com/mendersoftware/meta-mender/commit/58a741240c176c01b5b35d02249d023b9eb4352e) could have been to leave `ServerCert` setting in `mender.conf`. That would have prevented this error.
> 
> True. But we of course thought we had our bases covered when the tests passed. I'm not even sure what we could have done to reveal this kind of error, and there is a danger something like it could happen again, due to the duplicated `do_install`. If you have any suggestions...

Not really. Something similar happened to me on zeus/warrior with the overriden do_install targets. I'll keep thinking in the back of my head...

kacf:
> I launched [this job](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/245425279), building a production version, just to verify.

Success.

PR #1290, opened 2021-02-05
MEN-4420: Add mender-configure git recipe

lluiscampos:

PR #1294, opened 2021-02-08
mender-configure git: Use always AUTOREV

lluiscampos:
The mender_configure_autorev_if_git_version logic is only applicable
once we have other recipes to chose from. As it stands (only git recipe)
we should always build latest.


lluiscampos:
Follow-up (amend, really) of #1290

kacf (approved):

PR #1295, opened 2021-02-08
[requirements_py3.txt] Add cryptography

lluiscampos:
And set it to the previous to last version, not introducing Rust
dependency. We'll update it in a different PR.


lluiscampos:
Merged with no review: it blocks pipelines in `mender-convert` and `mender-test-containers`.

kacf:
A quite safe review-less merge though! :wink:

PR #1312, opened 2021-02-26
Make device matching independent of device prefix.

kacf:
All devices end in digits, and always have non-digit preceding those
digits, so use that to match instead of individual names.

Changelog: None

### DIFF
--- a/meta-mender-core/classes/mender-helpers.bbclass
+++ b/meta-mender-core/classes/mender-helpers.bbclass
@@ -33,20 +33,11 @@ get_uboot_device_from_device() {
 get_part_number_from_device() {
     dev_base="unknown"
     case "$1" in
-        /dev/mmcblk*p* )
-            dev_base=$(echo $1 | cut -dk -f2 | cut -dp -f2)
-            ;;
-        /dev/[sh]d[a-z][1-9])
-            dev_base=${1##*d[a-z]}
-            ;;
-        /dev/nvme[0-9]n[0-9]p[0-9])
-            dev_base=$(echo $1 | cut -dp -f2)
-            ;;
-        ubi*_* )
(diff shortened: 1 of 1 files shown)


kacf:

kacf:

meta-mender-core/classes/mender-helpers.bbclass
```diff
@@ -33,21 +33,12 @@ get_uboot_device_from_device() {
...
-        /dev/nvme[0-9]n[0-9]p[0-9])
-            dev_base=$(echo $1 | cut -dp -f2)
-            ;;
-        ubi*_* )
-            dev_base=$(echo $1 | cut -d_ -f2)
-            ;;
         /dev/disk/by-partuuid/* )
             bberror "Please enable mender-partuuid Distro feature to use PARTUUID"
             ;;
+        /dev/*)
```
lluiscampos, line 39: This won't match the `ubi` case.

lluiscampos (commented):
Except for the `ubi` case (probably only needs a clarification) the change looks sane to me. But what is the motivation?

meta-mender-core/classes/mender-helpers.bbclass
kacf: Darn, you're right, I was too quick there. I guess I can just match `*` to catch everything.

kacf:
> But what is the motivation?

The motivation is to be able to set it to `/dev/devicevg/rootfs1` (or other Volume Group based names), as described in [this blog post](https://mender.io/blog/using-lvm-to-reduce-dual-rootfs-footprint). I will present this at the review today.

kacf:
Updated!

@mender-test-bot: start pipeline

lluiscampos (approved):

PR #1313, opened 2021-03-01
Dunfell update

kacf:


lluiscampos (approved):
Looks good :+1: 

As a tip: when I update `zeus`, I usually just cherry-pick dependabot commit (from #1308, in your case) instead of manually updating the `mender-image-tests` submodule.

kacf:
> As a tip: when I update `zeus`, I usually just cherry-pick dependabot commit (from #1308, in your case) instead of manually updating the `mender-image-tests` submodule.

Good tip. I will have to drop it now anyway, as it introduces a dependency on mender-configure, which we can not introduce in dunfell yet. Will just grab it later instead.

kacf:
Finally passed with integration/2.6.x. The status checks are in shambles because of all the attempts though...

PR #1315, opened 2021-03-03
tests: Add missing meta-oe layer to Beaglebone builds.

kacf:
Changelog: None


kacf:
Merging this without review, since it is a very obvious fix, and blocking the pipeline.

PR #1321, opened 2021-03-09
Fix `EFI_PROVIDER` being set incorrectly when using certain layers.

kacf:
Use stronger `EFI_PROVIDER` assignment if related classes are set.
Since a default has already been set in some layers, and `?=`
assignments work in a first-one-counts fashion, our values do not
propogate if we use `?=`.

The original motivation and discussion for this fix is here:
https://hub.mender.io/t/simple-configuration-dunfell/2991/28

Changelog: Title

### DIFF
--- a/meta-mender-core/classes/mender-setup-grub.inc
+++ b/meta-mender-core/classes/mender-setup-grub.inc
@@ -11,10 +11,9 @@ PREFERRED_RPROVIDER_virtual/grub-bootconf ?= "grub-mender-grubenv"
 
 # Set EFI_PROVIDER.  Not all MACHINE configs use it but notably
 # intel-corei7-64 does and without this we use the default of systemd-boot.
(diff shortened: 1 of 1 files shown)


lluiscampos (approved):
Okey

> Interesting, so the priority of ?= is reverse, it’s the first one that stands, instead of the last one. This almost seems like a bug in Bitbake, it breaks layer priority.

Actually, I remember this from the book I read over Christmas because I thought it was counter-intuitive. The reasoning is that this assignment is to get a "default value", so the first one stands until a `=` assignment arrives.

drewmoseley (approved):
Looks reasonable to me and the user that reported the issue indicated it fixed for them so that's good.

kacf:
One spurious error, and one known error because the mender-connect protocol has changed, not related to this PR. Merging.

PR #1331, opened 2021-03-31
MEN-4594: Prepopulate the device config in mender-configure-demo.

kacf:
Changelog: Title

### DIFF
--- a/meta-mender-core/recipes-mender/mender-configure/mender-configure.inc
+++ b/meta-mender-core/recipes-mender/mender-configure/mender-configure.inc
@@ -24,6 +24,7 @@ SYSTEMD_SERVICE_${PN} = "mender-configure-apply-device-config.service"
 
 FILES_${PN}-demo = " \
     ${libdir}/mender-configure/apply-device-config.d/mender-demo-raspberrypi-led \
+    /data/mender-configure/device-config.json \
 "
 
 FILES_${PN}-scripts = " \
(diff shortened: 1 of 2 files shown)


meta-mender-raspberrypi/recipes-mender/mender-configure/mender-configure_%.bbappend
```diff
@@ -0,0 +1,4 @@
+do_install_append_rpi() {
+    # Insert key.
+    sed -i -e '1s/$/\n  "mender-demo-raspberrypi-led": "mmc0",/' ${D}/data/mender-configure/device-config.json
```
tranchitella, line 3: Smart

meta-mender-core/recipes-mender/mender-configure/mender-configure.inc
```diff
@@ -44,6 +45,15 @@ do_install() {
     install -d -m 755 ${D}/data/mender-configure
     install -d -m 755 ${D}/var/lib/
     ln -s /data/mender-configure ${D}/var/lib/mender-configure
+
+    # For demo
+    cat > ${D}/data/mender-configure/device-config.json <<EOF
+{
+  "timezone": "UTC"
```
tranchitella, line 52: Is UTC hardcoded? For my learning, isn't it something you can customize in Yocto?

tranchitella (approved):

meta-mender-core/recipes-mender/mender-configure/mender-configure.inc
kacf: I also thought so, but all I could find was [this variable](https://www.yoctoproject.org/docs/latest/mega-manual/mega-manual.html#var-INSTALL_TIMEZONE_FILE), which determines whether you want a timezone file installed or not. But it doesn't say anything about the content. I assume that what they want users to do is to set that variable to 0, and then install a file themselves. Unfortunately, since this can be done in any recipe, it's hard for us to pick it up at build time. If this was a production feature, I would have dug deeper into this, but I can't see a smarter solution by just scratching the surface.

PR #1339, opened 2021-04-14
Demo images

TheMeaningfulEngineer:

### Description

Want to start a conversation about this, so the best way to communicate is usually by having the code present.


oleorhagen:
Makes sense to me, some thoughts while perusing this:

* but why do you explicitly need the `inventory` script to persist on the data partition, instead of having it as a part of every image build? :thinking: 
* Should the `rw-image` recipe be a part of `meta-mender-commercial-demo` if we go down this route? I guess this could have been part of the `demo layer` we already have?

meta-mender-commercial-demo/recipes-extended/images/mender-demo-delta.bb
```diff
@@ -0,0 +1,6 @@
+require recipes-extended/images/mender-demo-minimal-rw.bb
+
+IMAGE_INSTALL_append = " mender-binary-delta"
+
+IMAGE_FEATURES_append = " read-only-rootfs"
+EXTRA_IMAGECMD_ext4_append = "${@bb.utils.contains('IMAGE_FEATURES', 'read-only-rootfs', ' -O ^64bit -O ^has_journal', '', d)}"
```
kacf, line 6: I'm not quite clear on what this layer adds compared to the original one. What is the intended use for this layer?

meta-mender-demo/recipes-mender/example-custom-inventory/files/mender-inventory-demo
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+KEY_NAME=Location
+KEY_VALUE=Europe
+
+echo "$KEY_NAME=$KEY_VALUE"
```
kacf, line 6: In general I think that a better design is to only store data on the data partition, and have the script (still on the rootfs) fetch data from there. This gives the best of both worlds: The data survives, but the script can be updated.

meta-mender-demo/recipes-extended/images/mender-demo-minimal-rw.bb
```diff
@@ -0,0 +1,4 @@
+inherit core-image
```
kacf, line 1: Should this be `core-image-minimal`?

meta-mender-demo/recipes-extended/images/mender-demo-minimal-rw.bb
```diff
@@ -0,0 +1,4 @@
...
+
+IMAGE_FEATURES += "ssh-server-openssh allow-empty-password debug-tweaks"
+IMAGE_INSTALL_append = " example-custom-inventory"
```
kacf, line 4: I support the idea of having such an image, but I'm a bit worried about discoverability and maintenance. Should we use this as the build target in our docs?

Also, since you're adding `example-custom-inventory`, you might consider adding `example-state-scripts` as well.

kacf (commented):
> `echo 'LICENSE_FLAGS_WHITELIST_append = " commercial_mender-binary-delta"' >> conf/local.conf`

For legal reasons, it is not possible to automate this. It must be a conscious action by the user or it's not legally binding. Well, technically it is signing our agreement which is legally binding, but it means we cannot automate the part which blocks the build until they have agreed to it.

> `echo "FILESEXTRAPATHS_prepend_pn-mender-binary-delta := \"${HOME}/mender-binary-delta-1.2.0/:\"" >> conf/local.conf`

Here I welcome suggestions! The tricky part is that it has to be manually downloaded first...

lluiscampos (commented):
Nothing to add that @kacf has not pointed out already.

If this PR was to go as-is, it requires changelogs at least one for the new `meta-mender-commercial-demo` and one for the `mender-demo-minimal` image.

meta-mender-commercial-demo/recipes-extended/images/mender-demo-delta.bb
TheMeaningfulEngineer: You're right. It provides almost nothing. Will rethink this.

meta-mender-demo/recipes-mender/example-custom-inventory/files/mender-inventory-demo
TheMeaningfulEngineer: I'm hoping to have it fully customizable for cases where I want to quickly add 5 new inventories to an existing RO image.
But I need this only for the demo images I build to give demos, it doesn't need to end up in the reference demo image.
Will make that separation in the images.

meta-mender-demo/recipes-extended/images/mender-demo-minimal-rw.bb
TheMeaningfulEngineer: > Should we use this as the build target in our docs?

Not yet, this is just a simplifies and aims to standardize generation of demo images.
I wasn't aiming for it to be used beyond that for now, but do hope to see it go in this direction eventually.

TheMeaningfulEngineer: I've also removed the `example-custom-inventory` from the image as that will never make sense beyond anything but giving demos.

meta-mender-demo/recipes-extended/images/mender-demo-minimal-rw.bb
TheMeaningfulEngineer: Changed

meta-mender-demo/recipes-mender/example-custom-inventory/files/mender-inventory-demo
TheMeaningfulEngineer: Removed it from the image, just exists as a helper recipe in the demo layer.

TheMeaningfulEngineer: Removed it completely to simplify the merge

kacf (approved):
Looks good to me!

kacf:
@TheMeaningfulEngineer: Btw, I noticed now that this is dunfell. Should it be submitted to master first?

PR #1352, opened 2021-05-03
MEN-4671: New recipe mender-modules-gen and update module tests

lluiscampos:


lluiscampos:
I still need to polish few things on the recipe, but I am launching the PR already to get feedback on the general direction. Is this a bad idea?

lluiscampos:
The tests are at https://github.com/mendersoftware/mender-image-tests/pull/46

lluiscampos:

lluiscampos:
Testing together with fixes from https://github.com/mendersoftware/mender/pull/750 @ [Pipeline 296214432](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/296214432)

meta-mender-demo/recipes-mender/mender-modules-gen/mender-modules-gen.bb
```diff
@@ -0,0 +1,74 @@
...
+        return d.getVar("AUTOREV")
+    else:
+        return "f6ffa190892202263fdb75975059fbb201adab6a"
+
+SRCREV = '${@mender_autorev_if_git_version(d)}'
+
+# TODO: take into account build tags and other "-git" expressions
+def mender_modules_gen_version_from_mender_version(d, srcpv):
+    import re
+    pref_version = d.getVar("PREFERRED_VERSION_pn-mender-client")
```
kacf, line 25: @lluiscampos: Ref what we talked about earlier: Here it appears to use `pn-`, but I thought without `-pn` was the correct approach. In the end I think this might stem from having initially misunderstood it, but `pn-` also happened to work, so this is what we have used in our builds. The whole issue might be a "poteito, potato" discussion at this point.

meta-mender-demo/recipes-mender/mender-modules-gen/mender-modules-gen.bb
```diff
@@ -0,0 +1,74 @@
...
+        # Else return master as branch.
+        return "master"
+
+MENDER_BRANCH = "${@mender_branch_from_mender_preferred_version(d)}"
+SRC_URI = "git://github.com/mendersoftware/mender;protocol=https;branch=${MENDER_BRANCH}"
+
+FILES_${PN} = " \
+    ${bindir}/directory-artifact-gen \
+    ${bindir}/single-file-artifact-gen \
+    ${bindir}/docker-artifact-gen \
```
kacf, line 56: Bitbake supports turning the output of one recipe into several packages (`PACKAGES` variable). While having these files in their own recipe may be less complex overall, there may be reasons for incorporating this into the mender-client recipe and making a separate package out of it. I'm not sure if the reasons are strong enough though. Here are my pros and cons:

Pros:
* With a separate recipe we need to create versioned recipes for yet another component. It's considered a bad thing in Yocto to lock something to AUTOREV, it gives non-deterministic builds. If we reuse the mender-client recipe, we get this for free.
* If we introduce something incompatible in the generators, you won't see any of that if you stay on an old client, because the versions automatically go hand in hand.
Cons:
* It might be more finicky to get the native part to work correctly together with the non-native part. But you should be able to repeat most of what you did here by using:
  ```
  do_configure_class-native() {
      true
  }
  ```
* You cannot use a different version of the client and the generators.

What do you think?

lluiscampos: Thanks for your input.

I don't love my current approach, it is a very bad practice from one recipe access `PREFERRED_VERSION` of another recipe...

That was my first approach, indeed, to add a package to `mender-client` recipe. The problem I run into is that all the dependencies of mender-client do not provide native versions. I get these kind of errors:
```
ERROR: Nothing RPROVIDES 'liblzma-native' (but virtual:native:/home/lluis/poky/meta-mender/meta-mender-core/recipes-mender/mender-client/mender-client_2.6.0.bb RDEPENDS on or otherwise requires it)
```

```
ERROR: Nothing RPROVIDES 'mender-server-certificate-native' (but virtual:native:/home/lluis/poky/meta-mender/meta-mender-core/recipes-mender/mender-client/mender-client_2.6.0.bb RDEPENDS on or otherwise requires it)
```

(probably others too - the build just aborts on the first missing dependency so I cannot get the full list).

This proved to be too hard to workaround for me, do you have advise in this direction? We should find a way to remove packages from `PACKAGES` (namely, the main mender-client package) from the native flavor; then split the `RDEPENDS` per package (i.e. mender-client depends on lzma and more, mender-client-modules-gen depends only on bash).

lluiscampos: @kacf I came up with an alternative implementation closer to your suggested approach, see #1353. It has a different set of issues but let's discuss it there.

lluiscampos:
Closing in favor of #1353

PR #1353, opened 2021-05-04
MEN-4671: mender-client: New package mender-modules-gen and Update Module tests 

lluiscampos:
This is an alternative approach for #1352 . See specifically discussion at https://github.com/mendersoftware/meta-mender/pull/1352#discussion_r625143882

PR #1356, opened 2021-05-05
[test_preferred_versions] Test mender-client-native recipe

lluiscampos:


lluiscampos:
This is a follow-up from #1353

@kacf is this the right thing to do? I first thought I would just *add* the recipe, but I noticed that `mender-artifact` is not duplicated (target and native recipes) and I _think_ the code tests both when the "recipe" have the `-native` suffix.

lluiscampos:
@mender-test-bot start pipeline :athletic_shoe:

lluiscampos:
@mender-test-bot start pipeline :mans_shoe:

kacf (changes requested):
Hmm, I'm pretty sure you need to add. I think the confusion comes from the matching of the `-native` suffix, but this is only used to set the variables correctly. You can see it by inspecting the final bitbake statement: This uses only `recipe`, which is the one from the parameters.

lluiscampos:
> Hmm, I'm pretty sure you need to add. I think the confusion comes from the matching of the `-native` suffix, but this is only used to set the variables correctly. You can see it by inspecting the final bitbake statement: This uses only `recipe`, which is the one from the parameters.

Right. I see now the final `bibake <recipe>` statement.

This means that we are not testing `mender-artifact` (non-native), so I will add that to this PR.

lluiscampos:

kacf:
> This means that we are not testing `mender-artifact` (non-native), so I will add that to this PR.

No, but we never intended to either. There can be some uses for having it on the device though, so by all means, add it if it works.

kacf (approved):
It's actually a bit confusing to read because of Black's decision not to break up the first entry. I guess it's only a few characters' difference. But not much we can do about it :-)

lluiscampos:
> It's actually a bit confusing to read because of Black's decision not to break up the first entry. I guess it's only a few characters' difference. But not much we can do about it :-)

I know! Annoying...

> so by all means, add it if it works.

I didn't really work. Seems like `arm-poky-linux-gnueabi-go install` install the binary in `.../build/bin/linux_arm/` instead of `.../build/bin/`. I'll drop the commit and (maybe) follow-up independently on a value add session.

lluiscampos:

PR #1367, opened 2021-05-22
MEN-4106: add rpm to the deb requirements to run acceptance tests

tranchitella:
Changelog: none

Signed-off-by: Fabio Tranchitella <fabio.tranchitella@northern.tech>

### DIFF
--- a/tests/acceptance/requirements_deb.txt
+++ b/tests/acceptance/requirements_deb.txt
@@ -47,6 +47,7 @@ qemu-kvm
 qemu-system-arm
 qemu-system-x86
 qemu-utils
+rpm
 s3cmd
 screen
 sudo


tranchitella:
> @tranchitella, Let me know if you want to start the integration pipeline by mentioning me and the command "start pipeline".

do that, @mender-test-bot

merlin-northern (approved):

lluiscampos:
I _think_ that the right thing to do here would have been adding `rpm-native` to [`mender-test-dependencies` recipe](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/recipes-testing/mender-test-dependencies/mender-test-dependencies_1.0.bb), instead of using `rpm` from the host OS.

PR #1369, opened 2021-05-27
Dunfell update

kacf:


kacf:
@lluiscampos: I'm a little bit unsure about "mender-client: New package mender-modules-gen", is it correct to bring it into dunfell? It seems to be required for tests, and should only affect the build host.

Also, there were some rather non-trivial conflicts, so would be nice if you could compare it to the original patch, and make sure it still makes sense.

lluiscampos:
> @lluiscampos: I'm a little bit unsure about "mender-client: New package mender-modules-gen", is it correct to bring it into dunfell? It seems to be required for tests, and should only affect the build host.
> 
> Also, there were some rather non-trivial conflicts, so would be nice if you could compare it to the original patch, and make sure it still makes sense.

@kacf We discussed this (I couldn't find the discussion on the original PR, so I guess was over Slack) and concluded in not bringing it. While the package should be harmless, it is needed only for new tests. Furthermore, see that in [the mender-image-tests PR](https://github.com/mendersoftware/mender-image-tests/pull/46/files#diff-efbcfc930efb6910245071e0aa1feb35e7d3d1cd66b1c3447fd66a526275272eR27) I did set the tests to not run for `dunfell` and earlier.

kacf:
Ok thanks for confirming, I will remove that commit then.

lluiscampos (approved):

PR #1379, opened 2021-06-01
MEN-4550: Add acceptance test for DBus Update Control endpoint.

kacf:
`mender-mock-server.py` has to be substantially extended for the calls
to work correctly. We have to consume the body of requests for
subsequent calls to work, and we need to be able to handle chunked
bodies. We also make it possible to schedule a deployment with it.

In addition we add a logger-update-module which we can use to track
exactly which states the Update Module is going through.

Changelog: None

### DIFF
--- a/tests/acceptance/mock_server.py
+++ b/tests/acceptance/mock_server.py
@@ -1,0 +1,122 @@
+#!/usr/bin/python
+# Copyright 2021 Northern.tech AS
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+import os
+import pytest
+import tempfile
+
+from utils.common import put_no_sftp
+
+# May appear excessive, but emulated QEMU is *really* slow to execute this.
+EXPIRATION_TIME = 60
+BOOT_EXPIRATION_TIME = 30
+
+MENDER_CONF = """{
+    "UpdatePollIntervalSeconds": 1,
+    "InventoryPollIntervalSeconds": 5,
+    "RetryPollIntervalSeconds": 1,
+    "UpdateControlMapExpirationTimeSeconds": %d,
+    "UpdateControlMapBootExpirationTimeSeconds": %d,
+    "ServerURL": "https://docker.mender.io:8443",
+    "DBus": {
+        "Enabled": true
+    }
+}
+""" % (
+    EXPIRATION_TIME,
+    BOOT_EXPIRATION_TIME,
+)
+
+
+@pytest.fixture(scope="function")
+def setup_mock_server(request, bitbake_variables, connection):
+    if bitbake_variables["MACHINE"] == "vexpress-qemu-flash":
+        pytest.skip(
+            "Mock server is not available on vexpress-qemu due to space constraints."
+        )
+
+    conffile = "/data/etc/mender/mender.conf"
+    conffile_bkp = f"{conffile}.backup"
+    bdir = os.path.dirname(conffile_bkp)
+    result = connection.run(
+        f"mkdir -p {bdir} && if [ -e {conffile} ]; then cp {conffile} {conffile_bkp}; fi"
+    )
+
+    tf = tempfile.NamedTemporaryFile()
+    with open(tf.name, "w") as fd:
+        fd.write(MENDER_CONF)
+
+    put_no_sftp(tf.name, connection, remote=conffile)
+
+    hostsfile = "/data/etc/hosts"
+    hostsfile_bkp = f"{hostsfile}.backup"
+    connection.run(
+        f"cp {hostsfile} {hostsfile_bkp} && echo 127.0.0.1 docker.mender.io >> {hostsfile}"
+    )
+
+    def fin():
+        connection.run(
+            f"if [ -e {conffile_bkp} ]; then dd if={conffile_bkp} of=$(realpath {conffile}); fi"
+        )
+        connection.run(
+            f"if [ -e {hostsfile_bkp} ]; then dd if={hostsfile_bkp} of=$(realpath {hostsfile}); fi"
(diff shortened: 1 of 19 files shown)


tests/acceptance/mock_server.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/python
+# Copyright 2020 Northern.tech AS
```
lluiscampos, line 2: 2021.

I guess we don't have these checks on meta-mender... :memo:

tests/acceptance/mock_server.py
```diff
@@ -0,0 +1,117 @@
...
+        "Enabled": true
+    }
+}
+""" % (
+    EXPIRATION_TIME,
+    BOOT_EXPIRATION_TIME,
+)
+
+
+@pytest.fixture
```
lluiscampos, line 42: Better to be explicit, I think
```suggestion
@pytest.fixture(scope="function")
```

tests/acceptance/mock_server.py
```diff
@@ -0,0 +1,117 @@
...
+def setup_mock_server(request, bitbake_variables, connection):
+    conffile = "/data/etc/mender/mender.conf"
+    conffile_bkp = f"{conffile}.backup"
+    bdir = os.path.dirname(conffile_bkp)
+    result = connection.run(
+        f"mkdir -p {bdir} && if [ -e {conffile} ]; then cp {conffile} {conffile_bkp}; fi"
+    )
+    assert result.exited == 0
```
lluiscampos, line 50: (Just because I have seen this often in modern tests)

In client acceptance tests, a `connection.run` command not exiting with `0` will raise an `Exception` and finish the test, so asserting here is not really adding value. If you want to run a command for which we can expect a non-0 return, then we add the `warn=True` parameter.

tests/meta-mender-ci/recipes-testing/logger-update-module/logger-update-module.bb
```diff
@@ -0,0 +1,30 @@
+LIC_FILES_CHKSUM = "file://${COREBASE}/meta/files/common-licenses/Apache-2.0;md5=89aea4e17d99a7cacdbeed46a0096b10"
```
lluiscampos, line 1: This is equivalent to:
```suggestion
LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/Apache-2.0;md5=89aea4e17d99a7cacdbeed46a0096b10"
```

Which we are already using in other recipes

tests/meta-mender-ci/recipes-testing/logger-update-module/logger-update-module.bb
```diff
@@ -0,0 +1,30 @@
...
+        echo "$(date +%s)" "$@" >> /data/logger-update-module.log
+        ;;
+esac
+EOF
+}
+
+do_install() {
+    install -m 755 -d ${D}${datadir}/mender/modules/v3
+    install -m 755 ${B}/logger-update-module ${D}${datadir}/mender/modules/v3/logger-update-module
+}
```
lluiscampos, line 30: Nice! :+1:

tests/acceptance/test_update_control.py
```diff
@@ -0,0 +1,431 @@
...
+    ]
+
+    @pytest.mark.min_mender_version("2.7.0")
+    @pytest.mark.parametrize(
+        "case_name,case",
+        [(case["name"], case["case"]) for case in test_update_control_maps_cases],
+    )
+    def test_update_control_maps(
+        self,
+        case_name,
```
lluiscampos, line 294: `case_name` is unused

tests/acceptance/test_update_control.py
```diff
@@ -0,0 +1,431 @@
...
+                            "ArtifactCommit_Enter": {"action": "force_continue",},
+                        },
+                        "priority": -1,
+                    },
+                ],
+                "success": False,
+                "last_successful_state": "ArtifactVerifyReboot",
+            },
+        },
+    ]
```
lluiscampos, line 285: Excellent tests. I was able to follow all the cases with no much trouble.

lluiscampos (approved):
I really couldn't find find much, see couple of minor comments below.

The test code is very good, the server mock code a tiny bit cumbersome, but I didn't had better suggestions neither.

Well done! Both with the tests and with the bugs that we have revealed (and fixed!).

tests/acceptance/test_update_control.py
kacf: It is required by `parametrize`, otherwise it complains about a missing parameter. The ultimate goal is that the test case be named instead of just an index, which is much harder to identify in the list of cases.

tests/acceptance/test_update_control.py
kacf: That's great! I was afraid the many labels would be a bit too much. At one point did I wonder if it should be split, but there is a lot of reuse.

tests/acceptance/mock_server.py
kacf: Right. This is preexisting, but I can remove it.

kacf:
> the server mock code a tiny bit cumbersome, but I didn't had better suggestions neither.

Agree on both counts!

Thanks for the review, I have fixed all the issues pointed out, as well as (hopefully) the test failures. All were minor things, so I squashed.

tests/acceptance/test_update_control.py
```diff
@@ -0,0 +1,432 @@
...
+            now = time.time()
+
+            make_and_deploy_artifact()
+
+            log = []
+            pause_state_observed = 0
+            continue_map_inserted = False
+            second_deployment_done = False
+            PAUSE_STATE_OBSERVE_COUNT = 2
+            while time.time() - now <= case.get("fail_after", FAIL_TIME):
```
oleorhagen, line 356: Clever

oleorhagen (approved):
Nice!

I wasn't really able to get into the matter too deep, but on the surface looks real nice

tests/acceptance/mock_server.py
```diff
@@ -0,0 +1,116 @@
...
+    "RetryPollIntervalSeconds": 1,
+    "UpdateControlMapExpirationTimeSeconds": %d,
+    "UpdateControlMapBootExpirationTimeSeconds": %d,
+    "ServerURL": "https://docker.mender.io:8443",
+    "DBus": {
+        "Enabled": true
+    }
+}
+""" % (
+    EXPIRATION_TIME,
```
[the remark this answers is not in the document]
kacf: That's what I started with, but it doesn't work because it gets confused by the `{}` opening and closing braces for JSON.

PR #1381, opened 2021-06-01
Fix copyright statements and add CI license checks

lluiscampos:
Fix copyright statements and add CI license checks
    
To bring the repo to Mender standards.
    
Namely:
* Include GitLab template to check license in CI
* Move COPYING to LICENSE and update year
* Add license checksums file
* Add Copyright and License in some missing files
* Update Copyright year in some test files

### DIFF
--- a/LIC_FILES_CHKSUM.sha256
+++ b/LIC_FILES_CHKSUM.sha256
@@ -1,0 +1,7 @@
+b4acfcfa2a0ba1a8c82ec3965fbcee886cff8394ca4214e0ddac0a36beb1e05a  LICENSE
+b40930bbcf80744c86c46a12bc9da056641d722716c378f5659b9e555ef833e1  meta-mender-core/recipes-mender/mender-client-migrate-configuration/files/LICENSE
+b40930bbcf80744c86c46a12bc9da056641d722716c378f5659b9e555ef833e1  meta-mender-core/recipes-mender/mender-wait-for-timesync/files/LICENSE
+beb140be4cd64599bedc691a55b2729c9cc611a4b9d6ec44e01270105daf18a2  meta-mender-core/recipes-core/lsb-ld/files/LICENSE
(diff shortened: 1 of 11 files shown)


LIC_FILES_CHKSUM.sha256
```diff
@@ -0,0 +1,7 @@
+b4acfcfa2a0ba1a8c82ec3965fbcee886cff8394ca4214e0ddac0a36beb1e05a  LICENSE
+b40930bbcf80744c86c46a12bc9da056641d722716c378f5659b9e555ef833e1  meta-mender-core/recipes-mender/mender-client-migrate-configuration/files/LICENSE
+b40930bbcf80744c86c46a12bc9da056641d722716c378f5659b9e555ef833e1  meta-mender-core/recipes-mender/mender-wait-for-timesync/files/LICENSE
+beb140be4cd64599bedc691a55b2729c9cc611a4b9d6ec44e01270105daf18a2  meta-mender-core/recipes-core/lsb-ld/files/LICENSE
+ceb1b36ff073bd13d9806d4615b931707768ca9023805620acc32dd1cfc2f680  meta-mender-demo/recipes-mender/boot-script/files/LICENSE
+ceb1b36ff073bd13d9806d4615b931707768ca9023805620acc32dd1cfc2f680  meta-mender-demo/recipes-mender/mender-reboot-detector/files/LICENSE
+b40930bbcf80744c86c46a12bc9da056641d722716c378f5659b9e555ef833e1  meta-mender-demo/recipes-mender/example-state-scripts/files/LICENSE
```
lluiscampos, line 7: These are actually outdated and, most importantly, we don't have an automated check to update them when being modified. What should we do?

kacf: Hmm, we don't have a good mechanism for this. The only suggestion I have is to extend [the mechanism for the top level license file](https://github.com/mendersoftware/mendertesting/blob/master/check_license.sh#L11-L18) to include arbitrary other files. But we need to make sure that git log only checks the year of that folder.

OTOH, these are very small and rarely changing components, so I don't consider it very critical.

kacf (approved):

PR #1395, opened 2021-06-11
Remove old helper recipes that were used in the old HW farm.

kacf:
It is extremely unlikely that we will use these in a new HW farm, if
we build one, since these existing scripts assume we don't have any
SD-MUX devices.

Changelog: None


kacf:

lluiscampos (approved):
These have been bugging me for a long while. Thank you! :raised_hands:

oleorhagen (approved):
Nice, I didn't even know these existed!

PR #1400, opened 2021-06-22
Fixes for test_update_control_maps_cleanup

lluiscampos:
* test_update_control.py: Print mender logs on erros
    
    And add missing cleanup in test_update_control_maps_cleanup

* mender-mock-server.py: Serve /deployments/next until user cleans-up

* test_update_control.py: Fix test_update_control_maps_cleanup
    
    Now that the client will be refreshing the map from /deployments/next,
    the test needs to use different priorities for the two maps. Also, for
    the case of "continue", the test shall use "force_continue".

### DIFF
--- a/tests/acceptance/test_update_control.py
+++ b/tests/acceptance/test_update_control.py
@@ -482,14 +482,16 @@ class TestUpdateControl:
             "name": "Cleanup after success",
             "case": {
                 "pause_map": {
+                    "priority": 0,
                     "states": {"ArtifactCommit_Enter": {"action": "pause",},},
                 },
                 "pause_state": "ArtifactVerifyReboot",
                 "continue_map": {
                     "id": MUID,
+                    "priority": 10,
                     "states": {
                         "ArtifactInstall_Enter": {"action": "fail",},
-                        "ArtifactCommit_Enter": {"action": "continue",},
+                        "ArtifactCommit_Enter": {"action": "force_continue",},
                     },
                 },
                 "expect_failure": False,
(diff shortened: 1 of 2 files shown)


lluiscampos:
To be tested together with https://github.com/mendersoftware/mender/pull/789

kacf (approved):
Nice catches!

oleorhagen (approved):

lluiscampos:
@oleorhagen I assigned this one to you. Let me know if you need more help here.

tests/acceptance/test_update_control.py
```diff
@@ -579,7 +583,13 @@ def wait_for_state(state_to_wait_for):
             log = wait_for_state("Cleanup")
             assert "ArtifactFailure" not in log
 
+        except:
+            connection.run("journalctl -u mender-client | cat")
```
oleorhagen, line 592: `print(connection.run(...).stdout)` ?

tests/acceptance/test_update_control.py
```diff
@@ -450,6 +450,10 @@ def test_update_control_maps(
                     pause_state_observed >= PAUSE_STATE_OBSERVE_COUNT
                 ), "Looks like the client did not pause!"
 
+        except:
+            connection.run("journalctl -u mender-client | cat")
```
oleorhagen, line 454: `print(connection.run(...).stdout))`

tests/acceptance/test_update_control.py
oleorhagen: Makes  sense right, @lluiscampos ? I don't know if we need `stderr` too? :thinking:

lluiscampos: Fabric will already print stdout/stderr to terminal when running the command, why do we want it wrapped it in `print`? If any maybe you mean to add it to `logging` so that `Pytest` appends it on failures? (which I dislike, I prefer the failure to be about the code, and then browse up and see the test stdout)

kacf: I tried it as a fix for the "would have blocked on write" error which happens for me when the output is very long and gets truncated. But the suggested fix didn't make any difference, I suspect it has wrapped stdout anyway, so it doesn't matter if you `print` it.

lluiscampos: Okey, thanks for the background. I am also having often that error and I suspected is related to Fabric too.

Anyway, if you say is not really helping, I leave it as is until we find a proper solution.

kacf:
Superseded by #1404.

PR #1406, opened 2021-06-28
MEN-3360: Test infinite test transitions for Update Control.

kacf:
Changelog: None

### DIFF
--- a/tests/acceptance/test_update_control.py
+++ b/tests/acceptance/test_update_control.py
@@ -30,7 +30,7 @@ from mock_server import (
     EXPIRATION_TIME,
     BOOT_EXPIRATION_TIME,
 )
-from utils.common import get_no_sftp
+from utils.common import get_no_sftp, put_no_sftp
 
 # Map UIDs. Randomly chosen, but used throughout for consistency.
 MUID = "3702f9f0-b318-11eb-a7b6-c7aece07181e"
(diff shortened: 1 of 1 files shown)


kacf:
Needs to be tested together with/after https://github.com/mendersoftware/mender/pull/813.

kacf:
Running in https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/328280824.

tests/acceptance/test_update_control.py
```diff
@@ -615,3 +616,96 @@ def wait_for_state(state_to_wait_for):
             clear_update_control_maps(connection)
             connection.run("systemctl stop mender-client")
             connection.run("rm -f /data/logger-update-module.log")
+
+    @pytest.mark.min_mender_version("2.7.0")
```
lluiscampos, line 620: It does not matter much, but these could be `3.0.0` to avoid (human) confusion.

tests/acceptance/test_update_control.py
```diff
@@ -615,3 +616,96 @@ def wait_for_state(state_to_wait_for):
...
+                    "ID": MUID2,
+                    "priority": 10,
+                    "states": {"ArtifactInstall_Enter": {"action": "force_continue"}},
+                },
+            )
+
+            # Rest of deployment should finish successfully.
+            log = wait_for_state(connection, "Cleanup")
+            assert "ArtifactFailure" not in log
+
```
lluiscampos, line 704: I suggest adding there the same `except` we have for the other tests:
```
        except:
            connection.run("journalctl -u mender-client | cat")
            raise
```

lluiscampos (approved):
Fun test. Looks good!

tests/acceptance/test_update_control.py
kacf: I agree, but there are many of these, including in mender-image-tests. I'd rather not have them diverge, but fix them all simultaneously.

tests/acceptance/test_update_control.py
kacf: Fixed!

tests/acceptance/test_update_control.py
lluiscampos: Good point.

PR #1418, opened 2021-07-05
meta-mender-ci: Debug logging for mender-client when MENDER_DEVMODE set

lluiscampos:
Changelog: Title


lluiscampos:
This is something that I did last Friday while debugging the client. What do you think about merging it?

lluiscampos:
FYI @oleorhagen

kacf:
I'm very much in favor of the concept. My only concern is the use of the `MENDER_DEVMODE` variable. This is also used [with mender-binary-delta](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-commercial/recipes-mender/mender-binary-delta/mender-binary-delta.inc#L27) to build dev versions of the software. Perhaps we should keep it for that purpose and call this something else. What about `MENDER_LOGLEVEL = "debug"`, and then you can insert the value of the variable in the systemd file, if it is set:

```
sed -i 's|ExecStart=.*|ExecStart=/usr/bin/mender --log-level ${MENDER_LOGLEVEL} daemon|'
```

lluiscampos:
@stale No please, I am going to work in this on my next value add :pray:

lluiscampos:
> I'm very much in favor of the concept. My only concern is the use of the `MENDER_DEVMODE` variable. This is also used [with mender-binary-delta](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-commercial/recipes-mender/mender-binary-delta/mender-binary-delta.inc#L27) to build dev versions of the software. Perhaps we should keep it for that purpose and call this something else. What about `MENDER_LOGLEVEL = "debug"`, and then you can insert the value of the variable in the systemd file, if it is set:
> 
> ```
> sed -i 's|ExecStart=.*|ExecStart=/usr/bin/mender --log-level ${MENDER_LOGLEVEL} daemon|'
> ```

Thank you for your review and sorry that I took so long to come back here :sweat_smile: 

I decided for `MENDER_CI_LOGLEVEL` variable (to make it explicit that only affects CI layer) and added some custom handling for `mender-connect` that doesn't really support `--log-level` directly.

This is mainly for debugging acceptance tests locally, so I think it should do as-is.

lluiscampos:

lluiscampos:
ping @kacf

kacf (approved):
👍

PR #1422, opened 2021-07-07
MEN-4737 Monitoring wrapper alignment

merlin-northern:
MEN-4737 Monitoring wrapper alignment

Goes with https://github.com/mendersoftware/monitor-client/pull/17
* include service file
* add a correct link to the mender-monitord

ChangeLog:title

### DIFF
--- a/meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor.inc
+++ b/meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor.inc
@@ -1,12 +1,18 @@
 LICENSE = "CLOSED"
 LICENSE_FLAGS = "commercial"
 
+inherit systemd
+
 RDEPENDS_${PN} = "bash mender-client (>= 2.5)"
 
 FILES_${PN} = " \
     ${bindir}/mender-monitord \
     ${sysconfdir}/mender-monitor \
     ${datadir}/mender-monitor \
+"
+
+FILES_${PN}_append_mender-systemd = " \
+    ${systemd_system_unitdir}/mender-monitor.service \
 "
 
 S ?= "${WORKDIR}/${PN}"
@@ -18,9 +24,19 @@ do_version_check() {
 }
 addtask do_version_check after do_unpack before do_install
 
+SYSTEMD_SERVICE_${PN}_mender-systemd = "mender-monitor.service"
+
 do_install() {
     oe_runmake \
         -C ${S} \
         DESTDIR=${D} \
         install
 }
+
+do_install_append_mender-systemd() {
+    oe_runmake \
+        -C ${S} \
+        DESTDIR=${D} \
+        install-systemd
+}
+


meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor.inc
```diff
@@ -23,4 +24,5 @@ do_install() {
         -C ${S} \
         DESTDIR=${D} \
         install
+    ln -sf ${datadir}/mender-monitor/mender-monitord ${D}${bindir}/mender-monitord
```
lluiscampos, line 27: Still pending discussion for this in https://github.com/mendersoftware/monitor-client/pull/17#discussion_r664089103

lluiscampos (changes requested):
Nice try, but in yocto (and more specifically our `meta-mender`) things are never that simple :P

The yocto way of dealing with systemd services is:
* Inherit `systemd` class: https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/recipes-mender/mender-configure/mender-configure.inc#L4
* Set the corresponding `SYSTEMD_SERVICE` variable: https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/recipes-mender/mender-configure/mender-configure.inc#L23
* Append the file only when using `mender-systemd` feature: https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/recipes-mender/mender-configure/mender-configure.inc#L19
* Have a dedicated snstall target for `mender-systemd` feature: https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/recipes-mender/mender-configure/mender-configure.inc#L59

merlin-northern:
> Nice try, but in yocto (and more specifically our `meta-mender`) things are never that simple :P
> 
> The yocto way of dealing with systemd services is:
> 
> * Inherit `systemd` class: https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/recipes-mender/mender-configure/mender-configure.inc#L4
> * Set the corresponding `SYSTEMD_SERVICE` variable: https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/recipes-mender/mender-configure/mender-configure.inc#L23
> * Append the file only when using `mender-systemd` feature: https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/recipes-mender/mender-configure/mender-configure.inc#L19
> * Have a dedicated snstall target for `mender-systemd` feature: https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/recipes-mender/mender-configure/mender-configure.inc#L59

right! thanks. fixed. tested, works.

meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor.inc
```diff
@@ -18,9 +24,20 @@ do_version_check() {
 }
 addtask do_version_check after do_unpack before do_install
 
+SYSTEMD_SERVICE_${PN}_mender-systemd = "mender-monitor.service"
+
 do_install() {
     oe_runmake \
         -C ${S} \
         DESTDIR=${D} \
         install
```
lluiscampos, line 33: Does this `install` target also include `install-systemd`? Then we cannot use it and we need to specify all the other `install-*` targets here.

lluiscampos (changes requested):
Looking better :nerd_face:

meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor.inc
merlin-northern: I would prefer removing install-systemd from the Makefile from monitor-client -- it makes better sense, can it be like that?

lluiscampos: As you wish.

meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor.inc
lluiscampos: Discussion there is done, decided to keep the link.

However this is already done by the `install` target, so we should not need it here.

merlin-northern: discussed over gmeet. thanks!

meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor.inc
merlin-northern: aligned with https://github.com/mendersoftware/monitor-client/pull/17

lluiscampos (approved):

PR #1440, opened 2021-07-22
MEN-4811: Conditionally install D-Bus interface files

lluiscampos:
MEN-4811: Conditionally install D-Bus interface files
    
    The interface files, shipped in mender-client-dev package, shall be
    present only when PACKAGECONFIG contains "dbus".
    
    Fixed also how the test was getting `deploy_dir_rpm`. If obtained from
    bitbake variables, the dir will point to the base build dir (and not the
    prepared one for this test!).

    Changelog: Fix: mender-client-dev package contains D-Bus interface files
    only when "dbus" is enabled in PACKAGECONFIG.

In addition:
* meta-mender-qemu: Remove mender-client_git.bbappend
* Testing https://github.com/mendersoftware/mender-image-tests/pull/55 (technically unrelated, though)

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client.inc
@@ -274,12 +274,19 @@ do_install() {
     # install dbus files
     if grep -q install-dbus ${B}/src/${GO_IMPORT}/Makefile; then
         if ${@bb.utils.contains('PACKAGECONFIG', 'dbus', 'true', 'false', d)}; then
+            # install the D-Bus policy file
             oe_runmake \
                 -C ${B}/src/${GO_IMPORT} \
                 V=1 \
                 prefix=${D} \
                 datadir=${datadir} \
                 install-dbus
+
+            # install the D-Bus interface file(s)
+            if [ -f ${B}/src/${GO_IMPORT}/Documentation/io.mender.Authentication1.xml ]; then
+                install -d ${D}${datadir}/dbus-1/interface
+                install -m 0644 ${B}/src/${GO_IMPORT}/Documentation/io.mender.*.xml ${D}${datadir}/dbus-1/interface
+            fi
         fi
     fi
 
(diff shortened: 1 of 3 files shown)


lluiscampos:
@mender-test-bot start pipeline :train2:

TheMeaningfulEngineer:
@lluiscampos 
Thanks,
I'm somewhat confused on the purpose of these files on the device.

Initially I thought that `mender-connect` uses them to know the dbus interfaces of `mender-client` but that isn't the case as they communicate successfully in versions without this file.

My last impression was that it's for generating dbus api documentation for the docs page. If that is correct, do they need to live on the device at all?

lluiscampos:
> @lluiscampos
> Thanks,
> I'm somewhat confused on the purpose of these files on the device.
> 
> Initially I thought that `mender-connect` uses them to know the dbus interfaces of `mender-client` but that isn't the case as they communicate successfully in versions without this file.
> 
> My last impression was that it's for generating dbus api documentation for the docs page. If that is correct, do they need to live on the device at all?

@TheMeaningfulEngineer let me help you: this files are **not** going into the device. They are packaged separately into `mender-client-dev`, which is to be used for developing applications, for example to autogenerate boilerplate code in C or other programming language that will talk D-Bus to Mender.

The only files shipped into the device is the D-Bus policy files, which go on /usr/share/dbus-1/system.d (see [here](https://github.com/mendersoftware/meta-mender/blob/40c3453f669459f5ce81c801464a133c87c5a7d1/meta-mender-core/recipes-mender/mender-client/mender-client.inc#L32-L33)), while as I said the D-Bus interface files are packaged in `mender-client-dev` and not meant to be installed on a device (see [here](https://github.com/mendersoftware/meta-mender/blob/40c3453f669459f5ce81c801464a133c87c5a7d1/meta-mender-core/recipes-mender/mender-client/mender-client.inc#L53-L55)).

lluiscampos:
> My last impression was that it's for generating dbus api documentation for the docs page. (...)

We also use them for the API documentation, correct. Think about these files as the REST API docs, that can be read by humans to understand the API, or by robots to generate code, tests, documentation pages, etc.

TheMeaningfulEngineer (approved):

lluiscampos:
@mender-test-bot start pipeline again, this time make it green! :rainbow:

kacf (approved):

lluiscampos:
Last force-push dropped the `mender-image-tests` submodule update (to be updated independently) - not re-running the pipeline.

PR #1448, opened 2021-08-03
mender-client: do not remove the directory we have just created

merlin-northern:
It was a hack to cerate /var/lib directory.

ChangeLog:title

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client.inc
@@ -300,7 +300,7 @@ do_install() {
         install -m 0600 ${B}/persistent_mender.conf ${D}/data/mender/mender.conf
     fi
 
-    install -d ${D}/${localstatedir}/lib/mender
+    install -d ${D}/${localstatedir}/lib
 
     # install artifact verification key, if any.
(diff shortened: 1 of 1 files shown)


merlin-northern:
@mender-test-bot start pipeline please

kacf:
Oh wait, now I see. The reason it was done like this is because the removal is inside the check for `mender-image`. If `mender-image` is not set, then the creation of the full path is necessary. So in short it means:

* With `mender-image`: `/var/lib/mender` is a symlink.
* Without `mender-image`: `/var/lib/mender` is a directory.

So the correct fix is actually to do the same in mender-monitor: Create the directory in the main install block, remove it and replace with a symlink in the `mender-image` block. Sorry, I missed this when first reviewing.

merlin-northern:
> Oh wait, now I see. The reason it was done like this is because the removal is inside the check for `mender-image`. If `mender-image` is not set, then the creation of the full path is necessary. So in short it means:
> 
> * With `mender-image`: `/var/lib/mender` is a symlink.
> * Without `mender-image`: `/var/lib/mender` is a directory.
> 
> So the correct fix is actually to do the same in mender-monitor: Create the directory in the main install block, remove it and replace with a symlink in the `mender-image` block. Sorry, I missed this when first reviewing.

ok, thanks. closing this one then.

PR #1463, opened 2021-09-01
mender-monitor_git: define unique PV based on SRCREV

lluiscampos:
This will force yocto to rebuild from source every time that the
filename of the tarball changes.

Now the PV for the git recipe will be master-git${SRCREV}

This change aligns this recipe with other Mender components that are
build from source, though in this case SRCREV is extracted from the
filename and not from an actual Git repository.

### DIFF
--- a/meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor_git.bb
+++ b/meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor_git.bb
@@ -1,6 +1,30 @@
 require mender-monitor.inc
 
-def mender_monitor_version_from_preferred_version(d):
+def mender_monitor_srcrev_from_src_uri(d, src_uri):
+    pref_version = d.getVar("PREFERRED_VERSION")
+    if pref_version is not None and pref_version.find("-build") >= 0:
+        # If "-build" is in the version, SRCREV won't be used for defining PV
+        return ""
+    else:
+        # Extract SRCREV from the tarball filename
+        import glob
+        import re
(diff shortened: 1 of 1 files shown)


lluiscampos:
Goes with https://github.com/mendersoftware/mender-qa/pull/527, see comments there.

meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor_git.bb
```diff
@@ -1,16 +1,41 @@
...
+    pref_version = d.getVar("PREFERRED_VERSION")
+    if pref_version is not None and pref_version.find("-build") >= 0:
+        # If "-build" is in the version, SRCREV won't be used for defining PV
+        return ""
+    else:
+        # Extract SRCREV from the tarball filename
+        import glob
+        import re
+        # Remove "file://" prefix
+        src_uri_glob = src_uri[len("file://"):]
```
kacf, line 13: I would use `src_uri_glob = src_uri.split()[0][len("file://"):]`, otherwise multiple `SRC_URI` entries will confuse it. It requires that the tarball is the first entry, but that is more reasonable to assume.

kacf (approved):

meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor_git.bb
lluiscampos: Good suggestion. Thanks

PR #1465, opened 2021-09-06
Consolidate two main U-Boot patches into one.

kacf:
This makes more sense, since both have one single purpose: Enable QEMU
support for vexpress_ca9x4.

Changelog: None


kacf:

PR #1469, opened 2021-09-08
meta-mender-ci: Support MENDER_CI_LOGLEVEL for mender-monitor

lluiscampos:
Aligning it with the other Mender components, see commit f3b44b9.

### DIFF
--- a/tests/meta-mender-ci/recipes-mender/mender-monitor/mender-monitor_%.bbappend
+++ b/tests/meta-mender-ci/recipes-mender/mender-monitor/mender-monitor_%.bbappend
@@ -1,0 +1,18 @@
+# Produce custom level logs when MENDER_CI_LOGLEVEL is enabled
+do_install_append_class-target_mender-image_mender-systemd() {
+    if ${@'true' if d.getVar('MENDER_CI_LOGLEVEL') else 'false'}; then
+        local logging_level=""
+        if [ "${MENDER_CI_LOGLEVEL}" = "debug" ]; then
+            logging_level="DEBUG"
+        elif [ "${MENDER_CI_LOGLEVEL}" = "trace" ]; then
+            logging_level="TRACE"
+        fi
+
+        if [ -n "$logging_level" ]; then
+            sed -i "s|CONFIG_LOG_LEVEL=.*|CONFIG_LOG_LEVEL=${logging_level}|" \
+                ${D}/${datadir}/mender-monitor/config/config.sh
+        else
+            bbwarn "Unknown MENDER_CI_LOGLEVEL ${MENDER_CI_LOGLEVEL} for mender-monitor. Ignoring."
+        fi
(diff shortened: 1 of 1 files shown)


kacf (approved):

tests/meta-mender-ci/recipes-mender/mender-monitor/mender-monitor_%.bbappend
```diff
@@ -0,0 +1,16 @@
+# Produce custom level logs when MENDER_CI_LOGLEVEL is enabled
+do_install_append_class-target_mender-image_mender-systemd() {
+    if ${@'true' if d.getVar('MENDER_CI_LOGLEVEL') else 'false'}; then
+        local logging_level=""
+        if [ "${MENDER_CI_LOGLEVEL}" = "debug" ]; then
```
merlin-northern, line 5: I would also check if the `MENDER_CI_LOGLEVEL` is among the: `FATAL TRACE ERROR DEBUG WARN INFO` otherwise fallback to one of them.

tests/meta-mender-ci/recipes-mender/mender-monitor/mender-monitor_%.bbappend
```diff
@@ -0,0 +1,16 @@
...
+            logging_level="DEBUG"
+        fi
+
+        if [ -n "$logging_level" ]; then
+            sed -i "s|CONFIG_LOG_LEVEL=.*|CONFIG_LOG_LEVEL=${logging_level}|" \
```
merlin-northern, line 10: nit pick, ignore if you feel like it:
* what if for some reason config/config.sh does not contain `CONFIG_LOG_LEVEL`?
* what if it contains: `# CONFIG_LOG_LEVEL = `?

tests/meta-mender-ci/recipes-mender/mender-monitor/mender-monitor_%.bbappend
lluiscampos: Thanks. In practice this will only be used for debug in our CI, so I'll check for `DEBUG` and `TRACE` (to align it with mender-connect).

tests/meta-mender-ci/recipes-mender/mender-monitor/mender-monitor_%.bbappend
lluiscampos: Only to be used for local debugging, so it is not worth worrying too much. See also how is done for `mender` and `mender-connect`.

tests/meta-mender-ci/recipes-mender/mender-monitor/mender-monitor_%.bbappend
lluiscampos: Added support for `trace` :+1:

PR #1470, opened 2021-09-08
MEN-5012: Remove `PACKAGECONFIG = "dbus"` for new Mender versions.

kacf:
It is no longer possible to compile the Mender client without dbus
support, and glib-2.0 is now a required dependency.

Changelog: Commit


meta-mender-core/recipes-mender/mender-client/mender-client.inc
```diff
@@ -272,23 +267,17 @@ do_install() {
...
+    oe_runmake \
+        -C ${B}/src/${GO_IMPORT} \
+        V=1 \
+        prefix=${D} \
+        datadir=${datadir} \
+        install-dbus
+
+    # install the D-Bus interface file(s)
+    install -d ${D}${datadir}/dbus-1/interface
+    install -m 0644 ${B}/src/${GO_IMPORT}/Documentation/io.mender.*.xml ${D}${datadir}/dbus-1/interface
```
oleorhagen, line 280: Question: Should we consider adding this to the `Makefile`?

oleorhagen (approved):

meta-mender-core/recipes-mender/mender-client/mender-client_2.5.0.bb
```diff
@@ -1,4 +1,4 @@
-require mender-client.inc
+require mender-client-pre-3.1.inc
```
lluiscampos, line 1: It should be pre 3.2, right? Mender client 3.1 we are tentatively releasing in few weeks from now, and this feature goes into the following one, 3.2.

tests/acceptance/test_build.py
```diff
@@ -1251,40 +1214,29 @@ def test_mender_dbus_interface_file(
...
-                bitbake_image,
-                target="-c clean mender-client",
-            )
+        # clean up the mender-client
+        build_image(
+            prepared_test_build["build_dir"],
+            prepared_test_build["bitbake_corebase"],
+            bitbake_image,
+            target="-c clean mender-client",
+        )
```
lluiscampos, line 1223: I don't think we need this clean call anymore.

lluiscampos (approved):

meta-mender-core/recipes-mender/mender-client/mender-client_2.5.0.bb
kacf: Yes! Good catch!

tests/acceptance/test_build.py
kacf: Probably not, but I think it's worth having as a safeguard anyway. A bit like doing `rm logfile && echo "first entry" > logfile`.

meta-mender-core/recipes-mender/mender-client/mender-client.inc
kacf: Good idea, I will add it.

PR #1471, opened 2021-09-09
monitor: Fix parse error when SRC_URI is empty.

kacf:
Which it is when not specifically set up to build the component.

Changelog: None


kacf:
I'm not adding a changelog entry for this, because this error was never included in any tag.

kacf:

lluiscampos (approved):

PR #1473, opened 2021-09-10
Revert "meta-mender-ci: Support MENDER_CI_LOGLEVEL for mender-monitor"

lluiscampos:
This reverts commit c41135b4816e4c00e405b28ede688d038012957b.


lluiscampos:
@merlin-northern @kacf I apologize for having merged without running the tests. I have no good excuse, I was just short on my judging of the implications.

We cannot really add anything `meta-mender-commercial` related in layer `meta-mender-ci` as `commercial` is not added by default in acceptance tests. If we want to do it, we should have a different layer like meta-mender-commercial-ci`.

I don't think is worth the effort because 1) this was meant to be used only locally and 2) we don't really have any runtime acceptance tests for `mender-monitor` (we have plenty integration tests, but the image there won't have `meta-mender-ci` layer anyway).

Sorry :weary:

merlin-northern (approved):

kacf:
> We cannot really add anything `meta-mender-commercial` related in layer `meta-mender-ci` as `commercial` is not added by default in acceptance tests. If we want to do it, we should have a different layer like meta-mender-commercial-ci`.

Although not quite as clean as keeping it in meta-mender-ci, you could consider using the `mender-testing-enabled` override and keep it in meta-mender-commercial. That would prevent it from being used by anyone who hasn't enabled that override, and I would consider it acceptable for such a small thing.

PR #1475, opened 2021-09-13
Add images used for giving mender demos

TheMeaningfulEngineer:
Cherry picked commit from the [Dunfell PR I created first by accident](https://github.com/mendersoftware/meta-mender/pull/1339).
Creating a PR for master first, after that was merged [will merge Dunfell](https://github.com/mendersoftware/meta-mender/pull/1339#issuecomment-893263626)

### DIFF
--- a/meta-mender-demo/recipes-extended/images/mender-demo-full-delta.bb
+++ b/meta-mender-demo/recipes-extended/images/mender-demo-full-delta.bb
@@ -1,0 +1,9 @@
+require recipes-extended/images/mender-demo-full-rw.bb
+
+IMAGE_FEATURES_append = " read-only-rootfs"
+IMAGE_INSTALL_append = " mender-binary-delta"
+
+# See https://tracker.mender.io/browse/MEN-3513 and
+# https://tracker.mender.io/browse/MEN-3781 and
+# https://tracker.mender.io/browse/MEN-3912.
+EXTRA_IMAGECMD_ext4_append = "${@bb.utils.contains('IMAGE_FEATURES', 'read-only-rootfs', ' -O ^64bit -O ^has_journal', '', d)}"
--- a/meta-mender-demo/recipes-extended/images/mender-demo-full-rw.bb
+++ b/meta-mender-demo/recipes-extended/images/mender-demo-full-rw.bb
@@ -1,0 +1,6 @@
+require recipes-extended/images/core-image-full-cmdline.bb
+
+IMAGE_FEATURES += "ssh-server-openssh allow-empty-password debug-tweaks"
+IMAGE_INSTALL_append = " mender-configure"
+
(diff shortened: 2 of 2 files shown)


meta-mender-demo/recipes-extended/images/mender-demo-full-delta.bb
```diff
@@ -0,0 +1,9 @@
+require recipes-extended/images/mender-demo-full-rw.bb
+
+IMAGE_FEATURES_append = " read-only-rootfs"
+IMAGE_INSTALL_append = " mender-binary-delta"
+
+# See https://tracker.mender.io/browse/MEN-3513 and
+# https://tracker.mender.io/browse/MEN-3781 and
+# https://tracker.mender.io/browse/MEN-3912.
+EXTRA_IMAGECMD_ext4_append = "${@bb.utils.contains('IMAGE_FEATURES', 'read-only-rootfs', ' -O ^64bit -O ^has_journal', '', d)}"
```
kacf, line 9: This shouldn't be necessary, since this fix [already exists in the meta-mender-commercial layer](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-commercial/conf/layer.conf), which you are forcing to include because of `MENDER_INSTALL_append = " mender-binary-delta"`.

TheMeaningfulEngineer: Thanks, would that hold true for all images then?
([This one](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-demo/recipes-extended/images/mender-image-full-cmdline-rofs.bb) was already in the demo layer.)

kacf: No, not all images. The one you pointed to is there because it is not normally built with the meta-mender-commercial layer. The reason is a bit complicated, but it comes down to security: We have to build an image without commercial components, and then add them in later. This is only for testing, so it doesn't affect the image you are introducing, which should always build with meta-mender-commercial enabled.

meta-mender-demo/recipes-extended/images/mender-demo-full-delta.bb
```diff
@@ -0,0 +1,9 @@
+require recipes-extended/images/mender-demo-full-rw.bb
+
```
lluiscampos, line 2: I think we should be more future ready and name this image `mender-demo-full-rw-commercial`, having delta updates and later on mender-monitor and any other commercial component that we add.

lluiscampos (approved):
I think we should rename the image. Also please elaborate a bit more the changelog. I suggest two, really:
* Changelog: Add `mender-demo-full-rw` image suitable to showcase Mender features
* Changelog: Add `mender-demo-full-delta` image suitable to showcase Mender Enterprise features (or `mender-demo-full-rw-commercial` as I suggest in my comment below)

lluiscampos:
@TheMeaningfulEngineer Same for this one :smiley:

PR #1476, opened 2021-09-14
Release Mender 3.1.0

lluiscampos:
Changelog: Add recipe for mender-client 3.1.0
Changelog: Add recipe for mender-artifact 3.6.1
Changelog: Add recipe for mender-monitor 1.0.0
Changelog: Add recipe for mender-binary-delta 1.3.0
Changelog: Add recipe for mender-configure 1.0.2

### DIFF
--- a/meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor_git.bb
+++ b/meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor_git.bb
@@ -23,7 +23,9 @@ def mender_monitor_srcrev_from_src_uri(d, src_uri):
         # Now extract the git sha from the filename
         m = re.match(r".*/mender-monitor-([0-9a-f]+)\.tar\.gz", filename)
         if m is None:
-            bb.error("Cannot extract git sha from filename: %s" % filename)
+            # No match probably means that the tarball is a tagged version, in
+            # which case this recipe is not to be used but still needs to parse
+            return ""
         return m.group(1)
 
 SRCREV = "${@mender_monitor_srcrev_from_src_uri(d, '${SRC_URI}')}"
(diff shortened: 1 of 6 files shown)


kacf (approved):

oleorhagen (approved):

meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor_git.bb
```diff
@@ -23,7 +23,9 @@ def mender_monitor_srcrev_from_src_uri(d, src_uri):
         # Now extract the git sha from the filename
         m = re.match(r".*/mender-monitor-([0-9a-f]+)\.tar\.gz", filename)
         if m is None:
-            bb.error("Cannot extract git sha from filename: %s" % filename)
+            # No match probably means that the tarball is a tagged version, in
+            # which case this recipe is not to be used but still needs to parse
+            return ""
```
lluiscampos, line 28: This is somewhat temporary to build the release in hand.

I think we should actually handle the final tags case in `mender_monitor_version_from_preferred_version` so that we skip calling this function all together. But then, it feels very weird to set a `PV` of `1.0.0` on a `mender-monitor_git.bb` recipe...

kacf: I think it will be called regardless, since the recipe is always parsed, even if not used. In the Mender client [we just set it to a known SHA](https://github.com/mendersoftware/meta-mender/blob/5425e36038b2b4d2b8ed6922c2a24fe6683b87ad/meta-mender-core/recipes-mender/mender-client/mender-client_git.inc#L17-L18). I'm not sure if that's appropriate here, since the repository, and therefore the SHAs, are not exposed. But maybe it will be confused by an empty value? I guess we'll see..

lluiscampos: It works, that much I can say :)

PR #1477, opened 2021-09-21
Mender monitor versioning

TheMeaningfulEngineer:
Managed to get a successful build locally.
Under way of testing the actual image.
Made some assumptions on the usage which I'd hope you can check.

Made me question if we need to have version specific recipes in production at all for this or binary delta.
In both cases the primary source of truth for the version is the downloaded archive so all of those `PREFFERED_VERSIONS` set up is just added complexity which can go wrong without unclear added value.
What do you think?

### DIFF
--- a/meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor_git.bb
+++ b/meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor_git.bb
@@ -1,5 +1,7 @@
 require mender-monitor.inc
 
+# This is the developent recipe for which the recipe version gets defined from the downloaded archive name.
+
 def mender_monitor_srcrev_from_src_uri(d, src_uri):
     pref_version = d.getVar("PREFERRED_VERSION")
     if pref_version is not None and pref_version.find("-build") >= 0:
@@ -17,13 +19,16 @@ def mender_monitor_srcrev_from_src_uri(d, src_uri):
         src_uri_glob = src_uri_list[0][len("file://"):]
         # Get the filename
         filenames = glob.glob(src_uri_glob)
-        if len(filenames) != 1:
-            bb.error("Expected exactly one file, found: %s" % filenames)
-        filename = filenames[0]
-        # Now extract the git sha from the filename
-        m = re.match(r".*/mender-monitor-([0-9a-f]+)\.tar\.gz", filename)
+        if len(filenames) == 0:
+            bb.error("Failed to find mender monitor on path %s" % src_uri_glob)
+            bb.error("Please make sure SRC_URI_pn-mender-monitor is pointing to the downloaded tarball ")
+        elif len(filenames) != 1:
+            bb.warning("Expected exactly one file, found: %s" % filenames)
+        filename = os.path.basename(filenames[0])
+        # Now extract the SRCREV from the filename
+        m = re.match(r"mender-monitor-(.*)\.tar\.gz", filename)
         if m is None:
-            bb.error("Cannot extract git sha from filename: %s" % filename)
+            bb.error("Cannot extract SRCREV from filename: %s" % filename)
         return m.group(1)
 
 SRCREV = "${@mender_monitor_srcrev_from_src_uri(d, '${SRC_URI}')}"
@@ -36,8 +41,7 @@ def mender_monitor_version_from_preferred_version(d, srcrev):
         # final tags, which will need their own recipe.
         return pref_version
     else:
-        # Else return "master-git${SRCREV}".
-        return "master-git%s" % srcrev
+        return "%s" % srcrev
 
 PV = "${@mender_monitor_version_from_preferred_version(d, '${SRCREV}')}"
 


kacf (changes requested):
The first commit looks good to me, although @lluiscampos probably should take a look as well.

A couple of change requests/questions for the second commit:

* I'm not sure about the renaming. At release time we will have real dedicated recipes, and the "git/tarball" recipe will be reserved just for our debugging. In this context I think the name "git" is still better than "tarball", even though it's lying a little bit, since "tarball" can be misunderstood by users to mean "any tarball". However, their tarballs will be handled by dedicated recipes.

* If we do keep the rename, then make sure to update the changelog. It now says "title", but the title doesn't say which component you have changed.

TheMeaningfulEngineer:
Thanks @kacf 
Can you please share your thoughts on using the tarball as a source of truth for all cases also thereby removing the need for new recipes once a new tarball is installed.

An example for `mender-binary-delta-1.2.1.tar.xz` but the same would apply for monitor.

* the prebuilt tarball which is manually downloaded is the single point of truth for the version
* yocto can't choose a version of the recipe, it just needs to accept whatever tarball it is being fed

Assuming above true, what's the value of maintaining a dedicated recipe for each version (i.e. `mender-binary-delta_1.2.0.bb`, `mender-binary-delta_1.2.1.bb` ...) ?

kacf:
I don't think the tarball should be the source of truth, because this hides upgrades. If we upgrade mender-binary-delta, the new recipe will take precedence, it will not match the tarball, and the recipe will fail, alerting the user that he needs to download a new tarball. If we make the tarball the source of truth, the user will never know that he should upgrade. That's why we want dedicated recipes.

TheMeaningfulEngineer:
@kacf 
Thanks, I get where you're coming from now.
Your goal is something like:

> The HEAD for a `meta-mender` branch should always attempt to build the latest bugfix version of mender apps

I guess I'm open to sacrifice the safer flow (`You need to download the tarball which matches the version you defined in yocto`) for a potentially simpler flow (`What's in the tarball is what you get`). At the end of the day we should probably revisit the "manual download of the tarball", solving that will get us both on the same page.

 Anyways, I am all good with rolling with your approach, just wanted to see the responses for an alternative. :)

kacf (approved):
@lluiscampos Ok with you?

meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor_git.bb
```diff
@@ -36,8 +41,7 @@ def mender_monitor_version_from_preferred_version(d, srcrev):
         # final tags, which will need their own recipe.
         return pref_version
     else:
-        # Else return "master-git${SRCREV}".
-        return "master-git%s" % srcrev
+        return "%s" % srcrev
```
lluiscampos, line 44: This won't anymore match [what `mender-qa` sets for CI builds](https://github.com/mendersoftware/mender-qa/blob/master/scripts/yocto-build-and-test.sh#L129). The trick is to have a known prefix that we can use from CI.

Right now it "works" because we don't have more recipes, but once 1.0.0 is out we still need a way from CI to prefer the master one

meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor_git.bb
```diff
@@ -17,13 +19,16 @@ def mender_monitor_srcrev_from_src_uri(d, src_uri):
...
-        # Now extract the git sha from the filename
-        m = re.match(r".*/mender-monitor-([0-9a-f]+)\.tar\.gz", filename)
+        if len(filenames) == 0:
+            bb.error("Failed to find mender monitor on path %s" % src_uri_glob)
+            bb.error("Please make sure SRC_URI_pn-mender-monitor is pointing to the downloaded tarball ")
+        elif len(filenames) != 1:
+            bb.warning("Expected exactly one file, found: %s" % filenames)
+        filename = os.path.basename(filenames[0])
+        # Now extract the SRCREV from the filename
+        m = re.match(r"mender-monitor-(.*)\.tar\.gz", filename)
```
lluiscampos, line 29: I think this is too open. I think your fix is good intended (allow master) but we should try to be as strict as possible. I hence suggest:
```suggestion
        m = re.match(r"mender-monitor-(master|[0-9a-f]+)\.tar\.gz", filename)
```

lluiscampos: If you really push it, we could even be strict in how long the git sha can be:
```suggestion
        m = re.match(r"mender-monitor-(master|[0-9a-f]7)\.tar\.gz", filename)
```

lluiscampos (changes requested):
@TheMeaningfulEngineer I am sorry I was not able to give my feedback before... and that now I come back with suggestions. See comments below and let's discuss there

lluiscampos:
Also, I didn't say... Thank you very much for following up with this! It is really good that not only internally but also externally we will have a way to test the master version of this component just following docs.mender.io instructions :+1:

lluiscampos:
@TheMeaningfulEngineer Ups, the stale bot is pressuring us :)

If you don't have the capacity, I am glad to take over your work and finalize it :+1:

TheMeaningfulEngineer:
@lluiscampos 
Hey,
that would be great thanks.
Have totally fallen out from the topic

lluiscampos:
Resubmitted at #1536.

PR #1511, opened 2021-11-04
Fix FTBFS due to host contamination with secure boot enabled

phatina:
meta-secure-core defines `password.inc` which triggers an error if I want to use secure boot:

```
to host contamination.
```

Removing `password.inc` file fixes this.

Please, review. Thanks.


kacf (changes requested):
Thanks! The code looks good to me, but can you please submit this to the master branch? We always merge there first, and then cherry-pick backwards from there.

phatina:
> Thanks! The code looks good to me, but can you please submit this to the master branch? We always merge there first, and then cherry-pick backwards from there.

Oh, sure thing. Will do. Thanks

phatina:
Done, see #1512. Thanks.

kacf:
Closing this, the fix will be cherry-picked from master separately.

phatina:
> Closing this, the fix will be cherry-picked from master separately.

Can you cherry pick this change from master to dunfell, please? So that I can rebase in my project. Thanks!

kacf:
Fixed in #1515.

PR #1512, opened 2021-11-04
Fix FTBFS due to host contamination with secure boot enabled [master]

phatina:
meta-secure-core defines `password.inc` which triggers an error if I want to use secure boot:

```
to host contamination.
```

Removing `password.inc` file fixes this.

Please, review. Thanks.


kacf (approved):
Thanks!

PR #1513, opened 2021-11-06
Provide options to customize mender kickstart file

ahsanhussain:
This patch provides the a variable to give user an option to keep the RFS partition B empty for the generated image, which results in a significantly smaller image file and quicker deployment/transfer etc. The default behavior however has not changed and we still get a populated partition B, unless changed by the user


ahsanhussain:
Hmm, rawcopy overwrites what's done by partition.py wic plugin. I guess we don't need the first commit. Removing

meta-mender-core/classes/mender-part-images.bbclass
```diff
@@ -43,6 +43,8 @@ IMAGE_NAME_SUFFIX = ""
 # Block storage
 ################################################################################
 
+MENDER_RFS_PART_EMPTY_B ??= "n"
```
kacf, line 46: We should follow the existing naming, IOW this should be `MENDER_ROOTFS_PART_B_EMPTY`.

meta-mender-core/conf/mender-vars.json
```diff
@@ -156,5 +156,6 @@
     "MENDER_CONNECT_USER": "",
     "MENDER_UPDATE_CONTROL_MAP_EXPIRATION_TIME_SECONDS": "",
     "MENDER_UPDATE_CONTROL_MAP_BOOT_EXPIRATION_TIME_SECONDS": "",
-    "MENDER_CI_LOGLEVEL": ""
+    "MENDER_CI_LOGLEVEL": "",
+    "MENDER_RFS_PART_EMPTY_B": ""
```
kacf, line 160: While you're switching name, maybe consider putting this up next to the others with similar names. I know, this isn't fully sorted, but might as well do it for this variable.

meta-mender-core/classes/mender-part-images.bbclass
kacf, line 46: I would even go one step further, and say that this should be the default. I can't really see any good reason why partition B should be populated at all.

It is possible this will fail in our platform tests though, I'm not sure about the assumptions they make. Since this is pretty easy to test, I submitted a simpler version [here](https://github.com/mendersoftware/meta-mender/pull/1519). Let's see how it goes before proceeding with this.

kacf (commented):
Good idea. I've attempted a simpler version of this [here](https://github.com/mendersoftware/meta-mender/pull/1519) (see comment below).

kacf:
Superseded by #1519.

PR #1515, opened 2021-11-08
Fix FTBFS due to host contamination with secure boot enabled

kacf:
to host contamination.

Changelog: Fix FTBFS due to host contamination with secure boot enabled

Signed-off-by: Peter Hatina <peter@hatina.eu>
(cherry picked from commit 490aa4120961316d201b2878fac1f8ac07dcbf85)


kacf:
Only unrelated failures. Merging.

PR #1532, opened 2021-11-24
MEN-5137: Add a demo recipe for the monitor-client example monitors

oleorhagen:
Changelog: Title

### DIFF
--- a/meta-mender-demo/conf/layer.conf
+++ b/meta-mender-demo/conf/layer.conf
@@ -1,5 +1,5 @@
 # Layer configuration for meta-mender
-# Copyright 2017 Northern.tech AS
+# Copyright 2022 Northern.tech AS
 
 # We have a conf and classes directory, add to BBPATH
 BBPATH .= ":${LAYERDIR}"
@@ -7,6 +7,10 @@ BBPATH .= ":${LAYERDIR}"
 # We have recipes-* directories, add to BBFILES
 BBFILES += "${LAYERDIR}/recipes-*/*/*.bb \
 	${LAYERDIR}/recipes-*/*/*.bbappend"
+
+BBFILES_DYNAMIC += "\
+  mender-commercial:${LAYERDIR}/mender-commercial/mender-monitor/mender-monitor*.bbappend \
+"
 
 BBFILE_COLLECTIONS += "mender-demo"
 BBFILE_PATTERN_mender-demo = "^${LAYERDIR}/"
--- a/meta-mender-demo/mender-commercial/mender-monitor/mender-monitor_%.bbappend
+++ b/meta-mender-demo/mender-commercial/mender-monitor/mender-monitor_%.bbappend
@@ -1,0 +1,8 @@
+
+
+do_install_append() {
+    oe_runmake \
+        -C ${S} \
+        DESTDIR=${D} \
+        install-example-monitors
+}


kacf (approved):

lluiscampos (changes requested):
This won't work. It is implicitly introducing a dependency between `meta-mender-demo` and `meta-mender-commercial`. IOW, bitbake won't be able to parse this `.bbapend` unless you also have included `meta-mender-commercial`, which won't be the case for most of the users.

Either we create a new demo layer `meta-mender/meta-mender-commercial-demo` or use existing layer with a different recipe/package (`mender-monitor-demo` ?).

By the way, in any case we probably need to update docs and `mender-qa` build script.

oleorhagen:
> This won't work. It is implicitly introducing a dependency between `meta-mender-demo` and `meta-mender-commercial`. IOW, bitbake won't be able to parse this `.bbapend` unless you also have included `meta-mender-commercial`, which won't be the case for most of the users.
> 
> Either we create a new demo layer `meta-mender/meta-mender-commercial-demo` or use existing layer with a different recipe/package (`mender-monitor-demo` ?).
> 

It seems one is able to work around this 'limitation' with `BBFILES_DYNAMIC`, and adding the append recipe only in the case the case the commercial layer is enabled.

> By the way, in any case we probably need to update docs and `mender-qa` build script.

In this way, at least we won't have to touch `mender-qa`.

I will have a look through the docs now.

meta-mender-demo/dynamic-recipes/mender-monitor/mender-monitor_%.bbappend
```diff
@@ -0,0 +1,8 @@
+
+
+do_install_append() {
+    oe_runmake \
+        -C ${S} \
+        DESTDIR=${D} \
+        enable-example-monitors
```
oleorhagen, line 7: Install and symlink the example monitors, not just simply install.

meta-mender-demo/conf/layer.conf
```diff
@@ -8,6 +8,10 @@ BBPATH .= ":${LAYERDIR}"
 BBFILES += "${LAYERDIR}/recipes-*/*/*.bb \
 	${LAYERDIR}/recipes-*/*/*.bbappend"
 
+BBFILES_DYNAMIC += "\
+  mender-commercial:${LAYERDIR}/recipes-*/*/mender-monitor*.bbappend \
```
lluiscampos, line 12: This still doesn't work because we need to exclude, somehow, these recipes from the `BBFILES +=` above. I get:

```
ERROR: No recipes available for:
  /home/lluis/poky/meta-mender/meta-mender-demo/recipes-mender/mender-monitor/mender-monitor_%.bbappend
```

I don't know the best way to do it, though. Easy suggestion: we have an explicit directory for the recipes belonging to `mender-commercial` collection, like:
```
meta-mender-demo/recipes-mender/...                                    <- bbappends to core recipes
meta-mender-demo/mender-commercial/recipes-mender/...   <- bbappends to commercial recipes
```

This way they are exluded in `BBFILES` and can easily be included on DYNAMIC with:
```
BBFILES_DYNAMIC += "\
  mender-commercial:${LAYERDIR}/mender-commercial/recipes-*/*/*.bbappend \
```

(extra advantage of this approach is that we don't need to further modify `layer.conf` when adding recipes there)

lluiscampos (changes requested):
See comment below. The fact that still doesn't work makes me think that we should have a test for this...

Also, nitpick, but the commit title (aka changelog) says "monitor-client" but from yocto point of view the recipe is "mender-monitor". Please amend :pray:

meta-mender-demo/dynamic-recipes/mender-monitor/mender-monitor_%.bbappend
lluiscampos: By the way, is this comment not applicable anymore?

meta-mender-demo/conf/layer.conf
oleorhagen: Hmm, interesting, previously I did what you mentioned above, but then I did not want another repo, so I changed it back, and it worked for me.

Let me retest it.

oleorhagen: I sincerely thought it would mask them from BBFILES, but maybe I was fooled.

oleorhagen: Added to a separate directory `mender-commercial`

meta-mender-demo/dynamic-recipes/mender-monitor/mender-monitor_%.bbappend
oleorhagen: Nope, the documentation instructs the user to enable the one he/she is testing manually, so that we can snipe out which alerts to show during demo.

oleorhagen:

oleorhagen:

oleorhagen:

lluiscampos (approved):

PR #1539, opened 2021-12-07
Add swap to fstab when swap partition is enabled

nandra:


kacf (approved):
:+1:

kacf:
@nandra: Oops, I missed that this was targeted to dunfell. Can you submit this to master and we will cherry-pick afterwards instead.

nandra:
@kacf opened: #1540

PR #1540, opened 2021-12-10
Add swap to fstab when swap partition is enabled

nandra:


kacf (approved):

kacf:
Error is unrelated, merging.

Thanks @nandra! This will be cherry-picked to dunfell during its next update.

PR #1561, opened 2022-01-04
prevent pseudo inode mismatch w/ efi-secure-boot

coreycothrum:
I get pseudo/abort inode mismatch errors on `LockDown.efi` when doing
back-to-back builds. The first build works fine. Building again (without cleaning and without changes)
fails.

I'm using the latest versions of dunfell.

`pseudo.log`:
```
path mismatch [2 links]: ino 1313785
db '/production-image/1.0-r0/rootfs/boot/efi/EFI/BOOT/LockDown.efi'
req '/production-image/1.0-r0/bootfs.image_uefiimg/EFI/BOOT/LockDown.efi'. 
```

Adding `PSEUDO_DSIABLE=1` to the rm/cleanup in
mender_merge_bootfs_and_image_boot_files resolves this issue, but I'm not 100% sure if this is the real fix
or just patching the symptoms.

I've looked [meta-efi-secure-boot](https://github.com/jiazhang0/meta-secure-core/tree/master/meta-efi-secure-boot/recipes-bsp/efitools) and don't see anything wrong/weird with `LockDown.efi`

----
What do y'all think?

### DIFF
--- a/meta-mender-core/classes/mender-helpers.bbclass
+++ b/meta-mender-core/classes/mender-helpers.bbclass
@@ -275,7 +275,7 @@ mender_merge_bootfs_and_image_boot_files() {
     W="${WORKDIR}/bootfs.${BB_CURRENTTASK}"
     rm -rf "$W"
 
-    cp -al "${IMAGE_ROOTFS}/${MENDER_BOOT_PART_MOUNT_LOCATION}" "$W"
+    cp -a "${IMAGE_ROOTFS}/${MENDER_BOOT_PART_MOUNT_LOCATION}" "$W"
 
     # Put in variable to avoid expansion and ';' being parsed by shell.
     image_boot_files="${IMAGE_BOOT_FILES}"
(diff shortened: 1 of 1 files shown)


kacf:
To be perfectly honest I don't fully understand neither the problem, nor the fix. Why is it a problem to delete something and recreate it, even under pseudo?

It's using hardlinks, from the looks of it (see several `cp -l` instances). Could this be handled badly by pseudo?

kacf:
FYI, we might get a chance to try to reproduce this in a few days, so we should have some updates soon.

coreycothrum:
> It's using hardlinks, from the looks of it (see several `cp -l` instances). Could this be handled badly by pseudo?

That might be it. I hadn't considered that before you said this.

I've just tried it with:
1. reverting my `PSEDUO_DISABLE=1` change
2. removing the `-l` option from `cp` for both instances `mender_merge_bootfs_and_image_boot_files ` ([here](https://github.com/mendersoftware/meta-mender/blob/08f632b7786772be6e846a9ec6e89bd859e813c4/meta-mender-core/classes/mender-helpers.bbclass#L278) and [here](https://github.com/mendersoftware/meta-mender/blob/08f632b7786772be6e846a9ec6e89bd859e813c4/meta-mender-core/classes/mender-helpers.bbclass#L310))

Seems to work for me. I think that is probably a better fix than what I've currently committed in this PR. Agree? If so I'll rebase with that change.

kacf:
Yes, I agree, that's better!

kacf (approved):

PR #1562, opened 2022-01-07
[dunfell] tests: Adapt to new polling policy for Update Control Maps in the client.

kacf:
The client now uses the shortest interval among
`UpdatePollIntervalSeconds` and
`UpdateControlMapExpirationTimeSeconds` when deciding how long to wait
before re-querying the server for an Update Control Map. So we can no
longer use a quick update check time together with a slow Update
Control Map time. Instead, use the same time for both, and use the
`check-update` flag to force a quick update check at the points where
we deploy.

Changelog: None

(cherry picked from commit dccf39a1085111a7599f9941ffa34c9d663bf244)


kacf:

PR #1564, opened 2022-01-10
Fix incorrect checksum detection on version branches.

kacf:
When using `PREFERRED_VERSION_mender-client` instead of
`PREFERRED_VERSION_pn-mender-client`, the recipe doesn't correctly
pick up the branch name, causing checksum verification to fail. The
former is what is documented in the Yocto manual, so let's make sure
we support that.

Changelog: none


lluiscampos (approved):

kacf:
The failures are due to me forgetting that 2.6.x and 3.0.x are not supported for meta-mender/master. The 3.1.x pipeline is what counts. Merging.

PR #1571, opened 2022-01-11
do not merge: mender-connect licence chksum update

merlin-northern:
ChangeLog:none


kacf:
For some reason the license year update was done on a release branch before master in mender-connect, which caused this build failure. So this is incorrect, you're just reverting what I just fixed for the release branches. For mender-connect/master, the fix is [here](https://github.com/mendersoftware/mender-connect/pull/55).

PR #1576, opened 2022-01-20
mender-client: clone the repo using the tag set in PV variable

clementperon:
Hi @lluiscampos,

I would like to know why you don't use directly the ${PV} to set the SRCREV.

Is it a bad practice? IMO this avoid a bad manipulation the SHA1


lluiscampos:
@clementperon I am not an expert, but according to [the docs](https://docs.yoctoproject.org/ref-manual/variables.html#term-SRCREV) both can work work but the tag needs to be re-feched. This makes senes to me because from Git point of view, tags are just like an "alias" to a commit and can be updated - so the only real way to make sure you are building an specific commit is with the commit sha.

PR #1581, opened 2022-01-25
meta-mender-demo: Move mender-monitor bbappend to 1.1.x

lluiscampos:
The make target is not available for 1.0.x

This is a temporary fix while I work on a better one that includes the
git recipe (and that it is tested in CI ;)


lluiscampos:
Closing in favor of #1582.

PR #1586, opened 2022-02-02
3.2.1 release version bumps

oleorhagen:
1) client 3.2.1

2) mender-connect recipe 2.0.1


kacf (changes requested):
Unless there is a security issue, we should keep the original patch recipes in the same series. Then people can fall back if the new version introduces a regression for them.

kacf (approved):

lluiscampos (approved):

PR #1591, opened 2022-02-09
Support more standardized install targets in mender-artifact Makefile.

kacf:
Changelog: None


kacf:
Goes together with https://github.com/mendersoftware/mender-artifact/pull/350.

PR #1599, opened 2022-02-16
MEN-5300: add a recipe for mender-gateway

lluiscampos:
Adds a closed source Yocto recipe for mender-gateway. Only the git
recipe is included in this commit. The recipe takes a tarball containing
the pre-compiled binaries and support files and just installs them on
the target.

A piece of code from mender-monitor has been refactored into a common
class for reuse in all closed source recipes that define SRCREV from the
tarball filename.

Acceptance test added accordingly.

Changelog: Add recipe for mender-gateway_git
Changelog: Add testing image mender-gateway-image-full-cmdline

### DIFF
--- a/meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
+++ b/meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
@@ -1,0 +1,49 @@
+LICENSE = "CLOSED"
+LICENSE_FLAGS = "commercial"
+
+inherit systemd
+
+SUB_FOLDER_arm = "arm"
+SUB_FOLDER_aarch64 = "aarch64"
+SUB_FOLDER_x86-64 = "x86_64"
+
+COMPATIBLE_HOSTS = "arm|aarch64|x86_64"
+
+FILES_${PN} = " \
+    ${bindir}/mender-gateway \
+"
+
+FILES_${PN}_append_mender-systemd = " \
+    ${systemd_system_unitdir}/mender-gateway.service \
+"
+
+S = "${WORKDIR}/mender-gateway-${PV}"
+
+do_version_check() {
+    if ! ${@'true' if d.getVar('MENDER_DEVMODE') else 'false'}; then
+        if ! strings ${WORKDIR}/${SUB_FOLDER}/mender-gateway | fgrep -q "${PN} ${PV}"; then
+            bbfatal "String '${PN} ${PV}' not found in binary. Is it the correct version? Check with --version. Possible candidates: $(strings ${WORKDIR}/${SUB_FOLDER}/mender-gateway | grep '${PN} [a-f0-9]')"
+        fi
+    fi
+}
+addtask do_version_check after do_unpack before do_install
+
+SYSTEMD_SERVICE_${PN}_mender-systemd = "mender-gateway.service"
+
+do_install() {
+    install -d -m 755 ${D}${bindir}
+    install -m 755 ${S}/${SUB_FOLDER}/mender-gateway ${D}${bindir}/mender-gateway
+
+    # install config file if provided
+    if [ -f ${WORKDIR}/mender-gateway.conf ]; then
+        install -d -m 755 ${D}/${sysconfdir}/mender
+        install -m 0600 ${WORKDIR}/mender-gateway.conf ${D}/${sysconfdir}/mender/mender-gateway.conf
+    else
+        bbwarn "No mender-gateway.conf found in SRC_URI, no mender-gateway.conf installed. The gateway will run with the default parameters"
+    fi
+}
+
+do_install_append_mender-systemd() {
+    install -d -m 755 ${D}${systemd_unitdir}/system/
+    install -m 644 ${S}/support/mender-gateway.service ${D}${systemd_unitdir}/system/mender-gateway.service
+}
--- a/meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway_git.bb
+++ b/meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway_git.bb
@@ -1,0 +1,19 @@
+require mender-gateway.inc
+
+inherit mender-closed-source-utils
+
+SRCREV = "${@mender_closed_source_srcrev_from_src_uri(d, '${SRC_URI}', 'mender-gateway')}"
+
+PV = "${@mender_closed_source_pv_from_preferred_version(d, '${SRCREV}')}"
+
+def tarball_directory_from_pv(d, pv):
+    return pv.split("master-git+")[-1]
+
+# Define S to work both on git sha and "master" tarballs
+S = "${WORKDIR}/mender-gateway-${@tarball_directory_from_pv(d, '${PV}')}"
+
+# Skip version check
+MENDER_DEVMODE = "true"
+
+# Downprioritize this recipe in version selections.
+DEFAULT_PREFERENCE = "-1"
--- a/meta-mender-commercial/classes/mender-closed-source-utils.bbclass
+++ b/meta-mender-commercial/classes/mender-closed-source-utils.bbclass
@@ -1,0 +1,55 @@
+
+def mender_closed_source_srcrev_from_src_uri(d, src_uri, repo_name):
+    pref_version = d.getVar("PREFERRED_VERSION")
+    if pref_version is None or pref_version == "":
+        pref_version = d.getVar("PREFERRED_VERSION_%s" % d.getVar('PN'))
+    if pref_version and pref_version.find("-build") >= 0:
+        # If "-build" is in the version, SRCREV won't be used for defining PV
+        return ""
+    else:
+        # Extract SRCREV from the tarball filename
+        import glob
+        import re
+        # Remove "file://" prefix
+        src_uri_list = src_uri.split()
+        if len(src_uri_list) == 0:
+            # No source specified. We won't be building this component.
+            return ""
+        src_uri_glob = src_uri_list[0][len("file://"):]
+        # Get the filename
+        filenames = glob.glob(src_uri_glob)
+        if len(filenames) == 0:
+            bb.error("Failed to find %s in path %s" % (src_uri_glob, repo_name))
+            bb.error("Please make sure SRC_URI_pn-%s is pointing to the downloaded tarball " % repo_name)
+        elif len(filenames) != 1:
+            bb.error("Expected exactly one file, found: %s" % filenames)
+        filename = os.path.basename(filenames[0])
+        # Now extract the version from the filename
+        m = re.match(repo_name + r"-master\.tar\.(?:xz|gz)", filename)
+        if m:
+            # Building from external tarball, do not append git info
+            return ""
+        m = re.match(repo_name + r"-([0-9]+\.[0-9]+\.[0-9]+(?:-build[0-9]+)?)\.tar\.(?:xz|gz)", filename)
+        if m:
+            # The tarball is a tagged version, in which case SRCREV is not to
+            # be used but still needs to parse
+            return ""
+        m = re.match(repo_name + r"-([0-9a-f]{7,})\.tar\.(?:xz|gz)", filename)
+        if m:
+            # Building from internal tarball, append git info
+            return "-git+" + m.group(1)
+        # At this point the version is unknown.
+        bb.fatal("Unknown version. Failed to parse %s" % filename)
+
+def mender_closed_source_pv_from_preferred_version(d, srcrev):
+    pref_version = d.getVar("PREFERRED_VERSION")
+    if pref_version is None or pref_version == "":
+        pref_version = d.getVar("PREFERRED_VERSION_%s" % d.getVar('PN'))
+    if pref_version is not None and pref_version.find("-build") >= 0:
+        # If "-build" is in the version, use the version as is. This means that
+        # we can build tags with "-build" in them from this recipe, but not
+        # final tags, which will need their own recipe.
+        return pref_version
+    else:
+        # Else return "master${SRCREV}".
+        return "master%s" % srcrev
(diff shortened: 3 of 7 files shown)


lluiscampos:
These four go together:
* https://github.com/mendersoftware/integration/pull/1932
* https://github.com/mendersoftware/mender-gateway/pull/13
* https://github.com/mendersoftware/meta-mender/pull/1599
* https://github.com/mendersoftware/mender-qa/pull/564

meta-mender-commercial/classes/mender-closed-source-tar-utils.bbclass
```diff
@@ -0,0 +1,40 @@
+
+def mender_closed_source_srcrev_from_src_uri(d, src_uri, repo_name):
+    pref_version = d.getVar("PREFERRED_VERSION")
```
kacf, line 3: When `PREFERRED_VERSION_pn-mender-client` is used, this value gets put into `PREFERRED_VERSION` when the recipe currently being parsed is the mender-client (`"${PN}" == "mender-client"`). However, the preferred style in Yocto is actually `PREFERRED_VERSION_mender-client`, which is not a special variable, and which is available only under that name. See example [here](https://www.yoctoproject.org/docs/latest/bitbake-user-manual/bitbake-user-manual.html#var-bb-PREFERRED_VERSION). The fact that the `pn-` version works is a combination of history and coincidence, I believe.

Therefore, if `PREFERRED_VERSION` is empty, you should also check `PREFERRED_VERSION_${PN}`, since this is the officially supported variant.

I know we are probably not consistent about this everywhere (and definitely not in mender-qa, which I just reviewed), but better to be prepared.

**Update:** I just realized that this is pre-existing, but I encourage fixing it anyway.

meta-mender-commercial/classes/mender-closed-source-tar-utils.bbclass
```diff
@@ -0,0 +1,40 @@
...
+            # The tarball is a tagged version, in which case SRCREV is not to
+            # be used but still needs to parse
+            return ""
+        m = re.match(repo_name + r"-([0-9a-f]{7,})\.tar\.(?:xz|gz)", filename)
+        if m is not None:
+            # Building from internal tarball, append git info
+            return "-git+" + m.group(1)
+        # At this point the version is unknown.
+        bb.fatal("Unknown version. Failed to parse %s" % filename)
+        return ""
```
kacf, line 40: Nitpick: `bb.fatal` is, well, fatal, so `return` is superfluous.

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
```diff
@@ -0,0 +1,49 @@
...
+SYSTEMD_SERVICE_${PN}_mender-systemd = "mender-gateway.service"
+
+do_install() {
+    install -d -m 755 ${D}${bindir}
+    install -m 755 ${S}/${SUB_FOLDER}/mender-gateway ${D}${bindir}/mender-gateway
+
+    # install config file if provided
+    if [ -f ${WORKDIR}/mender-gateway.conf ]; then
+        mkdir -p  ${D}/${sysconfdir}/mender
+        install -m 0600 ${B}/mender-gateway.conf ${D}/${sysconfdir}/mender/mender-gateway.conf
```
kacf, line 40: You are not installing the same file which you checked (except if they happen to be identical). Intentional?

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway_git.bb
```diff
@@ -0,0 +1,27 @@
+require mender-gateway.inc
+
+inherit mender-closed-source-tar-utils
+
+SRCREV = "${@mender_closed_source_srcrev_from_src_uri(d, '${SRC_URI}', 'mender-gateway')}"
+
+def mender_gateway_version_from_preferred_version(d, srcrev):
+    pref_version = d.getVar("PREFERRED_VERSION")
```
kacf, line 8: Same here with regards to `PREFERRED_VERSION`.

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway_git.bb
```diff
@@ -0,0 +1,27 @@
...
+        # final tags, which will need their own recipe.
+        return pref_version
+    else:
+        # Else return "master${SRCREV}".
+        return "master%s" % srcrev
+
+PV = "${@mender_gateway_version_from_preferred_version(d, '${SRCREV}')}"
+
+# Override S with a wildcard
+S = "${WORKDIR}/mender-gateway-*"
```
kacf, line 21: Woa! This one feels a bit risky... Doesn't the wildcard match `${PV}`? That's usually how it's defined...

kacf (approved):
Pre-approving, but please check the comments. Some improvements can still be made.

meta-mender-commercial/classes/mender-closed-source-tar-utils.bbclass
```diff
@@ -0,0 +1,40 @@
+
+def mender_closed_source_srcrev_from_src_uri(d, src_uri, repo_name):
+    pref_version = d.getVar("PREFERRED_VERSION")
+    if pref_version is not None and pref_version.find("-build") >= 0:
```
oleorhagen, line 4: Question:
Why not just `if prev_version and...`

meta-mender-commercial/classes/mender-closed-source-tar-utils.bbclass
```diff
@@ -0,0 +1,40 @@
...
+        # Remove "file://" prefix
+        src_uri_list = src_uri.split()
+        if len(src_uri_list) == 0:
+            # No source specified. We won't be building this component.
+            return ""
+        src_uri_glob = src_uri_list[0][len("file://"):]
+        # Get the filename
+        filenames = glob.glob(src_uri_glob)
+        if len(filenames) == 0:
+            bb.error("Failed to find %s on path %s" % (src_uri_glob, repo_name))
```
oleorhagen, line 20: nit:
`in path..`

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
```diff
@@ -0,0 +1,49 @@
...
+FILES_${PN}_append_mender-systemd = " \
+    ${systemd_system_unitdir}/mender-gateway.service \
+"
+
+S = "${WORKDIR}/mender-gateway-${PV}"
+
+do_version_check() {
+    if ! ${@'true' if d.getVar('MENDER_DEVMODE') else 'false'}; then
+        if ! strings ${WORKDIR}/${SUB_FOLDER}/mender-gateway | fgrep -q "${PN} ${PV}"; then
+            bbfatal "String '${PN} ${PV}' not found in binary. Is it the correct version? Check with --version. Possible candidates: $(strings ${WORKDIR}/${SUB_FOLDER}/mender-binary-delta | grep '${PN} [a-f0-9]')"
```
oleorhagen, line 25: Copy-pasta error: `mender-binary-delta`

oleorhagen: Which means we could probably add this to a `bbclass` and parametrize it, right?

oleorhagen: Also, I was about to suggest to use `nm` to check for the symbol, instead of extracting all the strings from the binary.

Which made me realize that the symbol we're trying to set for the version:
`github.com/mendersoftware/mender-gateway/config.Version` does not exist in the binary...

I will add a PR to fix this up :) 

But the suggestion still persides

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
```diff
@@ -0,0 +1,49 @@
...
+
+SYSTEMD_SERVICE_${PN}_mender-systemd = "mender-gateway.service"
+
+do_install() {
+    install -d -m 755 ${D}${bindir}
+    install -m 755 ${S}/${SUB_FOLDER}/mender-gateway ${D}${bindir}/mender-gateway
+
+    # install config file if provided
+    if [ -f ${WORKDIR}/mender-gateway.conf ]; then
+        mkdir -p  ${D}/${sysconfdir}/mender
```
oleorhagen, line 39: Q:
Why `mkdir` here? Don't we usually do `install -d` ? (No matter really, just curious)

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
```diff
@@ -0,0 +1,49 @@
...
+        install -m 0600 ${B}/mender-gateway.conf ${D}/${sysconfdir}/mender/mender-gateway.conf
+    else
+        bbwarning "No mender-gateway.conf found in SRC_URI, not configuring mender-gateway"
```
oleorhagen, line 42: Suggestion:
`, no mender-gateway.conf installed. The gateway will run with only the default parameters when started up`, or something similar?

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
```diff
@@ -0,0 +1,49 @@
...
+    fi
+}
+
+do_install_append_mender-systemd() {
```
oleorhagen, line 46: Q:
Doesn't the `systemd` inherited recipe do this automatically when the file is given?

oleorhagen (approved):
Lookin' good

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
lluiscampos: > Which made me realize that the symbol we're trying to set for the version:
github.com/mendersoftware/mender-gateway/config.Version does not exist in the binary...
>
> I will add a PR to fix this up :)

Alf is taking care of this, see acceptance criteria at https://tracker.mender.io/browse/MEN-5435

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway_git.bb
lluiscampos: @kacf This seems to be the problem in CI. It works on my machine...

`PV` is not really reliable to use here, at least not with the current design. `PV` can be `master-git+86e78e1` or `master`. And for both cases directory inside the tarball is `mender-gateway-86e78e1`.

Can we have a short f2f to discuss options?

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
lluiscampos: > Copy-pasta error: mender-binary-delta

Thanks

> Which means we could probably add this to a bbclass and parametrize it, right?

Maybe. But it is a bit different in `mender-monitor` than in `mender-binary-delta`. I think I'll leave it for the future...

> Also, I was about to suggest to use nm to check for the symbol, instead of extracting all the strings from the binary

As we don't really have it in the binary as per today I cannot experiment. I think I'll leave it as it is that is already proven working for `mender-binary-delta`.

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
lluiscampos: Yes, you are right. I am changing it.

I copied it from [here](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/recipes-mender/mender-connect/mender-connect.inc#L67) though.

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway_git.bb
lluiscampos: Resolved f2f, the `master` package is now modified when copying it and here `S` becomes deterministic.

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
lluiscampos: Mistake, thanks.

meta-mender-commercial/classes/mender-closed-source-tar-utils.bbclass
lluiscampos: Removed also the `is not None` from the expressions below.

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
lluiscampos: Ok

lluiscampos: Although this I might rework it when adding the demo certs to `meta-mender-demo`.

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
lluiscampos: I don't know exactly how it works with `systemd` to be honest. Do I then need to define `SYSTEMD_SERVICE_${PN}_mender-systemd`? I'll investigate a bit :)

meta-mender-commercial/classes/mender-closed-source-tar-utils.bbclass
lluiscampos: Got it. I remember this :smiley: 

Thanks. I fixed it now.

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway_git.bb
lluiscampos: I fixed this one too and moved the function to the shared class, so that the fix gets into `mender-monitor` too.

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
oleorhagen: Yeh, seems ur right :+1:

lluiscampos:
Pipeline [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/475533320).

PR #1606, opened 2022-02-24
mender-gateway: Fix do_version_check

lluiscampos:
Two fixes:
* The source file was incorrect (use S instead of WORKDIR)
* The string "mender-gateway <PV>" is not present on the go binary, look
  look for only the version instead.

### DIFF
--- a/meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
+++ b/meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
@@ -21,8 +21,8 @@ S = "${WORKDIR}/mender-gateway-${PV}"
 
 do_version_check() {
     if ! ${@'true' if d.getVar('MENDER_DEVMODE') else 'false'}; then
-        if ! strings ${WORKDIR}/${SUB_FOLDER}/mender-gateway | fgrep -q "${PN} ${PV}"; then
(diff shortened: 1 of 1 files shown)


lluiscampos:
@oleorhagen  I tried using `nm` as suggested [here](https://github.com/mendersoftware/meta-mender/pull/1599#discussion_r807905662). But although i mange to find the correct symbol `main.Version` I couldn't find a way to get its value.

lluiscampos:

oleorhagen (approved):
The version is not present _yet_, right?

lluiscampos:
> The version is not present _yet_, right?

It is [implemented](https://github.com/mendersoftware/mender-gateway/blob/d0160fe6ca08856c4743bc59e13f7aaf1a6bf73e/main.go#L46). Although we have no tags so it only produces Git sha based strings.

I locally tested and experimented with this by creating a local tag, then `make dist-package`, then create a recipe in meta-mender... and then finally debug `do_version_check` :wink:

oleorhagen:
> > The version is not present _yet_, right?
> 
> It is [implemented](https://github.com/mendersoftware/mender-gateway/blob/d0160fe6ca08856c4743bc59e13f7aaf1a6bf73e/main.go#L46). Although we have no tags so it only produces Git sha based strings.
> 
> I locally tested and experimented with this by creating a local tag, then `make dist-package`, then create a recipe in meta-mender... and then finally debug `do_version_check` wink

Solid! :smile:

PR #1624, opened 2022-03-10
Dunfell update

kacf:


TheYoctoJester (commented):
Concerning 5e9e15706a2812c25c8d749cacaa39a6caddaf0e

Whats the rationale for this particular commit, respectively how did it get in? I trust Leon to know what he's doing, but the reasoning is not clear to me.

TheYoctoJester (commented):
I would like to push an additional note concerning the compatibility section in the hosted docs. Can the merge be postponed for a couple of hours and incorporate this then?

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
```diff
@@ -0,0 +1,49 @@
...
+            bbfatal "String '${PN} ${PV}' not found in binary. Is it the correct version? Check with --version. Possible candidates: $(strings ${WORKDIR}/${SUB_FOLDER}/mender-gateway | grep '${PN} [a-f0-9]')"
+        fi
+    fi
+}
+addtask do_version_check after do_unpack before do_install
+
+SYSTEMD_SERVICE_${PN}_mender-systemd = "mender-gateway.service"
+
+do_install() {
+    install -d -m 755 ${D}${bindir}
```
TheYoctoJester, line 38: I think that is unnecessary?

lluiscampos (approved):

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
```diff
@@ -0,0 +1,49 @@
+LICENSE = "CLOSED"
```
TheYoctoJester, line 1: While this works, it is technically wrong. `LICENSE="CLOSED"` means to Yocto: "don't care about this". It is a common misconception that it means "closed/commerial source". This probably does not mandate a real change, but in the future we should add and ship the real license that this is under.

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
lluiscampos: Maybe unnecessary, but is it harmless? I think it is a good practice to install the dir before the files. We do it in most of our recipes.

TheYoctoJester: Certainly not harmful, just an observation.

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
lluiscampos: Interesting, we didn't know about this. I created [MEN-5517](https://tracker.mender.io/browse/MEN-5517) to follow-up.

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
kacf: AFAIK, this is necessary in all recipes. Directories in `${D}` do not generally pre-exist.

Even if there might be an exception for `bindir` (I haven't checked) I think we should keep it for consistency.

meta-mender-commercial/recipes-mender/mender-gateway/mender-gateway.inc
kacf: Interesting, I did not know that. In any case, we need to direct such improvements to master. This PR is more about making sure the cherry-picks from master into dunfell are sensible. If we fix such things on master, we will also fix them on dunfell with subsequent cherry-picks.

PR #1632, opened 2022-03-21
Kirkstone fixes.

kacf:
This is easier since the build script is cross branch, so it would
have to handle the dunfell/kirkstone syntax difference.

Changelog: None


TheYoctoJester:
Looks good to me!

PR #1643, opened 2022-03-24
Raspberrypi honister fixes

nandra:


kacf (approved):
Thanks!

nandra:
One more thing which is not working is that CMDLINE variable cannot be updated in linux-raspberrypi (at least from my findings -> removal of root=/dev/mmcblk0p2 and replace with root=${kernel_mender_root})
meta-raspberrypi have recipe called rpi-cmdline which process CMDLINE variable but I got error (bad substitution) when just copy actual linux-raspberrypi_%.bbappend to rpi-cmdline_%.bbappend. So I came up with:
```
CMDLINE_ROOTFS:remove = "root=/dev/mmcblk0p2"

# simple adding CMDLINE_ROOTFS:append = " root=\${mender_kernel_root}"
# doesn't work as we get Bad substitution error
do_install:append () {
        if ! grep -r 'mender_kernel_root' "${WORKDIR}/cmdline.txt"; then
                sed -i '$ s/$/ root=\${mender_kernel_root}/' ${WORKDIR}/cmdline.txt
        fi
}

```
Maybe there exists better idea? Thanks

kacf:
Indeed somewhat hacky. I don't have a better idea off the top of my head though. I'll eventually take a look at this myself, once I get closer to finishing kirkstone support. Maybe I'll have a bright idea then. 🙂

PR #1644, opened 2022-03-24
Tests: add messages to known spurious failures

merlin-northern:
Just an idea, we could add some extra messages for spurious failures.
Makes it easy for BMs. Probably could have been done better, maybe some pytest
assert override, but I am not that smart.
Opinions?

ChangeLog: none

### DIFF
--- a/tests/acceptance/test_update_control.py
+++ b/tests/acceptance/test_update_control.py
@@ -127,7 +127,7 @@ def make_and_deploy_artifact(
         )
 
 
-def wait_for_state(connection, state_to_wait_for):
+def wait_for_state(connection, state_to_wait_for, message=None):
     log = []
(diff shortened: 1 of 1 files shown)


kacf:
Not sure what @lluiscampos thinks about this, but I think it's better to handle spurious failures by looking at recent failures, not by having it in the code. Then it sort of becomes permanent, which is not the goal.

lluiscampos:
I agree with Kristian, we should have this kind of information in our tests results aggregator (https://qastatus.mender.io/) or similar site, not in the code.

kacf:
Consensus rules then! 🙂

merlin-northern:
> Consensus rules then! 🙂

okey

PR #1076, opened 2020-09-04
Remove e2fsprogs patch to fix bmaptool issue.

hacpa:
This patch, which removes  metadata_csum from the available ext4 features, breaks flashing with bmaptool.
It was created at a time when tools such as fsck.ext4 and tune2fs did not handle that feature.
Nowadays, these tools have been updated and it is safe to remove the patch.

Changelog: Title

Signed-off-by: Pascal Hache <hacpa@touchtunes.com>

# External Contributor Checklist

🚨 Please review the [guidelines for contributing](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md) to this repository.

- [x] Make sure that all commits have a [`Changelog`](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#changelog-tags) tag. If nothing should be added to the Changelog, add a `Changelog: None` tag. If there is a change, add `Changelog: Commit`, or `Changelog: Title`, depending on what should be included in the changelog.

- [x] Make sure that all commits are signed with [`git --signoff`](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work). Also note that the signoff author must match the author of the commit.

### Description

Please describe your pull request.

Thank you!

PR #1645, opened 2022-03-24
Refresh e2fsprog patch and mark layer compatible with hardknott

clementperon:
Hi,

just a quick patch refresh for e2fsprog and mark the layer compatible with hardknott

### DIFF
--- a/meta-mender-core/recipes-core/systemd/systemd-boot_%.bbappend
+++ b/meta-mender-core/recipes-core/systemd/systemd-boot_%.bbappend
@@ -1,3 +1,3 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
 
-SRC_URI += "file://systemd-boot-slotconfig.patch"
+SRC_URI:mender-systemd-boot += "file://systemd-boot-slotconfig.patch"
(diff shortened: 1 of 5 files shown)


kacf (approved):
Thanks!

clementperon:
@kacf I think it was a mistake to mark these layer compatible with dunfell / hardknott.

If I understand correctly, now that new override syntax for Kirkstone / Honnister has been merged it's no more compatible with dunfell / hardknott. What do you think?

I can sent a patch to remove all dunfell / Hardknott compatibility

TheYoctoJester:
@clementperon that is not completely necessary, `dunfell` gained backported syntax compatibility. `hardknott` should be removed, though.

clementperon:
@TheYoctoJester, After searching for this I found :
```
The new syntax will work on the dunfell, gatesgarth, hardknott, master and new 
release branches going forward. That applies now and that support will be in the
next point releases for dunfell (3.1.11) and hardknott (3.3.3).
```

https://lists.yoctoproject.org/g/yocto/message/54368

kacf:
Good!

Note that eventually all except kirkstone will be removed though, because we only support LTS branches in Mender. What we have now is temporary until we've released it.

nandra:
@kacf  I'm using honister and this patch doesn't apply properly.

clementperon:
@nandra sorry for that, are you talking about the e2fsprog ? 
I have indeed tested it on hardknott without checking honnister support.

I will check and resend to make it compatible with honnister.

nandra:
@clementperon yes about this merge request. Thanks

kacf:
> I'm using honister and this patch doesn't apply properly.

I have removed the patch in my latest local branch, I suspect it's not needed anymore. I'll push it soon, after I've cleaned it up.

clementperon:
@kacf  found this thread trying to remove the patch https://github.com/mendersoftware/meta-mender/pull/1076

kacf:
Very good find, @clementperon! I had forgotten about that. Looks like we need it then.

I'm currently based on what will become kirkstone (poky/master, currently at 3b8d67d4f0c05), so I'll refresh the patch relative to that.

nandra:
@kacf when do you please expect to have this change merged? Thanks.

kacf:
@nandra: You can already see it [here](https://github.com/mendersoftware/meta-mender/pull/1632) together with a lot of other fixes. I'm still having some problems compiling the client though (related to [this](https://github.com/mendersoftware/openssl/pull/37), which doesn't quite work yet), so it's still WIP.

PR #1658, opened 2022-04-07
More kirkstone fixes

kacf:
A lot of these commits are from merging with master. The ones that are actually new, are these:

* 93639f02 Adapt to new kernel which puts QEMU storage at sda, not hda.
* 7dc79fac Fix unstable build checksum because of override.
* ec141f44 Fix unstable build checksum by removing `BB_CURRENTTASK`.
* 72abd1d2 Remove now incorrect comment, should have been removed in f9e734acf5a.
* 5f8c58a2 Fix two overlooked syntax conversions.


TheYoctoJester (approved):
The bad "gut feeling" concerning the override expansion has already been discussed on Slack. As I don't have a better suggestion for now, okay with me.

kacf:
@TheYoctoJester, @oleorhagen: I'm still nowhere close to getting all this passing, but maybe we could just review this, and get it out of the way. It's only merging into `master-next` anyway, so it's not production grade.

The things that need reviewing is the stuff from "fix: Fix broken dependencies when building mender-client-native" and later. The rest is either already reviewed or merged in from the master branch.

PR #1674, opened 2022-04-26
test: moving the logic of clean images into extract_fs

merlin-northern:


merlin-northern:
@mender-test-bot --pr integration/1995 please

meta-mender-qemu/scripts/docker/extract_fs
```diff
@@ -23,10 +23,16 @@ IMG_PREFIX=${1}
...
 [[ $IMG == *gz ]] && gzip -d "$IMG"
 IMG=$(ls /${IMG_PREFIX}*img | head -n 1)
-OUTPUT=/output/$(sed -e 's/\.[^.]*img$/.ext4/' <<<$(basename $IMG))
+
+# if we are given the prefix, we need to remove it.
+CLEAN_FLAGS="-e s/^${IMG_PREFIX}//"
+
+# if the prefix we did not get, we need to remove the default prefix
+[[ "${IMG_PREFIX}" == "" ]] && CLEAN_FLAGS="-e s/^clean-//"
+OUTPUT=/output/$(sed -e 's/\.[^.]*img$/.ext4/' "${CLEAN_FLAGS}" <<<$(basename $IMG))
```
kacf, line 32: Shouldn't `"${CLEAN_FLAGS}"` be unquoted here, if it contains more than one argument?

kacf (approved):
Looks good to me, just one question below.

Smart to move it here instead of in integration.

meta-mender-qemu/scripts/docker/extract_fs
merlin-northern: I quoted it for extra safety in case of `*` or `?` -- glob in general. believe it or not, but this magic somehow is supported by sed:
```bash
echo "here some1 and some2" | sed "-e s/some1/_/"
```
```
here _ and some2
```

while that one:

```bash
#echo "here some1 and some2" | sed "-e s/some1/_/ -e s/some2/--/"
```
```
sed: -e expression #1, char 13: unknown option to `s'
```

is not.

having said that, I modified a bit, what do you think now?

kacf: Interesting, so it has a special case for that. I suppose that it's equivalent to `-es/a/b/`, and it just ignores the extra space.

Looks good now!

merlin-northern: > Interesting, so it has a special case for that. I suppose that it's equivalent to `-es/a/b/`, and it just ignores the extra space.

that must be it! sed will never stop to amaze me in many ways, one day perhaps I will even understand some of it.

> Looks good now!

thank you.

PR #1696, opened 2022-05-24
Fixes for Pseudo abort issue

kacf:
```
commit f6c1a8cf55c34898909ff7a6ed577f0dc50920f0
Date:   Mon May 23 21:00:30 2022

    test: nodbus: Clean the mender-client between builds.
    
    We need it because tag changes ("nodbus") are not picked up without
    cleaning first.
    
commit 8b4eb998b9552203792de3ba9b66c027df64a64e
Date:   Mon May 23 16:46:35 2022

    fix: Force building in separate build directory, even with `externalsrc`.
    
    The problem with building in `${S}` is that files remain there after
    a build, due to bad interoperability between the `externalsrc` and
    `go` classes, the `clean` target does not clean this up. Therefore,
    force the creation of a separate build directory always, even if it's
    based on an `externalsrc` directory.
    
    This was always needed for coverage analysis, so just steal the code
    from there.
    
    Changelog: Title
    
```

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client.inc
@@ -129,6 +129,23 @@ _MENDER_TAGS = "${@bb.utils.contains('PACKAGECONFIG', 'dbus', '', 'nodbus', d)}"
 EXTRA_OEMAKE_append = " TAGS='${_MENDER_TAGS}'"
 GOPTESTBUILDFLAGS_append = " -tags '${_MENDER_TAGS}'"
 
+do_configure_prepend () {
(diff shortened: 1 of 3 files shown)


oleorhagen (approved):

kacf:
One failure in `test_configure` which I'm 95% sure is spurious. I'll merge an keep an eye on it.

oleorhagen:
> One failure in `test_configure` which I'm 95% sure is spurious. I'll merge an keep an eye on it.

For future reference, this is where we keep track of spurious failures: https://tracker.mender.io/browse/QA-276

kacf:
Yes, this one is quite rare I think. I have seen it before, but very seldomly. If it's several weeks between each one it's probably not worth investigating. If I see it again I will add it though.

PR #1701, opened 2022-05-31
chore: moved qemu docker image to python base image.

mzedel:
- since the python version installed was unversioned until now, the default target would be the latest stable
- to both prevent unintended versioning issues & speed up the build, the base image is switched to python
- also removed another image layer by avoiding the apk cache

Signed-off-by: Manuel Zedel <manuel.zedel@northern.tech>


mzedel:

PR #1706, opened 2022-06-03
test: Get rid of rngd which slows down QEMU considerably.

kacf:
(cherry picked from commit 2faafa7ac7a82b5156af9ddb37a91001d364c232)


oleorhagen (approved):

kacf:
Dammit, unfortunately there is a problem with this fix on vexpress-qemu-flash. I'll leave this alone for now, because I anyway have to retest this with kirkstone. I might be in a better position then to figure out what happens.

kacf:
@mzedel: Unfortunately it looks like this change has no effect on the performance on dunfell. I know for a fact that it helps on the upcoming kirkstone, but it may be that it simply brings it back to this base level, since dunfell simply isn't affected by it. I'm closing this, since kirkstone already has this patch.

PR #1707, opened 2022-06-07
More kirkstone fixes

kacf:


TheYoctoJester (approved):
Not sure I understand all possible implications, but nothing that seems to fishy or blocking for me here. As this is not production (yet), better merge small and often. :-)

kacf:
Thanks, @TheYoctoJester!

PR #1710, opened 2022-06-09
fix: Set the image size to 1027 MiB for vexpress-qemu

oleorhagen:
This is part of the migration to Ubuntu 22.04, which has a newer version of
qemu, in which the sd-card size has to be a multiple of two larger than the
running image (which as of writing is around 600MiB's).

Hence, simply set the default size for vexpress-qemu to 1024 MiB.

Changelog: None
Ticket: QA-390


oleorhagen:

kacf (approved):

PR #1715, opened 2022-06-13
Changes to support interoperability with kirkstone.

kacf:
Just cherry-picks, which have already been reviewed in the master-next branch.


kacf:
Deliberately ignoring commit check, because this is a cherry-pick of a commit from before the new syntax was introduced.

PR #1716, opened 2022-06-13
Dunfell: Changes to support interoperability with kirkstone.

kacf:
Just cherry-picks, but I did fork the mender-image-tests repository, see https://github.com/mendersoftware/mender-image-tests/pull/68.


kacf:
I'm deliberately ignoring the commit hook failures here, because these are cherry-picks and I don't want them to look different from the originals, for search purposes.

PR #1685, opened 2022-05-06
Simplify mender-client-resize-data-part and make it compatible with "partuuid" setups

ColinFinck:
This PR is split into two parts:
* **Simplify mender-client-resize-data-part by figuring out the storage device from `MENDER_DATA_PART`**
  This way, we don't need to populate the script with a valid value for `MENDER_STORAGE_DEVICE` anymore. That variable doesn't exist in a "partuuid" setup and is already auto-derived in a similar way in a "partlabel" setup.
* **Make mender-client-resize-data-part compatible with "partuuid" setups**
  This commit derives the expected value for `MENDER_DATA_PART` from the PARTUUID in "partuuid" setups, in a similar way as it's already done for "partlabel" setups.

I have backported this fix to dunfell (which only supports "partuuid" but not "partlabel") in https://github.com/enlyze/meta-mender/tree/dunfell-mender-client-resize-data-partuuid and successfully tested it.
I'd be grateful if you could apply both my PR to "master" as well as my backport to "dunfell".

Fixes [MEN-3701](https://tracker.mender.io/browse/MEN-3701)

PR #1720, opened 2022-06-14
dunfell: Simplify mender-client-resize-data-part and make it compatible with "partuuid" setups

ColinFinck:
Backport of #1685 to the dunfell branch.

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/files/mender-client-resize-data-part.sh.in
+++ b/meta-mender-core/recipes-mender/mender-client/files/mender-client-resize-data-part.sh.in
@@ -4,11 +4,11 @@ set -e
 
 # All sizes are in blocks of 512 bytes
 
-MENDER_STORAGE_DEVICE=@MENDER_STORAGE_DEVICE@
 MENDER_DATA_PART=@MENDER_DATA_PART@
 
-mender_storage_device_base=$(basename ${MENDER_STORAGE_DEVICE})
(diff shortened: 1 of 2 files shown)


kacf (changes requested):
Hey @ColinFinck, can you please use `git cherry-pick -x <SHA>` when backporting. They should be the same as the originals, just with the extra cherry-pick string added.

ColinFinck:
@kacf While dunfell suffers from the same mender-client-resize-data-part bug in a "partuuid" setup, it does not support the "partlabel" feature.
This is why an exact cherry-pick of the two commits from master with no code changes is not possible.

Shall I split this up into the original two commits with added cherry-pick string anyway, and just hide the necessary changes for dunfell in the diffs?
Or do you propose another method to make it clear that this is not an exact copy of the commits to master?

kacf:
I think the first method is preferred, as long as the patch at least resembles the original. If it's completely different then one can reevaluate, but solving conflicts and getting a slightly different patch is quite common during backporting. It's still useful that they look similar for tracking purposes.

ColinFinck:
@kacf Ok, cherry-picked both commits as suggested. Should be ready for review now.

kacf (approved):

kacf:
Retrying this pipeline in #1727.

kacf:
Retrying this pipeline in #1727.

PR #1721, opened 2022-06-14
test: Adapt test to cover all three GetJwtToken server addresses.

kacf:


oleorhagen (approved):

merlin-northern (commented):
ah yes, forgot about that one. thanks. (and sorry being late to the party)

PR #1725, opened 2022-06-16
More kirkstone fixes

kacf:

PR #1728, opened 2022-06-20
fix:Remove lmdb recipe, which meta-openembedded/kirkstone already has.

kacf:
Based on #1725, only the bottom commit is new here.


meta-mender-commercial/conf/layer.conf
```diff
@@ -13,7 +13,7 @@ BBFILE_PATTERN_mender-commercial = "^${LAYERDIR}/"
 BBFILE_PRIORITY_mender-commercial = "6"
 
 LAYERSERIES_COMPAT_mender-commercial = "dunfell kirkstone"
-LAYERDEPENDS_mender-commercial = "mender"
+LAYERDEPENDS_mender-commercial = "mender openembedded-layer"
```
kacf, line 16: @TheYoctoJester: Do you have an opinion on best practice here? I'm removing lmdb which exists in meta-openembedded/meta-oe. This is required only for mender-monitor. Is it then appropriate to have this dependency here? You could build without mender-monitor, and then you don't need it. But I can't think of a good way to express it conditionally.

kacf: Something like that maybe. But it feels a bit hacky, especially in this global section here.

PR #1731, opened 2022-06-22
feat: Replace dedicated mender.conf variables with MENDER_CONF.

kacf:
This uses the PACKAGECONFIG method of using var flags together with
a list (see example below).

I considered using this form:
```
MENDER_CONF[TenantToken] = "mytoken"
```
in other words using var flags directly to express JSON keys. But this
has several problems. First of all you cannot separate between numbers
and strings in the value. And second, it is not possible to use
overrides on var flag values. Hence I opted for the method below, with
generic flag names and full JSON instead.

I also opted to keep the rootfs and artifact verify key variables,
since they have side effects besides just being set in the
configuration file.

Ticket: MEN-4911

Changelog: These variables are no longer supported:
* `MENDER_INVENTORY_POLL_INTERVAL_SECONDS`
* `MENDER_RETRY_POLL_INTERVAL_SECONDS`
* `MENDER_SERVER_URL`
* `MENDER_TENANT_TOKEN`
* `MENDER_UPDATE_CONTROL_MAP_BOOT_EXPIRATION_TIME_SECONDS`
* `MENDER_UPDATE_CONTROL_MAP_EXPIRATION_TIME_SECONDS`
* `MENDER_UPDATE_POLL_INTERVAL_SECONDS`
To set these configuration values, one can now use the generic
`MENDER_CONF` mechanism instead. Here's an example of how it works:
```
MENDER_CONF[MyTenantToken] = '{"TenantToken":"mytoken"}'
MENDER_CONF += "MyTenantToken"
```
Alternatively, use a `mender-client_%.bbappend` file to include a
`mender.conf` file in `SRC_URI`.

BREAKING CHANGE


meta-mender-core/conf/mender-vars.json
```diff
@@ -96,6 +109,7 @@
     "MENDER_PARTITION_ALIGNMENT_DEFAULT": "",
     "MENDER_PARTITION_ALIGNMENT_KB": "",
     "MENDER_PARTITION_ALIGNMENT_KB_DEFAULT": "",
+    "MENDER_PERSISTENT_CONFIGURATION_VARS": "",
```
oleorhagen, line 110: haha, finally had to go digging in here again I take it!

.gitlab-ci.yml
```diff
@@ -30,7 +30,7 @@ trigger:mender-qa:
   rules:
     - if: '$CI_COMMIT_BRANCH == "master"'
 
-test:order
+test:order:
```
oleorhagen, line 33: Nit: Wrong commit

oleorhagen (approved):
I like it a lot!

Much much better imo

And `BREAKING CHANGE` even :)

kacf:
Rejecting this PR after discussions, see [this comment](https://tracker.mender.io/browse/MEN-4911?focusedCommentId=125091&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-125091).

PR #1734, opened 2022-06-28
MEN5689: fix bootfiles rename to match upstream

TheYoctoJester:
fix: adjust package rename of bootfiles to upstream in testing

Changelog:  None
Ticket: MEN5689

This adjusts the test scripts to the already existing adjustment of the payload scripts to upstream.


kacf (approved):
Hehe, that turned out simple!

PR #1750, opened 2022-07-11
fix: Race condition in mender-client-resize-data-part (dunfell)

ColinFinck:
Modifying the partition table in any way may trigger udev at some point.
We don't want to continue right while udev is recreating /dev/disk symlinks.
Instead, we want this to happen deterministically now and be done when the script is done.

Changelog: Title
Ticket: MEN-5717
Signed-off-by: Colin Finck <c.finck@enlyze.com>
(cherry picked from commit 6939aaae4f423c5b2aac58c88816972953847822)


ColinFinck:
CC @kacf

kacf:
Will need to re-cherry-pick this one then.

ColinFinck:
Done :)

kacf (approved):
Thanks!

PR #1751, opened 2022-07-14
feat:mender-binary-delta now takes tarball as input instead of folder.

kacf:
Ticket: MEN-4980

Changelog: The `mender-binary-delta` recipe now takes a tarball as
input instead of a folder containing unpacked files. Remove the line
starting with `FILESEXTRAPATHS`, and instead insert this line:
```
SRC_URI:pn-mender-binary-delta = "file://${HOME}/mender-binary-delta-1.4.0.tar.xz"
```
where the path points to the downloaded tarball.

BREAKING CHANGE

### DIFF
--- a/meta-mender-commercial/recipes-mender/mender-binary-delta/mender-binary-delta.inc
+++ b/meta-mender-commercial/recipes-mender/mender-binary-delta/mender-binary-delta.inc
@@ -5,10 +5,6 @@ SUB_FOLDER:arm = "arm"
 SUB_FOLDER:aarch64 = "aarch64"
 SUB_FOLDER:x86-64 = "x86_64"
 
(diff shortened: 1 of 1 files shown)


kacf:
This was way easier than I expected...

kacf:
No client developers around, so I've chosen a bit random list of reviewers...

kacf:
If reviewing this, it would be great if you could also look at my proposed change for Mender Hub: https://gist.github.com/kacf/0280bff9e9d8810335481deaf397cace

tranchitella (approved):
Looks good (both PR & Mender hub post).

TheYoctoJester (commented):
Looks good, without actually running it.

PR #1757, opened 2022-08-17
MEN-2585 feat: mender-bootstrap Artifact support

oleorhagen:
This commit adds a new recipe `mender-boostrap-artifact` which when included
will generate the bootstrap Artifact, and install it on the datadir in the image
generated.

The recipe is included for all newer recipe's of the Mender client, but removed
and replaced with the old `mender-artifact-info` recipe for the 3.3.0 client.

For older branches this will have to be ported to all older recipes.

Changelog: Add support for generating bootstrap Artifacts as a part of the move
to populating the client database on boot, as opposed to on the first update
install.
Ticket: MEN-2585

### DIFF
--- a/meta-mender-core/classes/mender-dataimg.bbclass
+++ b/meta-mender-core/classes/mender-dataimg.bbclass
@@ -2,6 +2,7 @@
 # filesystem.
 
 IMAGE_CMD_dataimg() {
+    generate_bootstrap_artifact
     if [ ${MENDER_DATA_PART_FSTYPE_TO_GEN} = "btrfs" ]; then
         force_flag="-f"
         root_dir_flag="-r ${IMAGE_ROOTFS}/data"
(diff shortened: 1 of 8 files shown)


lluiscampos:
Note: this PR should target feature branch.

meta-mender-core/classes/mender-dataimg.bbclass
```diff
@@ -35,6 +35,78 @@ IMAGE_CMD_dataimg_mender-image-ubi() {
     install -m 0644 "${WORKDIR}/data.ubifs" "${IMGDEPLOYDIR}/${IMAGE_NAME}.dataimg"
 }
 
+do_generate_bootstrap_artifact() {
```
oleorhagen, line 38: Straight copy-pasta from the ArtifactImg generation.

I think it's okay, but can unify this (somewhow), if you guys feel it's needed.

meta-mender-core/classes/mender-dataimg.bbclass
```diff
@@ -35,6 +35,78 @@ IMAGE_CMD_dataimg_mender-image-ubi() {
...
+        --artifact-name ${MENDER_ARTIFACT_NAME} \
+        --provides "rootfs-image.version:${MENDER_ARTIFACT_NAME}" \
+        --provides "rootfs-image.checksum:${img_checksum}" \
+        $extra_args \
+        ${MENDER_ARTIFACT_EXTRA_ARGS} \
+        --output-path "${IMAGE_ROOTFS}/data/mender/bootstrap.mender"
+
+}
+
+# In order to generate the bootstrap Artifact the rootfs image needs to be
```
oleorhagen, line 101: Embarassed to say, but the below three lines took me two days (and @kacf 15 minutes to spit out) :shamrock:

meta-mender-core/recipes-mender/mender-client/mender-client.inc
```diff
@@ -111,7 +111,7 @@ PACKAGECONFIG_append = "${@bb.utils.contains('MENDER_FEATURES', 'mender-client-i
 PACKAGECONFIG_append = "${@bb.utils.contains('MENDER_FEATURES', 'mender-uboot', ' u-boot', '', d)}"
 PACKAGECONFIG_append = "${@bb.utils.contains('MENDER_FEATURES', 'mender-grub', ' grub', '', d)}"
 
-PACKAGECONFIG[mender-client-install] = ",,,mender-artifact-info ca-certificates"
+PACKAGECONFIG[mender-client-install] = ",,,ca-certificates"
```
oleorhagen, line 114: I opted for removing the standard packageconfig, and having an explicit `RDEPENDS` in the old client recipe.

Not sure whether this is _the_ way or not...

tests/acceptance/test_bootstrap_artifact.py
```diff
@@ -0,0 +1,130 @@
...
+
+from utils.common import (
+    build_image,
+    signing_key,
+    reset_build_conf,
+)
+
+
+@pytest.fixture(scope="function")
+def build_image_with_signed_bootstrap_artifact(
```
oleorhagen, line 31: Also copy-pasta from the `mender-image-test` fixtures, with one notable difference.

These are `function` scoped, and they provide a different set of build-variables.

Feel it's okay, since it's only used here locally.

Otherwise, we still have to duplicate the functions somewhere, as I found no good way of _borrowing_ a fixture, and change it's scope... But feel free to take me to school here.

meta-mender-core/recipes-mender/mender-client/mender-client.inc
kacf: I believe it is for the extremely theoretical case where someone wants to install the Mender client, but without all the supporting stuff around it (which is enabled by the `mender-client-install` packageconfig). But honestly it was never well defined, nor documented anywhere, so no problem to remove it.

tests/acceptance/test_bootstrap_artifact.py
kacf: After f2f: Sort it first and comment on it in here.

kacf (approved):

oleorhagen:
```
Build started.
Link: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/623741584
```

oleorhagen:
https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/625277729

meta-mender-core/classes/mender-dataimg.bbclass
```diff
@@ -31,10 +32,85 @@ IMAGE_CMD_dataimg() {
...
+    img_checksum="$(sha256sum --binary ${IMGDEPLOYDIR}/${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.${ARTIFACTIMG_FSTYPE} | cut --delimiter=' ' --fields=1)"
+
+    if [ -z "${img_checksum}" ]; then
+        bberror "The image checksum cannot be empty"
+    fi
+
+    mender-artifact write bootstrap-artifact \
+        --artifact-name ${MENDER_ARTIFACT_NAME} \
+        --provides "rootfs-image.version:${MENDER_ARTIFACT_NAME}" \
+        --provides "rootfs-image.checksum:${img_checksum}" \
```
lluiscampos, line 100: Missing:

```
--clears-provides "rootfs-image.*"
```

lluiscampos (changes requested):
Looks good otherwise :+1: Nice test and nice ninja pytest hack!

meta-mender-core/classes/mender-dataimg.bbclass
kacf: It's probably smart to include it, yes. When the bootstrap artifact is used as we intend it to, then this flag is guaranteed to be redundant, because the database must be empty for this artifact to be installed. But it is possible for someone to take the same artifact and use it for a manual installation, in which case it would be relevant.

If we add it here, then we probably should in mender-convert as well.

lluiscampos: As I see it, it is the same as with signing the Artifact. It is redundant for the intended use, but we keep it for other use cases like you describe.

> If we add it here, then we probably should in mender-convert as well.

I agree.

I am sorry @oleorhagen that I didn't catch up this in the other PR :crying_cat_face:

oleorhagen: Why, we haven't released squat, it's still even in feature branches :)

We can do whatever we want :muscle:

oleorhagen: Fixing btw

oleorhagen: convert-pr: https://github.com/mendersoftware/mender-convert/pull/471

PR #1759, opened 2022-08-24
chore: Correct syntax on build-conf files

lluiscampos:
Gets the rid of:
```
WARNING: IMAGE_FSTYPES:remove += is not a recommended operator combination, please replace it.
```


kacf (approved):
This is not related to the RPi3 failure, is it?

lluiscampos:
> This is not related to the RPi3 failure, is it?

I don't think so. It is just something I found while building the images.

PR #1762, opened 2022-08-29
Dunfell update

kacf:


kacf:
The failures are known and due to other problems. The pipeline which matters is [this one](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/626199315), with one spurious failure for vexpress-qemu-flash.

PR #1765, opened 2022-08-31
MEN-5834: fix: cmdline.txt generation in meta-mender-raspberrypi

lluiscampos:
fix: cmdline.txt generation in meta-mender-raspberrypi
    
    The generation of cmdline.txt has been moved upstream to a new recipe,
    so our bbappend was not being applied.
    
    See:
    https://github.com/agherzan/meta-raspberrypi/commit/8827040d9c218f68ae3ee6a5814929888ed61581
    
    Ticket: MEN-5834
    Changelog: None

    fix(rpi-cmdline-mender.inc): Fix cmdline.txt after generation
    
    Due to an implementation detail, we cannot use the standard syntax for
    appending, see inline comment.
    
    Ticket: MEN-5834
    Changelog: None

### DIFF
--- a/meta-mender-raspberrypi/recipes-bsp/bootfiles/rpi-cmdline-mender.inc
+++ b/meta-mender-raspberrypi/recipes-bsp/bootfiles/rpi-cmdline-mender.inc
@@ -1,0 +1,2 @@
+CMDLINE_ROOTFS:remove = "root=/dev/mmcblk0p2"
+CMDLINE_ROOTFS:append = " root=\${mender_kernel_root} "
--- a/meta-mender-raspberrypi/recipes-bsp/bootfiles/rpi-cmdline-mender.inc
+++ b/meta-mender-raspberrypi/recipes-bsp/bootfiles/rpi-cmdline-mender.inc
@@ -1,2 +1,12 @@
 CMDLINE_ROOTFS:remove = "root=/dev/mmcblk0p2"
-CMDLINE_ROOTFS:append = " root=\${mender_kernel_root} "
+
+do_compile:append() {
+    # Modify cmdline.txt after its creation to append root=... parameter
+    #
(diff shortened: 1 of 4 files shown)


lluiscampos:
@kacf @TheYoctoJester The second commit is a workaround because it is not possible (or I have no idea how to do it!) to escape `${...}` in an string that is later processed as `"${@...'${CMDLINE}'..}"`. If you find an alternative way I will be happy to drop the workaround :)

meta-mender-raspberrypi/recipes-bsp/bootfiles/rpi-cmdline-mender.inc
```diff
@@ -1,2 +1,12 @@
 CMDLINE_ROOTFS:remove = "root=/dev/mmcblk0p2"
-CMDLINE_ROOTFS:append = " root=\${mender_kernel_root} "
```
kacf: I couldn't find out how to escape either. I'm not sure it's possible.

But I have a simpler approach:

```
CMDLINE_ROOTFS:append = " root=${@'$%s' % '{mender_kernel_root}'} "
```

And btw: Why did this work previously? 🙄

kacf (approved):

meta-mender-raspberrypi/recipes-bsp/bootfiles/rpi-cmdline-mender.inc
kacf: Ok. I see the problem. I think their upstream version is invalid actually. @TheYoctoJester said recently after talking to some OpenEmbedded people, that `${CMDLINE}` syntax is not valid within Python snippets, and has been working sort of by accident only in certain circumstances. I believe it should be changed to `d.getVar('CMDLINE')`.

So you could maybe attempt to upstream that. But I won't stand in the way of this PR, this is just a small and insignificant cleanup.

PR #1767, opened 2022-08-31
test: Test that we only listen on one port, and only on 127.0.0.1.

kacf:
Ticket: MEN-5824

### DIFF
--- a/tests/acceptance/test_dbus.py
+++ b/tests/acceptance/test_dbus.py
@@ -109,6 +109,14 @@ class TestDBus:
             assert f'string "{self.JWT_TOKEN}' in output
             if version_is_minimum(bitbake_variables, "mender-client", "3.3.1"):
                 assert 'string "http://127.0.0.1:' in output
+                # Check that this really is the only port we're listening on.
+                lines = connection.run("netstat -t -u -l -p -n").stdout.split("\n")
+                lines = [line for line in lines if line.strip().endswith("/mender")]
+                assert len(lines) == 1
+                address_and_port_index = 3
+                listen = lines[0].split()[address_and_port_index]
+                assert listen.startswith("127.0.0.1:")
+                assert f'string "http://{listen}"' in output
             elif version_is_minimum(bitbake_variables, "mender-client", "3.2.0"):
                 assert 'string "http://localhost:' in output
             else:


kacf:

tests/acceptance/test_dbus.py
```diff
@@ -109,6 +109,13 @@ def test_dbus_get_jwt_token(self, bitbake_variables, connection, setup_mock_serv
             assert f'string "{self.JWT_TOKEN}' in output
             if version_is_minimum(bitbake_variables, "mender-client", "3.3.1"):
                 assert 'string "http://127.0.0.1:' in output
+                # Check that this really is the only port we're listening on.
+                lines = connection.run("netstat -t -u -l -p -n").stdout.split("\n")
```
merlin-northern, line 113: I once went down this road, and then found [this one](https://github.com/mendersoftware/integration/blob/master/tests/tests/test_keepalive.py#L135) an alternative. have you considered doing something like:

```bash
cat /proc/`pidof mender`/net/tcp | grep -E '[^:]+: [^ ]+ [^ ]+ 0A '
```

kacf: Is it better though? The proc file needs to be decoded before it matches the string.

merlin-northern: > Is it better though? The proc file needs to be decoded before it matches the string.

I said "an alternative" on purpose :) I do not know about better, advantages:

* explicitly checks the ports in the listening state for given process
* parsing a file maybe seen as more elegant solution than running netstat and parsing output looking for /mender

do you think you can make it work with /proc parsing?

kacf: Well, my command also checks for a given process. As for parsing a file, well, maybe, if we assume that the output from `netstat` is unpredictable depending on the exact version for example. But I don't think this is the case, and using the file is quite cumbersome since it is based on hex. When using hex values, neither the port number nor the address is greppable in the source code, and it is hard to read as well. In addition, my command also checks UDP.

I vote for simplicity in this case. If it turns out that `netstat` output changes incompatibly during the next build host upgrade, then you have my official permission to say: I told you so! 🙂

kacf: @merlin-northern: With that in mind, do you think it is worth an approval? :pray:

merlin-northern: one thing about me: no matter what I never say: "I told you so" :) ok, let do it the netstat way.

tests/acceptance/test_dbus.py
```diff
@@ -109,6 +109,13 @@ def test_dbus_get_jwt_token(self, bitbake_variables, connection, setup_mock_serv
...
+                lines = [line for line in lines if line.strip().endswith("/mender")]
+                assert len(lines) == 1
+                listen = lines[0].split()[3]
```
merlin-northern, line 116: a bit of a nit, but I always say it: `local_address_column_index=3` (as in: could we have it defined instead of a magic number?)

merlin-northern (approved):

tests/acceptance/test_dbus.py
kacf: Alright, makes sense. Fixed!

kacf:
This already passed in https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/627423720, and the only change is adding a named variable, so I will merge this.

PR #1768, opened 2022-08-31
perf(update-firmware-state-script) Improve robustness

lluiscampos:
Changelog: meta-mender-raspberrypi: update-firmware-state-script: Make
copies of firmware binaries safer. On VFAT file-systems `mv` is not
atomic, but still using it narrows the time window in which the files
could be corrupted.


kacf (approved):
👍

lluiscampos:
Tested manually - it works.

Not running CI as this script is not part of any automated test.

PR #1774, opened 2022-09-05
Revert "fix(rpi-cmdline-mender.inc): Fix cmdline.txt after generation"

lluiscampos:
Upstream has accepted and merged our patch, so this workaround is not
necessary anymore :)

See https://github.com/agherzan/meta-raspberrypi/pull/1080

This reverts commit 446c941f70d0068d9d480d7e605952525fd92cb4.


kacf (approved):
Nice, and quick reaction time from them!

PR #1776, opened 2022-09-07
fix: Update pseudo db when excluding content from rootfs

kacf:
This commit is inspired the commit below from poky: https://git.yoctoproject.org/poky/commit/?id=f85a4a1462bc2105f7c62f2b09c1d092c7677ada

Although the code is different, it is applying the exact same type of fix, and therefore I'm also reusing (most of) their commit message below:

---

To exclude content from the rootfs, the prefunc makes a copy (using hardlinks if possible) of the rootfs directory and associated pseudo db, then removes files & directories as needed. However if these files and directories are removed using the python functions os.remove and shutil.rmtree, the copied pseudo db will not be updated correctly. For files copied from the original rootfs, if hardlinks were used successfully when copying the rootfs this should mean that the relevant inodes can't be reused and so the risk of pseudo aborts should be avoided. However, this logic doesn't apply for directories (as they can't be hardlinked) and so there remains some risk of inodes being reused and the pseudo db becoming corrupted.

To fix this, use the 'rm' command under pseudo when removing files & directories from the copied rootfs to ensure that the copied pseudo db is updated.

Changelog: Fix occasional pseudo abort issue when building certain images, especially in parallel. The symptom was this message:
```
abort()ing pseudo client by server request.
See https://wiki.yoctoproject.org/wiki/Pseudo_Abort for more details on this
```

Ticket: MEN-5857

### DIFF
--- a/meta-mender-core/classes/mender-setup-image.inc
+++ b/meta-mender-core/classes/mender-setup-image.inc
@@ -109,7 +109,7 @@ python prepare_excluded_directories() {
         bb.debug(1, "'IMAGE_ROOTFS_EXCLUDE_PATH' is set but 'respect_exclude_path' variable flag is 0 for this image type, so ignoring it")
         return
 
-    import shutil
+    import subprocess
     from oe.path import copyhardlinktree
 
     exclude_list = exclude_var.split()
@@ -120,7 +120,7 @@ python prepare_excluded_directories() {
     new_rootfs = os.path.realpath(os.path.join(d.getVar("WORKDIR"), "rootfs.%s" % taskname))
 
     if os.path.lexists(new_rootfs):
-        shutil.rmtree(os.path.join(new_rootfs))
+        subprocess.check_call(["rm", "-rf", new_rootfs])
 
     copyhardlinktree(rootfs_orig, new_rootfs)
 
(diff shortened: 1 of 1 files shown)


meta-mender-core/classes/mender-setup-image.inc
```diff
@@ -146,12 +146,12 @@ python prepare_excluded_directories() {
...
                 if os.path.isdir(full_entry) and not os.path.islink(full_entry):
-                    shutil.rmtree(full_entry)
+                    subprocess.check_call(["rm", "-rf", full_entry])
                 else:
-                    os.remove(full_entry)
+                    subprocess.check_call(["rm", "-f", full_entry])
         else:
             # Delete whole directory.
-            shutil.rmtree(full_path)
+            subprocess.check_call(["rm", "-rf", full_entry])
```
lluiscampos, line 154:
```suggestion
            subprocess.check_call(["rm", "-rf", full_path])
```

lluiscampos (changes requested):
Careful with one misspell :point_down: 

Otherwise look good! Such a simple fix for such a complex issue :sweat_smile:

meta-mender-core/classes/mender-setup-image.inc
kacf: Wow! Nicely spotted! I suppose this is never called for us, because all our entries end in a slash.

kacf:
@lluiscampos: Fixed! Approve?

I don't think I will rerun it. The fact that [it passed last time](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/633507454), proves that the fixed code never ran, because we don't use that variant of the exclude patterns in this repo.

oleorhagen (approved):

lluiscampos (approved):

PR #1791, opened 2022-09-23
chore: Work around bug in interaction between pseudo and pycache.

kacf:
At the time of writing, this appears to affect dunfell only, not kirkstone, although the reason for this is unclear.

Ticket: MEN-5857

### DIFF
--- a/tests/build-conf/beagleboneblack/local.conf
+++ b/tests/build-conf/beagleboneblack/local.conf
@@ -243,6 +243,10 @@ BB_DISKMON_DIRS = "\
 # this doesn't mean anything to you.
 CONF_VERSION = "1"
 
+# MEN-5857: Get rid of pseudo abort errors which happen because __pycache__
+# directories are created inside ${COREBASE}/bitbake/lib.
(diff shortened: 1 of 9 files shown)


lluiscampos:
Being tested at #1792.

lluiscampos:
@kacf Unfortunately, this does not solve the issue. See [this build](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/3080631920) from [this PR](https://github.com/mendersoftware/meta-mender/pull/1792) including your fix.

kacf:
@lluiscampos: Thanks. I knew about that location actually (it had shown up when i searched for `__pycache__` directories on the build host), but I hadn't seen it in build errors before, so I omitted it. I added it now, and hopefully this is the last one.

lluiscampos:
> @lluiscampos: Thanks. I knew about that location actually (it had shown up when i searched for `__pycache__` directories on the build host), but I hadn't seen it in build errors before, so I omitted it. I added it now, and hopefully this is the last one.

I refreshed now the patch on my PR and I will use it in the upcoming build.

lluiscampos (approved):
I have not seen pseudo aborts while building the images in the last batch of pipelines :+1:

PR #1795, opened 2022-09-28
ci: enforce the use of KVM in our CI

oleorhagen:
This will only try and run QEMU with kvm enabled.

If KVM is not enabled, the script will fail, and we will be able to catch that something is amiss with the CI setup of test slaves.

Ticket: QA-414

### DIFF
--- a/meta-mender-qemu/scripts/mender-qemu
+++ b/meta-mender-qemu/scripts/mender-qemu
@@ -123,8 +123,15 @@ $QEMU_SYSTEM --version || exit 1
 
 echo "--- starting qemu"
 
+USE_KVM=' -enable-kvm ""'
+if [ -n "${GITLAB_CI}" ]; then
+    # Enforce kvm in our CI
+    USE_KVM=' -enable-kvm'
+fi
+
+
 ret=0
-for maybe_kvm in -enable-kvm ""; do
+for maybe_kvm in ${USE_KVM}; do
     if QEMU_AUDIO_DRV=none \
             $QEMU_SYSTEM \
             -m 256M \


oleorhagen:
Slightly different approach, and a tad more hardcore (in fact it will fail our acceptance pipelines as is now).

if you think it's a bit too hardcore, let me know. However, if CI is going to fail anyways, I don't see a problem with this.

Feel free to tell me off however :)

meta-mender-qemu/scripts/mender-qemu
```diff
@@ -123,8 +123,15 @@ $QEMU_SYSTEM --version || exit 1
 
 echo "--- starting qemu"
 
+USE_KVM=' -enable-kvm ""'
```
kacf, line 126: I believe this kind of expansion won't work, string quoting is done before variable expansions, so you get the literal `""` value instead.

kacf (changes requested):
I think the method itself is fine, nothing wrong with being heavy handed in this particular aspect. However, this is a user facing script, so I don't think we should put CI specific code in here. My initial suggestion would be to use a `FORCE_KVM` variable instead of `GITLAB_CI`, and then set that variable in the `.gitlab-ci.yml` file. This makes it more generic.

OTOH, I have a suggestion which is perhaps easier. Simply add this early in the `yocto-build-and-test.sh` file:

```
test -w /dev/kvm
```

And that's it. Check for writable `/dev/kvm`.

meta-mender-qemu/scripts/mender-qemu
oleorhagen: It works :)

oleorhagen:
> And that's it. Check for writable `/dev/kvm`.

Great idea!

lluiscampos:
> OTOH, I have a suggestion which is perhaps easier. Simply add this early in the `yocto-build-and-test.sh` file:
> 
> ```
> test -w /dev/kvm
> ```
> 
> And that's it. Check for writable `/dev/kvm`.

But this will not prevent us again to end up in the same situation (`/dev/kvm` not being writable by CI and tests silently running). I think an explicit CI variable is the best, so that it will fail if our CI infra changes again for some reason.

oleorhagen:
> But this will not prevent us again to end up in the same situation (`/dev/kvm` not being writable by CI and tests silently running). I think an explicit CI variable is the best, so that it will fail if our CI infra changes again for some reason.

I'm not following. How will this silently run? I'm guessing the idea is to fail when `/dev/kvm` is not writable.

lluiscampos:
> > But this will not prevent us again to end up in the same situation (`/dev/kvm` not being writable by CI and tests silently running). I think an explicit CI variable is the best, so that it will fail if our CI infra changes again for some reason.
> 
> I'm not following. How will this silently run? I'm guessing the idea is to fail when `/dev/kvm` is not writable.

Aha! I see. I got it wrong. Basically we assume that `yocto-build-and-test.sh` is only run by CI, so there we can safely fail. Ignore my concern, then.

oleorhagen:
Closing in favour of: https://github.com/mendersoftware/mender-qa/pull/612/files

PR #1797, opened 2022-09-30
Allow to use Label with Partuuid

clementperon:
Hi,

At the moment, we don't allow to use both PartUUID and PartLabel (which is a bad name choose by me :( because it's based on Label and not PartLabel)

But It can make sense in case a user want to use the label on /etc/fstab and boot linux with the cmdline root=PARTUUID=XXX 

Also in my case it makes sense I have already shipped some hardware where the PARTUUID in the GPT are random but the LABEL are properly set, I want to be able to support this machine so I need to keep my artifact to use the LABEL in the /etc/fstab and mender-grow-data-part but I want my next machine to have a proper PartUUID feature for booting so in the grub.cfg and mender.conf (see: https://hub.mender.io/t/qemu-kvm-vs-hyper-v-dev-sdx-or-dev-vdx/5254).

Hope it's clear,
BR

### DIFF
--- a/meta-mender-core/classes/mender-maybe-setup.bbclass
+++ b/meta-mender-core/classes/mender-maybe-setup.bbclass
@@ -78,8 +78,8 @@ python() {
         # Use PARTUUID to set fixed drive locations.
         'mender-partuuid',
 
-        # Use PARTLABEL to avoid hardcoded drive device path.
-        'mender-partlabel',
+        # Use LABEL to avoid hardcoded drive device path.
+        'mender-fslabel',
 
         # Setup the systemd machine ID to be persistent across OTA updates.
         'mender-persist-systemd-machine-id',
--- a/meta-mender-core/classes/mender-setup-image.inc
+++ b/meta-mender-core/classes/mender-setup-image.inc
@@ -39,12 +39,12 @@ mender_update_fstab_file() {
     local tmpDataPart="${MENDER_DATA_PART}"
     local partuuid="${@bb.utils.contains('MENDER_FEATURES', 'mender-partuuid', 'true', 'false', d)}"
     local partlabel="${@bb.utils.contains('MENDER_FEATURES', 'mender-partlabel', 'true', 'false', d)}"
-    if [ ${partuuid} = "true" ]; then
-        tmpBootPart="PARTUUID=${@mender_get_partuuid_from_device(d, '${MENDER_BOOT_PART}')}"
-        tmpDataPart="PARTUUID=${@mender_get_partuuid_from_device(d, '${MENDER_DATA_PART}')}"
-    elif [ ${partlabel} = "true" ]; then
+    if [ ${partlabel} = "true" ]; then
         tmpBootPart="LABEL=${MENDER_BOOT_PART_LABEL}"
         tmpDataPart="LABEL=${MENDER_DATA_PART_LABEL}"
+    elif [ ${partuuid} = "true" ]; then
+        tmpBootPart="PARTUUID=${@mender_get_partuuid_from_device(d, '${MENDER_BOOT_PART}')}"
+        tmpDataPart="PARTUUID=${@mender_get_partuuid_from_device(d, '${MENDER_DATA_PART}')}"
     fi
 
     if ! ${@bb.utils.contains('MENDER_FEATURES', 'mender-image-ubi', 'true', 'false', d)}; then
--- a/meta-mender-core/classes/mender-setup-image.inc
+++ b/meta-mender-core/classes/mender-setup-image.inc
@@ -38,8 +38,8 @@ mender_update_fstab_file() {
     local tmpBootPart="${MENDER_BOOT_PART}"
     local tmpDataPart="${MENDER_DATA_PART}"
     local partuuid="${@bb.utils.contains('MENDER_FEATURES', 'mender-partuuid', 'true', 'false', d)}"
-    local partlabel="${@bb.utils.contains('MENDER_FEATURES', 'mender-partlabel', 'true', 'false', d)}"
-    if [ ${partlabel} = "true" ]; then
+    local fslabel="${@bb.utils.contains('MENDER_FEATURES', 'mender-fslabel', 'true', 'false', d)}"
+    if [ ${fslabel} = "true" ]; then
         tmpBootPart="LABEL=${MENDER_BOOT_PART_LABEL}"
         tmpDataPart="LABEL=${MENDER_DATA_PART_LABEL}"
     elif [ ${partuuid} = "true" ]; then
--- a/meta-mender-core/classes/mender-setup.bbclass
+++ b/meta-mender-core/classes/mender-setup.bbclass
@@ -262,7 +262,7 @@ python mender_sanity_handler() {
     if bb.utils.contains('MENDER_FEATURES_ENABLE', 'mender-partuuid', True, False, d) and bb.utils.contains('MENDER_FEATURES_ENABLE', 'mender-uboot', True, False, d):
         bb.fatal("mender-partuuid is not supported with mender-uboot.")
 
-    if bb.utils.contains('MENDER_FEATURES_ENABLE', 'mender-partuuid', True, False, d) and bb.utils.contains('MENDER_FEATURES_ENABLE', 'mender-partlabel', True, False, d):
+    if bb.utils.contains('MENDER_FEATURES_ENABLE', 'mender-partuuid', True, False, d) and bb.utils.contains('MENDER_FEATURES_ENABLE', 'mender-fslabel', True, False, d):
         bb.note("mender-partuuid will be only used for GPT and bootloader.")
 }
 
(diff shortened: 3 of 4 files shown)


oleorhagen:
Hi @clementperon :smile_cat: 

Thank you for this. I will take a loot at this later today. We are sorry for the long delay in reviewing this - and thank you for your patience!

First off, your commits look good, they are just failing CI because we require a `Ticket` in all `feats`. So since this is untracked, just add a:

`Ticket: None`

verbatim, and the linter should be happy :smiley_cat:

oleorhagen:
I will commence after you have fixed the commit messages :pray:

oleorhagen:
Alright, I'm not so familiar with this that I want to sign off on this. Passing it off to layer-owner

kacf (commented):
I'm not sure about this. It seems inconsistent to have different priority for fstab and the boot scripts. I understand your particular use case, but it seems that this is more about fixing a mistake made in the past.

Or put differently: Why would a new user want to do this?

clementperon:
Indeed as a new user I don't see why I should not use PartUUID (except if some devices doesn't support GPT?).

But as of today PartLabel feature only touch the non-persistent files:
 - /etc/fstab
 - /usr/bin/mender-client-resize-data-part
 Whereas PartUUID also touch (persistent files): 
 - /data/mender/mender.conf
 - grub.cfg
 - Partition table
 
 In case a user enable both Partlabel and PartUUID it doesn't make sense to mix behavior between non-persistent files.
 And it doesn't make sense to keep the full PartUUID behavior as it will be same as disabling PartLabel.
 
 This MR allow to enable both PartLabel and PartUUID and keep the same behavior for non-persistent files using LABEL for both files (fstab and resize-data-part)
 
Hope it makes sense but agree with you If I could go back in time I would use grub on all my devices + PartUUID everywhere.

kacf:
Hmm ok, I see. So this is already inconsistent. What it actually means is that PartUUID affects both rootfs partitions, as well as the boot and data partitions. PartLabel on the other hand, affects only the boot and data partitions. Also, I realize that your earlier comment about PartLabel actually meaning Label (as in FS label) is important, because it is the reason that Labels can not be used for rootfs partitions.

To be honest I think a better fix here would be:

* Clean up the naming: Rename `mender-partlabel` to `mender-fslabel`.

* Instead of relying on priorities, make it possible to specify explicitly which feature you want to use in each case. I suggest something like this (see [docs on rootfs variables](https://docs.mender.io/system-updates-yocto-project/variables#mender_rootfs_part_a)):
  ```
   MENDER_ROOTFS_PART_A_FEATURES = "mender-fslabel"
  ```
  to enable the feature for that partition specifically. This fits right into how extra partitions work, which would then be something like (see [docs on extra partitions](https://docs.mender.io/system-updates-yocto-project/variables#mender_extra_parts)):
  ```
  MENDER_EXTRA_PARTS_FEATURES[part1] = "mender-fslabel"
  ```

* If only one of the `mender-partuuid` and `mender-fslabel` global features are enabled, use that one for everything. If both are enabled, require the partition specific variables above in cases where there would be a conflict.

* We keep this on master only, which will become the next Yocto LTS.

This is a bit more work, but I think this is ultimately a better and less surprising approach. What do you think?

clementperon:
> Hmm ok, I see. So this is already inconsistent. What it actually means is that PartUUID affects both rootfs partitions, as well as the boot and data partitions. PartLabel on the other hand, affects only the boot and data partitions. Also, I realize that your earlier comment about PartLabel actually meaning Label (as in FS label) is important, because it is the reason that Labels can not be used for rootfs partitions.
> 
> To be honest I think a better fix here would be:
> 
> * Clean up the naming: Rename `mender-partlabel` to `mender-fslabel`.

Agree with that

> * Instead of relying on priorities, make it possible to specify explicitly which feature you want to use in each case.
> * If only one of the `mender-partuuid` and `mender-fslabel` global features are enabled, use that one for everything. If both are enabled, require the partition specific variables above in cases where there would be a conflict.

In case of data and boot partition I still want to generate fixed PartUUID for new machine in my GUID Partition Label and keep using LABEL in /etc/fstab to be compatible with previous machine.

My approach was more simple as it's doesn't make sense to prioritize `mender-partuuid` over `mender-fslabel`
let's prioritize `mender-fslabel` over `mender-partuuid` and we can display a note / warning about it.
Also `mender-fslabel` will not broke user that already use `mender-partuuid` which is not true for the opposite.

> * We keep this on master only, which will become the next Yocto LTS.

Not an issue for me, I will keep patch downstream until next Yocto LTS. 

> 
> This is a bit more work, but I think this is ultimately a better and less surprising approach. What do you think?

Thanks for your review :)

clementperon:
@kacf I have pushed a commit to rename the feature partlabel to fslabel as you suggest.

kacf:
Apologies for the delay, @clementperon, I'm a bit backed up these days.

> @kacf I have pushed a commit to rename the feature partlabel to fslabel as you suggest.

Great!

> Also it's not possible to have `mender-fslabel` on ROOTFS_A / ROOTFS_B as the LABEL are stored in the partition and not the partition table. Those could be changed at the next update.

Agreed.

> My approach was more simple as it's doesn't make sense to prioritize `mender-partuuid` over `mender-fslabel`
> let's prioritize `mender-fslabel` over `mender-partuuid` and we can display a note / warning about it.
> Also `mender-fslabel` will not broke user that already use `mender-partuuid` which is not true for the opposite.

I had a chance to discuss this with @TheYoctoJester as well. We both agreed that doing prioritization is too "magical", especially for something as sensitive as this. Also, I don't agree that "it doesn't make sense to prioritize `mender-partuuid` over `mender-fslabel`", on the contrary I think this makes perfect sense, because a UUID is more specific than a label. You might want to use a label just for informational purposes, without disturbing the PARTUUID logic which is happening on the technical level.

But that's anyway slightly irrelevant, because we don't want prioritization at all. So I refer back to my suggestion from earlier, but I will amend it slightly, incorporating a suggestion by @TheYoctoJester:

* ~Clean up the naming: Rename `mender-partlabel` to `mender-fslabel`.~ (Already done)

* Instead of relying on priorities, make it possible to specify explicitly which feature you want to use in each case. I suggest something like this (see [docs on rootfs variables](https://docs.mender.io/system-updates-yocto-project/variables#mender_rootfs_part_a)):
  ```
   MENDER_ROOTFS_PART_A_FEATURES = "mender-fslabel"
  ```
    
  to enable the feature for that partition specifically. This fits right into how extra partitions work, which would then be something like (see [docs on extra partitions](https://docs.mender.io/system-updates-yocto-project/variables#mender_extra_parts)):
  ```
  MENDER_EXTRA_PARTS_FEATURES[part1] = "mender-fslabel"
  ```

* **Only allow one of the `mender-partuuid` and `mender-fslabel` global features. Use that feature for everything where the `PARTS_FEATURES` variable is empty**.

* We keep this on master only, which will become the next Yocto LTS.

clementperon:
Agree prioritization is a bit magical.

Things is that the fslabel feature is not attached to a specific partition.

I'm not sayin I want to use LABEL for /data and /boot, but I want to use LABEL for /etc/fstab and /usr/bin/mender-grow-data-part

I also want to keep both PARTUUID and LABEL for new machine in partition table / grub configuration.
So with your suggestion my configuration will be something like this:
```
MENDER_BOOT_FEATURES = "mender-fslabel"
MENDER_ROOTFS_PART_A_FEATURES = "mender-partuuid"
MENDER_ROOTFS_PART_B_FEATURES = "mender-partuuid"
MENDER_DATA_FEATURES = "mender-fslabel"
```
or 
```
MENDER_FEATURES = "mender-partuuid"
MENDER_BOOT_FEATURES = "mender-fslabel"
MENDER_DATA_FEATURES = "mender-fslabel"
```
Did I unsterdand correctly?

kacf:
Correct! And if any of those aren't set, the global option takes effect.

clementperon:
> Correct! And if any of those aren't set, the global option takes effect.

I understand it's no more magic but how do you deal with something like this:

```
MENDER_BOOT_FEATURES = "mender-partuuid"
MENDER_ROOTFS_PART_A_FEATURES = "mender-fslabel"
MENDER_ROOTFS_PART_B_FEATURES = "mender-partuuid"
MENDER_DATA_FEATURES = "mender-fslabel" 
```

My problem is that the feature are not really attached to the partitions themselves but more to the boot logic (partuuid) and mender logic in rootfs (fslabel)

kacf:
Yeah, a mix would not work for the two rootfs partitions. Actually, now that I think of it, `mender-fslabel` makes no sense for the rootfs partitions; they only support partition-level identifiers. But a potential future `mender-partlabel` (and I mean an actual partition label, not the misnamed one) could be introduced in the future.

For now I think we can simply skip the `MENDER_ROOTFS_PART_A_FEATURES` and `MENDER_ROOTFS_PART_B_FEATURES` variables, since they only have one possible value, and focus on the other three: `MENDER_BOOT_FEATURES`, `MENDER_DATA_FEATURES` and `MENDER_EXTRA_PARTS_FEATURES`.

PR #1800, opened 2022-10-02
QA-414: Handle the QEMU client running out of memory

oleorhagen:
The `map-insert` shell script fills up the running memory of the client, when run for long enough.

Simply fix this through logging `Error` and above.


oleorhagen:
Not a valid fix

PR #1801, opened 2022-10-03
test(update_control): fix the issue with /run logs filling up mem

oleorhagen:
After a bit of digging, I discovered that when KVM is once again re-enabled in our acceptance test jobs, the client would run out of memory, as a result of the 'systemd-run' job filling up the `/run`, and hence the running memory of the QEMU instance.

The fix is to create a stricter journald config for our CI-layer.

The settings are pretty strict, but I don't necessarily see a big problem here.

With these settings a vacuum of the log is triggered at tighter intervals, so less logs are kept. But still plenty of MiB's available for log storage.

Ticket: QA-414

### DIFF
--- a/tests/meta-mender-ci/recipes-core/systemd-conf/files/journald.conf
+++ b/tests/meta-mender-ci/recipes-core/systemd-conf/files/journald.conf
@@ -1,0 +1,11 @@
+[Journal]
+SyncIntervalSec=1m
+RateLimitBurst=1000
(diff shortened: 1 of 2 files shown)


kacf (approved):
Code looks alright, but why do we need to forward to syslog? Doesn't that just duplicate the logs and use more space?

lluiscampos:
* kirkstone

PR #1804, opened 2022-10-06
chore: Remove python3-websockets append, not needed on kirkstone

lluiscampos:


lluiscampos:
Build @ [Pipeline 660125513](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/660125513)

oleorhagen (approved):
Why?

Is it already included?

lluiscampos:
> Why?
> 
> Is it already included?

In theory this should not be needed in the first place. Our software (the mock server) imports `websockets`, and `python3-websockets` package should take care of its dependencies. At the time of implementing this in `dunfell` it was crashing with the error that you can see in this append, and to not push upstream to fix this I introduced a workaround in the append file. Now with `kirkstone` I just "tried" and it works :)

PR #1816, opened 2022-10-14
Dunfell: Fix build error with mender-artifact < 3.9.

kacf:


meta-mender-core/classes/mender-bootstrap-artifact.bbclass
```diff
@@ -1,5 +1,16 @@
 IMAGE_CMD_bootstrap-artifact() {
 
+    # Write a simple hardcoded bootstrap-artifact first, which we use to detect
+    # whether this version of mender-artifact supports bootstrap artifacts.
+    if ! mender-artifact write bootstrap-artifact \
+            --artifact-name test \
+            --device-type test \
+            --output-path "${WORKDIR}/test-artifact.mender"
```
oleorhagen, line 8: Suggestion: Output to /dev/null?

oleorhagen (approved):

meta-mender-core/classes/mender-bootstrap-artifact.bbclass
kacf: Admittedly I haven't tried it, but I don't think it will work because [we rename it](https://github.com/mendersoftware/mender-artifact/blob/master/cli/write.go#L246) to its final destination.

PR #1846, opened 2022-11-10
properly get the SRCPV inside the function

clementperon:
@kacf related to https://github.com/mendersoftware/meta-mender/commit/b31ab892d6ccdbec7b880c73f0e2d795e9866e34

Switch to 'd.getVar()' style, since this now fails on kirkstone and later.

We can get it directly inside the function

Changelog: None

Signed-off-by: Clément Péron <peron.clem@gmail.com>


kacf (changes requested):
Thanks, @clementperon!

You need to prefix your commit title with `fix: `, and add a `Ticket: None` next to the changelog tag (I assume there is no ticket for this).

clementperon:
@kacf done !

PR #1847, opened 2022-11-10
devtools: drop dosfstools and use the upstreamed one

clementperon:
The dosfstools recipe in openembedded-core is also 4.2 avoid maintaining this one and use the upstream one

SEE: https://layers.openembedded.org/layerindex/recipe/300194/


kacf (changes requested):
Thanks!

Please prefix the commit title with `fix: ` and add a `Ticket: None` next to the changelog entry.

clementperon:
@kacf done

PR #1859, opened 2022-11-24
Several test_update_control fixes

kacf:


tests/acceptance/test_update_control.py
```diff
@@ -578,7 +578,15 @@ def test_update_control_maps(
 
             if case.get("take_at_least"):
                 assert (
-                    time.time() - now >= case["take_at_least"]
+                    # Why subtract one second from `case` here? Turns out that this fails
```
oleorhagen, line 583: Rhetorical questions - I like it! :laughing:

oleorhagen: Still, a little worrying...

oleorhagen (approved):
Nice find!

There are quite a few treasures strewn around in our test setup

tests/acceptance/test_update_control.py
kacf: > Rhetorical questions - I like it! laughing

Haha, I was thinking to myself as I wrote it: Am I too philosophical? 😂

It is a little worrying, but I've never seen it get worse than a few tens of milliseconds. I guess this will either prove or disprove that.

lluiscampos: Nice comment. I overall agree with the approach - it will be interesting to see if we can revert this after the C++ rewrite :thinking:

PR #1866, opened 2022-12-09
test(test-connect): Fix the mender-connect spurious failure tests

oleorhagen:
This fixes the mender-connect spurious failure tests.

The issue turned out to be that the time is passed in between the host and the QEMU instance, and is being used as a single source of truth.

This created a window of opportunity in which the test would get the system time on the host, then pass it into the QEMU instance to use as a timestamp in `since=<time>` when filtering the logs with `journald`.

Hence, mender-connect would actually behave, but log the correct output in the window of time which arose in between the host time, and QEMU's time, and it was simple filtered out.

The fix is dead simple: Use the QEMU instance time as the single source of truth

Ticket: QA-483

### DIFF
--- a/tests/acceptance/test_mender-connect.py
+++ b/tests/acceptance/test_mender-connect.py
@@ -22,6 +22,7 @@ import pytest
 from utils.common import (
     cleanup_mender_state,
     put_no_sftp,
+    qemu_system_time,
 )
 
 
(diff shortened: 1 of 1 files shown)


oleorhagen:
Depends on: https://github.com/mendersoftware/mender-image-tests/pull/78

oleorhagen:

lluiscampos (approved):

PR #1868, opened 2022-12-09
chore: Add variables from meta-mender-raspberry-pi-demo

lluiscampos:
Fixes build warnings for not recognized MENDER_ variable when using
`meta-mender-raspberry-pi-demo` layer.


TheYoctoJester:
Hmm actually I don't think it is a good idea. If people look at the layer and how to do thing, they might take it as blueprints for their own actions. So instead of "fixing" the core layer to adapt it to the demo, My understanding is that we should rather move the variables of the demo layer out of the `MENDER_` namespace.

Thoughts @kacf maybe?

lluiscampos:
> Hmm actually I don't think it is a good idea. If people look at the layer and how to do thing, they might take it as blueprints for their own actions. So instead of "fixing" the core layer to adapt it to the demo, My understanding is that we should rather move the variables of the demo layer out of the `MENDER_` namespace.
> 
> Thoughts @kacf maybe?

The thing is that any variable with `MENDER_` prefix that is not in the list will issue a warning. If you look at the existing list, there are others that don't belong to `meta-mender-core`.

What you are suggesting, if valid, we need to apply to the existing variables too.

TheYoctoJester:
> The thing is that any variable with `MENDER_` prefix that is not in the list will issue a warning. If you look at the existing list, there are others that don't belong to `meta-mender-core`.
> 
> What you are suggesting, if valid, we need to apply to the existing variables too.

Exactly. Instead of applying a band-aid in the form of adding the variable names to the list just to silence the warnings for one specific usecase (meta-mender-raspberrypi-demo), I would rather change the layer to not use any `MENDER_` prefix at all. I would even prefer to have the warning rather than the band-aid, because it makes it immediately clear to users that putting things into the namespace is a bad idea.

The `MENDER_DEMO_HOST_IP` one is semantically different because it directly affects the client.

oleorhagen (approved):

kacf (approved):

PR #1871, opened 2022-12-19
feat: Add the mender-client-cpp build support

oleorhagen:
Ticket: MEN-5796
Changelog: Add the mender-client-cpp recipe while still keeping support for the old golang client builds.


oleorhagen:
Obviously not finished, but I'd like to get you in the loop, so that we can discuss where this goes from here.

Unresolved issues: 

* client-cpp is currently just a separate recipe, and not pulled in by default
* The `mender-modules-gen` package is now provided by both `mender-client` and `mender-client-cpp` which Yocto is not happy about.

meta-mender-core/recipes-mender/mender-client/mender-client-cpp_git.bb
```diff
@@ -0,0 +1,14 @@
+require mender-client-cpp.inc
```
oleorhagen, line 1: Question is whether or not we should replace this with the `golang` mender-client_git recipe at this point in time.

meta-mender-core/recipes-mender/mender-client/mender-client.inc
```diff
@@ -14,7 +14,6 @@ RDEPENDS:${PN}:append:mender-growfs-data:mender-systemd:mender-partlabel = " uti
 RDEPENDS:${PN}:append:mender-growfs-data:mender-systemd:mender-partuuid = " util-linux-blkid"
 RDEPENDS:mender-modules-gen = "bash"
```
oleorhagen, line 15: Question is: like we discussed. What to do with this?

You suggested moving it to `mender-artifact`. But having a separate recipe for it is also a good option in my opinion. What do you think?

meta-mender-core/recipes-mender/mender-client/mender-client-cpp_git.bb
```diff
@@ -0,0 +1,14 @@
...
+LIC_FILES_CHKSUM = " \
+    file://LICENSE;md5=4cd0c347af5bce5ccf3b3d5439a2ea87 \
+    file://vendor/googletest/LICENSE;md5=cbbd27594afd089daa160d3a16dd515a \
+    file://vendor/json/LICENSE.MIT;md5=f969127d7b7ed0a8a63c2bbeae002588 \
+    file://vendor/json/docs/mkdocs/docs/home/license.md;md5=970ea048f6ea7d04eeb3ba3ef9ebca40 \
+"
+
+SRCREV = "${AUTOREV}"
+
+SRC_URI = "gitsm://github.com/oleorhagen/mender;protocol=https;branch=yocto-build-branch"
```
oleorhagen, line 14: Obviously not the final branch

meta-mender-core/recipes-mender/mender-client/mender-client.inc
oleorhagen: If we just move the `client-cpp` recipe to be the current git recipe, I guess we don't have to do anything really.

lluiscampos: > You suggested moving it to `mender-artifact`. 

My point was not to move it to `mender-artifact` _recipe_, but that ideally the source of these scripts could have been moved to `mender-artifact` repo and then here it will be natural to be installed by mender-artifact (which provides only a `native` package). Also I don't think you can have one yocto recipe pulling from two different git sources (and if it is possible smells like a very bad practice).

> But having a separate recipe for it is also a good option in my opinion. What do you think?

That would be ideal and if so should also be applied to the current golang recipes. But somewhat is a side little project on itself and it is independent of the cpp client

meta-mender-core/recipes-mender/mender-client/mender-client.inc
```diff
@@ -66,49 +61,43 @@ FILES:mender-modules-gen = " \
...
-# like. We disable those checks here.
-INSANE_SKIP:${PN} = "ldflags textrel"
-INSANE_SKIP:${PN}-ptest = "ldflags textrel"
-
-GO_IMPORT = "github.com/mendersoftware/mender"
-
 _MENDER_PACKAGECONFIG_DEFAULT = "${@mender_feature_is_enabled("mender-client-install", "dbus modules inventory-network-scripts", "", d)}"
 PACKAGECONFIG ?= "${_MENDER_PACKAGECONFIG_DEFAULT}"
 
+# TODO - How should this be split (?)
```
lluiscampos, line 100: They are all related to `mender-updated`, right? `mender-authd` has nothing to do with "client install" or bootloaders.

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
```diff
@@ -0,0 +1,17 @@
+require mender-client.inc
+
+RPROVIDES:${PN} = "mender-client"
```
kacf, line 5: You probably need to specify this:

```
RPROVIDES:mender-update = "mender-client"
RDEPENDS:mender-update = "mender-auth"
```

since it is that package which provides the complete thing (it depends on mender-auth, so will pull in that too).

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
```diff
@@ -0,0 +1,17 @@
...
+
+DESCRIPTION = "Mender tool for doing OTA software updates."
+HOMEPAGE = "https://mender.io"
+
+SYSTEMD_SERVICE:${PN} = "mender-updated.service mender-authd.service"
```
kacf, line 10: This needs to be split into the package names that contain the services. So it should be:

```
SYSTEMD_SERVICE:mender-auth = "mender-authd.service"
SYSTEMD_SERVICE:mender-update = "mender-updated.service"
```

This goes for quite a lot of our definitions. We used to use `${PN}`, but we will need to start using explicit names.

A slightly unusual thing about this recipe is that I think none of our packages will actually be named "mender-client". I don't know if this is problematic or not, but definitely not the common approach. Usually you have a "main" package with the same name as the recipe, with extra packages on the side.

I asked @TheYoctoJester, and he suggested we name the recipe simply "mender", and then produce the "mender-auth" and "mender-update" packages. Quote:

> if we want the packages `mender-auth` and `mender-update`, then the recipe name should be just `mender`. If the recipe name is `mender-client`, then the packages should be called `mender-client-auth` and `mender-client-update`.

Another possibility would be to completely split the recipe into two recipes with the corresponding names, but this would mean two instances of `SRC_URI` and the various hacks around it.

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
```diff
@@ -0,0 +1,17 @@
...
+
+MENDER_CLIENT ?= "mender-updated"
+
+FILES:${PN} += "\
```
kacf, line 14: Same, `${PN}` needs split.

meta-mender-core/recipes-mender/mender-client/mender-client-cpp_git.bb
```diff
@@ -0,0 +1,14 @@
...
+LICENSE = "Apache-2.0 & BSD-3-Clause"
+
+LIC_FILES_CHKSUM = " \
+    file://LICENSE;md5=4cd0c347af5bce5ccf3b3d5439a2ea87 \
+    file://vendor/googletest/LICENSE;md5=cbbd27594afd089daa160d3a16dd515a \
+    file://vendor/json/LICENSE.MIT;md5=f969127d7b7ed0a8a63c2bbeae002588 \
+    file://vendor/json/docs/mkdocs/docs/home/license.md;md5=970ea048f6ea7d04eeb3ba3ef9ebca40 \
+"
+
+SRCREV = "${AUTOREV}"
```
kacf, line 12: Shouldn't this be the `mender_autorev_if_git_version` variant from the original recipe?

meta-mender-core/recipes-mender/mender-client/mender-client-go.inc
```diff
@@ -0,0 +1,169 @@
+require mender-client.inc
+
+RCONFLICTS:${PN} = "mender-client-cpp"
```
kacf, line 3: AFAIK this need to be a package name, so it should be `mender-auth`.

meta-mender-core/recipes-mender/mender-client/mender-client.inc
```diff
@@ -50,62 +41,57 @@ FILES:${PN} += "\
     ${datadir}/mender/modules/v3/script \
     ${datadir}/mender/modules/v3/single-file \
     ${sysconfdir}/udev/mount.blacklist.d/mender \
-    ${systemd_unitdir}/system/${MENDER_CLIENT}.service \
     /data/mender/device_type \
     /data/mender/mender.conf \
 "
+
+FILES:${PN} += "\
```
kacf, line 48: Is `mender-client.inc` shared between Go and C++ variants? I suspect that this will lead to problems because `${PN}` will always be different for the two.

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
```diff
@@ -14,4 +16,7 @@ FILES:${PN} += "\
     ${systemd_unitdir}/system/mender-authd.service \
 "
 
-inherit autotools
+# Don't download the googletest source in Mender CMakefiles.txt
+EXTRA_OECMAKE:append = " -DMENDER_DOWNLOAD_GTEST=NO"
```
kacf, line 20: Even though we save bandwidth, this feels a bit hacky, especially since the license is also affected by it. Why can't we just clone it, even if we don't use it?

oleorhagen: It is because Yocto does not allow (at least without some hacking) downloading during anything but the `fetch` stage. 

I changed the client CMake build to accept this option, and instead rely on the pre-provided `googletest` recipe.

I would suspect this is not a problem?

kacf: Ok, that makes sense. But then it probably should be downloaded in the fetch step, so that the license check still passes, or? It might require some extra parameters to the `SRC_URI` string.

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
oleorhagen: While we're on the topic, should we split these further, like have sub-overrides like `mender-dbus` etc?

kacf: `mender-dbus` is not necessary I believe, since dbus is now required to run mender (you can't communicate with mender-auth otherwise), but in general we should consider splitting where appropriate.

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
oleorhagen: The recipe has a build dependency on `googletest`. Is that not enough?

oleorhagen: Oh, I see what you mean.

Remember, we don't vendor googletest anylonger

meta-mender-core/recipes-mender/mender-client/mender-client-cpp_git.bb
oleorhagen: Yup

oleorhagen: I will leave this be until the last iteration

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
oleorhagen: :+1: 

Alright, done the split. Feel free to have another look over

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
kacf: You're right, I missed that part. Ignore my ranting then!

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
```diff
@@ -0,0 +1,96 @@
...
+FILES:mender-update:append:mender-persist-systemd-machine-id = " \
+    ${systemd_unitdir}/system/mender-systemd-machine-id.service \
+    ${systemd_unitdir}/system/${MENDER_CLIENT}.service.wants/mender-systemd-machine-id.service \
+    ${bindir}/mender-set-systemd-machine-id.sh \
+"
+
+FILES:mender-update += "\
+    ${bindir}/mender-update \
+    ${datadir}/mender/identity \
+    ${datadir}/mender/identity/mender-device-identity \
```
kacf, line 57: These two should be part of mender-auth, not mender-update.

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
```diff
@@ -0,0 +1,96 @@
...
+    ${sysconfdir}/udev/mount.blacklist.d/mender \
+    /data/mender/device_type \
+    /data/mender/mender.conf \
+    ${datadir}/dbus-1/system.d/io.mender.UpdateManager.conf \
+"
+
+FILES:mender-auth += "\
+    ${bindir}/mender-auth \
+    /usr/share/dbus-1/interface \
+    /usr/share/dbus-1/interface/io.mender.Update1.xml \
```
kacf, line 79: This one should be part of mender-update.

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
```diff
@@ -0,0 +1,96 @@
...
+PACKAGES =+ "mender-auth mender-update"
+
+BBCLASSEXTEND = "native"
+RDEPENDS:mender-update:class-native = ""
+DEPENDS:class-native = ""
+
+RPROVIDES:mender-update = "mender-client"
+
+# TODO - This is really all it needs to RPROVIDE for the other recipes
+RPROVIDES:mender-auth = "virtual/mender-client"
```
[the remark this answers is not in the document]
oleorhagen: Hmm, ur probably right. 

I was thinking in terms of the fact that `configure` etc relies on `mender-client` through `RDEPENDS`.

But maybe we should keep it simple, and have `mender-update` provide it for now...

Yes.. Thinking about it. Lets!

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
oleorhagen: Good catch - thanks!

meta-mender-core/recipes-mender/mender-client/mender-client-cpp_git.bb
oleorhagen: Done

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
```diff
@@ -0,0 +1,96 @@
...
+RDEPENDS:mender-update = "mender-auth"
+
+PROVIDES += "mender-auth mender-update"
+
+DESCRIPTION = "Mender tool for doing OTA software updates."
+HOMEPAGE = "https://mender.io"
+
+SYSTEMD_SERVICE:mender-update = "mender-authd.service"
```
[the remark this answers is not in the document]
kacf: Look again. 🙂

meta-mender-core/recipes-mender/mender-client/mender_git.bb
```diff
@@ -0,0 +1,14 @@
...
+LIC_FILES_CHKSUM = " \
+    file://LICENSE;md5=4cd0c347af5bce5ccf3b3d5439a2ea87 \
+    file://vendor/json/LICENSE.MIT;md5=f969127d7b7ed0a8a63c2bbeae002588 \
+    file://vendor/json/docs/mkdocs/docs/home/license.md;md5=970ea048f6ea7d04eeb3ba3ef9ebca40 \
+    file://vendor/lmdbxx/UNLICENSE;md5=7246f848faa4e9c9fc0ea91122d6e680 \
+"
+
+SRCREV = "${@mender_autorev_if_git_version(d)}"
+
+SRC_URI = "gitsm://github.com/oleorhagen/mender;protocol=https;branch=cpp-cmake-files"
```
oleorhagen, line 14: Will fix once the this has passed CI alongside the mender-client PR: https://github.com/mendersoftware/mender/pull/1112

meta-mender-core/recipes-mender/mender-client/mender-client.inc
```diff
@@ -373,23 +192,84 @@ do_install:append:class-target:mender-growfs-data:mender-systemd() {
...
+    match = re.match(r"^[0-9]+\.[0-9]+\.", version)
+    if match is not None:
+        # If the preferred version is some kind of version, use the branch name
+        # for that one (1.0.x style).
+        return match.group(0) + "x"
+    else:
+        # Else return master as branch.
+        return "master"
+
+def mender_version_from_preferred_version(d):
```
kacf, line 245: We shouldn't keep these in a file which is shared between official recipes and Git recipes. There is some hacky stuff in the Git recipes, and we don't want that to "leak" into the official recipes (for example the license check is less strict).

kacf (changes requested):

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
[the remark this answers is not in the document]
oleorhagen: :see_no_evil:

kacf (approved):
I would still consider targeting `feature-c++-client`, but other than that this looks good. I'll leave the decision to you, since you know who will need to clean up if master breaks! 😉

oleorhagen:
> I would still consider targeting `feature-c++-client`, but other than that this looks good. I'll leave the decision to you, since you know who will need to clean up if master breaks! wink

Indeed :seedling:

meta-mender-core/recipes-mender/mender-client/mender-client-cpp_git.bb
TheYoctoJester: Yeah I just ran into that too. Wouldn't it be better to have a preliminary branch on the `mendersoftware/mender` repo and point to that instead of referencing a personal repo?

oleorhagen: Before it's merged, does it matter?

oleorhagen:
:green_circle: pipeline https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/736898492

PR #1894, opened 2023-01-31
Add support to create mender artifact for UEFI capsule

jainvikas8:
### Description

The existing mender artifact class can only be used for generating system/rootfs update artifacts. For the UEFI capsule, we need a custom module image artifact with specific flags/conditions enabled

These are the following changes:
* Add new class, this has a dependency on the UEFI capsule class from
  other layers which generates the required UEFI firmware capsule
* Add new variables
* Generate the `provides` and `depends` in the below format:
  `uefi-firmware.<guid>.name:<name>`
  `uefi-firmware.<guid>.version:<version>`
  by using `MENDER_ARTIFACT_CAPSULE_PROVIDES` and
  `MENDER_ARTIFACT_CAPSULE_DEPENDS` variables
* Add tests for mender capsule artifact

Please note the `meta-mender-core/classes/mender-artifactcapsule.bbclass` depends on `uefi_capsule` class from meta-arm


jainvikas8:
@kacf Please review. Thanks.

meta-mender-core/classes/mender-artifactcapsule.bbclass
```diff
@@ -0,0 +1,151 @@
...
+MENDER_ARTIFACT_CAPSULE_NAME ??= "${MENDER_ARTIFACT_CAPSULE_NAME_DEFAULT}"
+MENDER_ARTIFACT_CAPSULE_NAME_DEFAULT = "uefi-capsule-${MACHINE}"
+
+MENDER_ARTIFACT_CAPSULE_NAME_DEPENDS ?= ""
+
+MENDER_ARTIFACT_CAPSULE_PROVIDES ?= ""
+
+MENDER_ARTIFACT_CAPSULE_DEPENDS ?= ""
+
+MENDER_ARTIFACT_CAPSULE_CLEAR_PROVIDES ?= ""
```
kacf, line 20: Nitpick: Should be `CLEARS_PROVIDES` (extra 'S'). This is what the mender-artifact tool uses, and is also what you used below.

meta-mender-core/classes/mender-artifactcapsule.bbclass
```diff
@@ -0,0 +1,151 @@
...
+    if guids is None:
+        return
+    clears_provides = ""
+    for guid in guids:
+        clears_provides += " --clears-provides " + capsule_type + "." + guid + ".version"
+    return clears_provides
+
+MENDER_PROVIDES = "${@get_mender_capsule_artifact_provides(d)}"
+MENDER_CLEARS_PROVIDES = "${@get_mender_capsule_artifact_clears_provides(d)}"
+MENDER_DEPENDS = "${@get_mender_capsule_artifact_depends(d)}"
```
kacf, line 81: These are quite generic names, and they will be global. I suggest using `CAPSULE` in the name here as well, to avoid ambiguity.

You could also consider a `_MENDER` prefix, which are used in a few other places in meta-mender, since these variables aren't meant to be set directly by the user. That also means they no longer have to be in the `mender-vars.json` file.

meta-mender-core/classes/mender-artifactcapsule.bbclass
```diff
@@ -0,0 +1,151 @@
...
+
+    if [ -n "${MENDER_ARTIFACT_SIGNING_KEY}" ]; then
+        extra_args="$extra_args -k ${MENDER_ARTIFACT_SIGNING_KEY}"
+    fi
+
+    if mender-artifact write rootfs-image --help | grep -e '-u FILE'; then
+        image_flag="-u"
+    else
+        image_flag="-f"
+    fi
```
kacf, line 119: Cleanup: This conditional can be removed, even from the original location you grabbed it from if you feel like cleaning up some more. The version of mender-artifact which required the `-u` option is not available in the layer anymore (< 3.0).

meta-mender-core/classes/mender-maybe-setup.bbclass
```diff
@@ -89,6 +89,9 @@ python() {
 
         # Enable the testing/* layers and functionality
         'mender-testing-enabled',
+
+        # Enable UEFI Capsule artifact generation
+        'mender-image-capsule',
```
kacf, line 94: Should it be `mender-image-uefi-capsule`, for clarity? :thinking:

meta-mender-core/classes/mender-setup-image.inc
```diff
@@ -1,4 +1,5 @@
 IMAGE_FSTYPES += "${@bb.utils.contains('MENDER_FEATURES', 'mender-image', ' mender mender.bmap', '', d)}"
+IMAGE_FSTYPES += "${@bb.utils.contains('MENDER_FEATURES', 'mender-image-capsule', ' mender-capsule', '', d)}"
```
kacf, line 2: Same here, ref `uefi-capsule`.

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -230,7 +230,7 @@ MENDER_DATA_PART_LABEL ??= "data"
 # --------------------------- END OF CONFIGURATION -----------------------------
 
 IMAGE_INSTALL:append = " mender-client"
-IMAGE_CLASSES += "mender-part-images mender-ubimg mender-artifactimg mender-dataimg mender-bootimg mender-datatar"
+IMAGE_CLASSES += "mender-part-images mender-ubimg mender-artifactimg mender-artifactcapsule mender-dataimg mender-bootimg mender-datatar"
```
kacf, line 233: Even though the original is concatenated, it's better to split up the name I think. I suggest `mender-artifact-uefi-capsule` here as well.

meta-mender-core/classes/mender-artifactcapsule.bbclass
kacf, line 20: It would be great if these could be documented [here](https://github.com/mendersoftware/mender-docs/blob/master/05.System-updates-Yocto-Project/99.Variables/docs.md)!

kacf (changes requested):
Excellent work, @jainvikas8 and @gowthamsk-arm!

I have a few nitpicks regarding naming, but this is looking very good! Even acceptance tests, which makes me very happy! :smile:

meta-mender-core/classes/mender-artifactcapsule.bbclass
jainvikas8: There is no `master-next` branch there, should we expose those variables there, as this would be an unreleased feature?

meta-mender-core/classes/mender-setup.bbclass
jainvikas8: done

meta-mender-core/classes/mender-setup-image.inc
jainvikas8: done

meta-mender-core/classes/mender-maybe-setup.bbclass
jainvikas8: done

jainvikas8:
Forced pushed to address review comments.

jainvikas8:
@kacf Should we bump layer compatibility on master-next here to `langdale` as this is dependent on `meta-arm` having the `uefi_capsule` support?

meta-mender-core/classes/mender-artifactcapsule.bbclass
kacf: I created a new branch, `yocto-4.4`, you can submit the changes there!

meta-mender-core/classes/mender-artifactcapsule.bbclass
```diff
@@ -0,0 +1,151 @@
...
+        provides += " --provides " + capsule_type + "." + guid + ".version:" + splt[1]
+        provides += " --provides " + capsule_type + "." + guid + ".name:" + splt[0]
+    return provides
+
+def get_mender_capsule_artifact_depends(d):
+    # This function creates the depends format required by mender artifact to
+    # generate the capsule artifact.
+    # MENDER_ARTIFACT_CAPSULE_DEPENDS needs to be used in the following way.
+    #MENDER_ARTIFACT_CAPSULE_DEPENDS[e2bb9c06-70e9-4b14-97a3-5a7913176111] = "2"
+    #MENDER_ARTIFACT_CAPSULE_DEPENDS[e2bb9c06-70e9-4b14-97a3-5a7913176222] = "2"
```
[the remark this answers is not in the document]
kacf: I meant that the parsing would also be changed to accommodate it. It's not very important though, I just thought it might be easier to read. I leave it for you to decide.

meta-mender-core/classes/mender-artifactcapsule.bbclass
kacf: Exactly, they are not exposed. We often use `_MENDER` as a prefix for such variables to keep them "private", since public variables usually don't start with an underscore.

meta-mender-core/classes/mender-setup.bbclass
kacf: I meant the name as well, rename from `mender-artifactcapsule` to `mender-artifact-uefi-capsule`.

kacf:
> Should we bump layer compatibility on master-next here to `langdale` as this is dependent on `meta-arm` having the `uefi_capsule` support?

Yes, please do!

meta-mender-core/classes/mender-artifactcapsule.bbclass
jainvikas8: Done

meta-mender-core/classes/mender-setup.bbclass
jainvikas8: Done

jainvikas8:
Forced pushed to address review comments

meta-mender-core/classes/mender-artifactcapsule.bbclass
jainvikas8: @kacf PR - https://github.com/mendersoftware/mender-docs/pull/2008

kacf (approved):
Excellent, thank you @jainvikas8!

PR #1919, opened 2023-03-10
boot-partition-devicetree: proper SPDX License

moto-timo:
```
fix: boot-partition-devicetree QA warning about obsolete license (GPL-2.0)

Changelog: boot-partition-devicetree use proper SPDX License GPL-2.0-only
```
### Description
GPL-2.0 is no longer a valid license, avoid a warning by using GPL-2.0-only (which is the same as the kernel itself).


kacf:
Thanks @moto-timo! I resubmitted this in #1923 to get it into master first; then it will be cherry picked from there.

PR #1923, opened 2023-03-13
fix: boot-partition-devicetree QA license warning

kacf:
GPL-2.0 is no longer a valid license, avoid a warning by using GPL-2.0-only (which is the same as the kernel itself).

Changelog: boot-partition-devicetree use proper SPDX License GPL-2.0-only


kacf:
Failure is due to https://github.com/mendersoftware/mendertesting/pull/249.

PR #1926, opened 2023-03-22
feat: Make the build succeed if gateway examples aren't present

TheMeaningfulEngineer:
Changelog: Title
Ticket: None

If you have meta-mender-demo enabled for other reasons than the gateway, the build will fail if you didn't get the examples tar. This makes it a warning not a hard build fail.


TheMeaningfulEngineer:
Good point, didn't think.
Let me do a master PR.

PR #1928, opened 2023-03-24
systemd: Do not modify DISTRO_FEATURES_BACKFILE for mender-systemd config

drewmoseley:
Ultimately the root cause of the failure discussed in this post:
    https://hub.mender.io/t/toradex-verdin-imx8mm-error-building-the-sdk-with-mender/5654/3

is because mender-setup-systemd.inc adds "systemd" to DISTRO_FEATURES_BACKFILL. When building without the Mender layers, DISTRO_FEATURES differs when building a target recipe versus when building an SDK recipe:

    $ MACHINE=apalis-imx6 bitbake-getvar -r util-linux DISTRO_FEATURES --value
        acl alsa argp bluetooth debuginfod ext2 ipv4 ipv6 largefile pcmcia usbgadget usbhost wifi xattr \
        nfs zeroconf pci 3g nfc x11 vfat seccomp largefile opengl multiarch wayland vulkan bluez5 pam systemd \
        opengl wayland x11 polkit pulseaudio  gobject-introspection-data ldconfig

    $ MACHINE=apalis-imx6 bitbake-getvar -r nativesdk-util-linux DISTRO_FEATURES --value
        debuginfod opengl wayland x11 pulseaudio sysvinit gobject-introspection-data ldconfig

When systemd is not part of DISTRO_FEATURES, the logic in rm_systemd_unitdir()
  [https://github.com/openembedded/openembedded-core/blob/7df46e003ea76cf7d5b7263f23bd6e6a781bd22c/meta/classes/systemd.bbclass#L204]
takes care of cleaning up unneeded system service and timer files.  But since meta-mender adds systemd to DISTRO_FEATURES unconditionally (via DISTRO_FEATURES_BACKFILL) this cleanup does not happen and builds of SDKs will fail.

Note specifically that init-manager-systemd.inc
  [https://github.com/openembedded/openembedded-core/blob/kirkstone/meta/conf/distro/include/init-manager-systemd.inc]
does _not_ set DISTRO_FEATURES_BACKFILL. The best fix may be to set the INIT_MANAGER variable rather than explicitly setting all the other variables in mender-setup-systemd.inc but that is a much more invasive change that I have not tested.

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

Thank you!

### DIFF
--- a/meta-mender-core/classes/mender-setup-systemd.inc
+++ b/meta-mender-core/classes/mender-setup-systemd.inc
@@ -2,7 +2,6 @@
 # boot process while looking for devices. This is almost always because the
 # kernel feature CONFIG_FHANDLE is not enabled.
 
-DISTRO_FEATURES_BACKFILL:append = "${@mender_feature_is_enabled('mender-systemd', ' systemd', '', d)}"
 DISTRO_FEATURES_BACKFILL_CONSIDERED:append = "${@mender_feature_is_enabled('mender-systemd', ' sysvinit', '', d)}"
 VIRTUAL-RUNTIME_init_manager:mender-systemd = "systemd"
 VIRTUAL-RUNTIME_initscripts:mender-systemd = ""


drewmoseley:
@TheYoctoJester please take a look and let me know your thoughts.  I think the better fix would be to set something like

`INIT_MANAGER:mender-systemd = "systemd"
`

but I've  not had a chance to test that.

TheYoctoJester:
Hi @drewmoseley,

sorry, missed that one. Yes indeed I am gradually switching to `INIT_MANAGER`, so basically I tend to agree. Need to give it a test spin with the override, but if that behaves as expected then I am good with it. Will let you know.

TheYoctoJester:
Hi @drewmoseley,

played around a bit with it now and conclude that the PR is good. Can you please change the description to match the style (e.g. make pass check), then @kacf can merge next week. Thanks!

On a side note, I'll prepare one that uses `INIT_MANAGER` instead, but my understanding is that it should only go into `main`, not `kirkstone`.

kacf (approved):
Looks good to me too! If the CI agree we can merge it.

@drewmoseley: I took the liberty of force pushing your branch to fix the commit issue; it was just a "Ticket" reference missing.

PR #1935, opened 2023-04-13
fix: wic-tools does not need grub when using uboot

zach-welch-aquabyte:
This avoids building grub-related packages when using u-boot, just like what was done for systemd-boot.  In some cases, these dependencies actually interfere with the expected use of u-boot, and this fixes those issues.


kacf (approved):
Thanks! This looks good to me. Will give it a spin in the full CI, but I don't expect any problems.

PR #1932, opened 2023-04-11
feat: Make the build succeed if gateway examples aren't present

kacf:
If you have meta-mender-demo enabled for other reasons than the gateway, the build will fail if you didn't get the examples tar. This makes it a warning not a hard build fail.

Changelog: Title
Ticket: None

Signed-off-by: Alan <alan.martinovic@northern.tech>
(cherry picked from commit e6c0f10394f89840555c479711e779e9cb3429af)

PR #1940, opened 2023-05-01
feat: Make the build succeed if gateway examples aren't present

TheMeaningfulEngineer:
If you have meta-mender-demo enabled for other reasons than the gateway, the build will fail if you didn't get the examples tar. This makes it a warning not a hard build fail.

Changelog: Title
Ticket: None

Cherry pick from [master](https://github.com/mendersoftware/meta-mender/commit/e6c0f10394f89840555c479711e779e9cb3429af)


kacf:
I've actually cherry-picked this [here](https://github.com/mendersoftware/meta-mender/pull/1932) already, but we're having a lot of problems with dunfell, so it's stuck at the moment. Since you need this I'm going to take a calculated risk and merge this without tests. The failures I'm seeing are definitely not related to this code, and it should be pretty safe.

PR #1964, opened 2023-07-04
QA-512: chore: Alternative `extract_fs` script logic with no loopback dev

lluiscampos:
So that the script does not need to be run on a privileged container nor uses loopback devices from the host machine.

### DIFF
--- a/meta-mender-qemu/scripts/docker/extract_fs
+++ b/meta-mender-qemu/scripts/docker/extract_fs
@@ -3,23 +3,21 @@
 set -e
 
 failure() {
-    echo "Failed! Make sure you are using a privileged container." 1>&2
+    echo "Failed!" 1>&2
     cleanup
     exit 1
 }
 
-cleanup() {
-    # Ignore errors in cleanup.
-    set +e
-    kpartx -d $DEV
-    losetup -d $DEV
+trap failure ERR
+
+# Helpers from mender-convert
+disk_get_part_value() {
+    echo "$(partx -o ${3} -g -r --nr ${2} ${1})"
+}
+disk_extract_part() {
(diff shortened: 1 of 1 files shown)


lluiscampos:

meta-mender-qemu/scripts/docker/extract_fs
```diff
@@ -3,23 +3,21 @@
...
-trap cleanup EXIT
 
-DEV=$(losetup -f)
-IMG_PREFIX=${1}
+# Helpers from mender-convert
+disk_get_part_value() {
+    echo "$(partx -o ${3} -g -r --nr ${2} ${1})"
+}
+disk_extract_part() {
+    dd if=$1 of=$4 skip=$2 bs=512 count=$3 status=none
```
kacf, line 18: Although I haven't tested this line specifically, I remember in previous experiments `bs=512` being extremely slow compared to something like `bs=1M`. Might I suggest this instead:

```
dd if=$1 of=$4 skip=${2}b bs=1M count=${3}b status=none iflag=count_bytes,skip_bytes
```

This uses GNU extensions, but I guess it should be alright here. `b` (something confusingly) means 512 multiplicand.

kacf (commented):
Good cleanup!

meta-mender-qemu/scripts/docker/extract_fs
lluiscampos: Thanks! I went ahead in porting this suggestion back to `mender-convert`.

lluiscampos:

kacf (approved):

oleorhagen:
Real nice!

PR #1965, opened 2023-07-25
Add langdale and mickledore compatibility

TheYoctoJester:
This PR collects a number of changes to make the layer buildable with the current Yocto Project upstream.
- adding the LAYERCOMPAT declarations
- removing the e2fsutils patch
- fixing Patch metadata to include the Upstream-Status tag
- refreshing the u-boot patches for Raspberry Pi
- syntax fix for mender-setup.bbclass, this is also already hidden in the UEFI capsule contribution on the master branch.

Changelog: Title
Ticket: None


meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -230,13 +230,7 @@ MENDER_DATA_PART_LABEL ??= "data"
...
 
 IMAGE_INSTALL:append = " mender-client"
-IMAGE_CLASSES += "mender-artifactimg \
-                  mender-artifact-uefi-capsule \
-                  mender-bootimg \
-                  mender-dataimg \
-                  mender-datatar
-                  mender-part-images \
-                  mender-ubimg"
+IMAGE_CLASSES += "mender-artifactimg mender-artifact-uefi-capsule mender-bootimg mender-dataimg mender-datatar mender-part-images mender-ubimg"
```
kacf, line 233: The line breaks were wrong, `mender-datatar` was missing a backslash. Can we still use the line-broken one?

meta-mender-core/recipes-bsp/u-boot/patches/0001-Add-missing-header-which-fails-on-recent-GCC.patch
```diff
@@ -1,14 +1,16 @@
-From d98fa92ea27a6a421310ebe1676c07d34c0c9bc8 Mon Sep 17 00:00:00 2001
+From 2a84a94f31981c00efcdbe4a907bdffd9601caf4 Mon Sep 17 00:00:00 2001
 From: Kristian Amlie <kristian.amlie@mender.io>
 Date: Mon, 15 Feb 2016 15:57:32 +0100
-Subject: [PATCH 1/4] Add missing header which fails on recent GCC.
+Subject: [PATCH] Add missing header which fails on recent GCC.
+
+Upstream-Status: Inappropriate
```
kacf, line 6: What happens if they are not present?

kacf, line 6: Actually, this patch might be applicable. But I'm not even sure it's needed anymore, we could attempt to remove it.

meta-mender-raspberrypi/recipes-bsp/u-boot/patches/experimental/0001-rpi-always-set-fdt_addr-with-firmware-provided-FDT-address.patch
```diff
@@ -0,0 +1,48 @@
+From 226f02f8db76ad7d0431fc10d3f2431c7ce74b06 Mon Sep 17 00:00:00 2001
```
kacf, line 1: I can't see that this patch is listed anywhere. In what context is it to be used?

meta-mender-raspberrypi/recipes-bsp/u-boot/patches/0001-configs-rpi-enable-mender-requirements.patch
```diff
@@ -224,6 +242,46 @@ index 6e24eee55f..e05fbdc7ad 100644
  	default y if ARCH_SUNXI && MMC
  	default y if MMC_OMAP_HS && TI_COMMON_CMD_OPTIONS
  	select FS_FAT
+diff --git a/include/config_mender_defines.h b/include/config_mender_defines.h
```
kacf, line 245: I think you included this file by mistake, and the one below as well, or?

PR #1968, opened 2023-07-27
Add langdale and mickledore compatibility

TheYoctoJester:
This PR collects a number of changes to make the layer buildable with the current Yocto Project upstream.
- adding the LAYERCOMPAT declarations
- removing the e2fsutils patch
- fixing patch metadata to include the Upstream-Status tag
- refreshing the u-boot patches for Raspberry Pi
- syntax fix for mender-setup.bbclass

The PR is a refresh of #1965 

Changelog: Title
Ticket: None


meta-mender-core/recipes-bsp/u-boot/patches/0001-Add-missing-header-which-fails-on-recent-GCC.patch
```diff
@@ -1,23 +0,0 @@
-From 2a84a94f31981c00efcdbe4a907bdffd9601caf4 Mon Sep 17 00:00:00 2001
```
kacf: Good to have this removed, but this got mixed into the wrong commit. There's at least three different commits worth of changes here:

* The original rpi forward port, which is good.
* The removal of the "missing header" patch.
* The removal of the experimental boot counter patch.

Especially the middle one is subject to possible revert later if we discover problems, so I think it's important to keep these changes separate.

kacf (changes requested):
The code is good, but the commits got a bit messed up, see below.

kacf:
@TheYoctoJester: I decided to launch a full pipeline as well. Not to make this PR dependent on it or anything, I'm just curious what we're up against. Pipeline [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/948242192).

kacf:
Well, it [didn't get far](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/4758208146), which is not surprising I guess, but I still was hoping for a bit more.

TheYoctoJester:
> Well, it [didn't get far](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/4758208146), which is not surprising I guess, but I still was hoping for a bit more.

Any particular reason why the pipeline fell over?

kacf:
> > Well, it [didn't get far](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/4758208146), which is not surprising I guess, but I still was hoping for a bit more.
> 
> Any particular reason why the pipeline fell over?

This is the error message, but it doesn't really tell me anything without further investigation. Maybe it rings a bell to you?

```
ERROR: /builds/Northern.tech/Mender/meta/recipes-core/packagegroups/packagegroup-core-boot.bb: Please ensure that your setting of VIRTUAL-RUNTIME_init_manager (systemd) matches the entries enabled in DISTRO_FEATURES
```

PR #1988, opened 2023-10-20
mender-artifactimg: Remove old IMAGE_ID usage

ernstp:
When this variable was removed this place was missed.

Fixes: b23ec02 ("MEN-763: Split and rename artifact-info and trim information.")

### DIFF
--- a/meta-mender-core/classes/mender-artifactimg.bbclass
+++ b/meta-mender-core/classes/mender-artifactimg.bbclass
@@ -135,6 +135,6 @@ IMAGE_CMD:mender () {
         -o ${IMGDEPLOYDIR}/${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.mender
 }
 
-IMAGE_CMD:mender[vardepsexclude] += "IMAGE_ID"
+IMAGE_CMD:mender[vardepsexclude] += "MENDER_ARTIFACT_NAME"
 # We need to have the filesystem image generated already.
 IMAGE_TYPEDEP:mender:append = " ${ARTIFACTIMG_FSTYPE}"


ernstp:
Ok now the commit message was accepted

meta-mender-core/classes/mender-artifactimg.bbclass
```diff
@@ -135,6 +135,6 @@ IMAGE_CMD:mender () {
         -o ${IMGDEPLOYDIR}/${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.mender
 }
 
-IMAGE_CMD:mender[vardepsexclude] += "IMAGE_ID"
+IMAGE_CMD:mender[vardepsexclude] += "MENDER_ARTIFACT_NAME"
```
kacf, line 138: @ernstp: Good catch! However, I think it should just be removed. `MENDER_ARTIFACT_NAME` is not correct, since it does in fact depend on this variable (see line 131).

ernstp: This construct generally exists to say:
We shouldn't invalidate this task (ie. cause a rebuild) _only because the contents of this variable has changed_.

Otherwise, if you have a timestamp in there, you will always rebuild your image even though nothing has changed.

kacf: Yes, but we need to invalidate the task. The artifact image changes every time you change `MENDER_ARTIFACT_NAME`, because it gets embedded in the artifact header.

ernstp: > Yes, but we need to invalidate the task. The artifact image changes every time you change `MENDER_ARTIFACT_NAME`, because it gets embedded in the artifact header.

I think the most common usecase, which I want to use, and I think the original code was intended for is:
Have a dynamic MENDER_ARTIFACT_NAME like "my-artifact-${DATETIME}". We don't want to trigger a new build based on that.
Instead it works the other way around. When something has changed a new artifact will be built, and then it will use whatever value is in MENDER_ARTIFACT_NAME of course. But it won't be the sole reason for a whole new artifact.

kacf: I think this is an interesting use-case, and certainly useful, but I think we cannot apply this logic as is, because it affects everyone, also the ones that are not using `DATETIME`. A different use case which is also quite common is to try to reapply an artifact with the same content, using a different name to do so, in order to fix something on a board which was changed after the first deployment manually, for example during debugging. Not rebuilding anything would just be too surprising behavior.

If your change works for you, I propose you just add it to your own build conf files. And we can even add it to our documentation somewhere in a tips & tricks section, for example. But I don't think it should be the default.

kacf:
Closing, since this won't be merged as is.

PR #1990, opened 2023-10-25
Rework `-dgb` and `-dev` packages

lluiscampos:


meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
```diff
@@ -3,7 +3,9 @@ require mender-client.inc
 DESCRIPTION = "Mender tool for doing OTA software updates."
 HOMEPAGE = "https://mender.io"
 
-PACKAGES =+ "mender-auth mender-update mender-config"
+PACKAGES =+ "mender-auth mender-update mender-config mender-auth-dbg mender-update-dbg"
+
+NOAUTOPACKAGEDEBUG = "1"
```
kacf, line 8: Am I understanding correctly: The reason this is necessary is because we have two package names, not one?

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
```diff
@@ -15,6 +17,8 @@ RRECOMMENDS:mender-update = "mender-flash"
 # back to line based parsing, which is less resilient.
 RRECOMMENDS:mender-update:append = "${@bb.utils.contains('PACKAGECONFIG', 'modules', ' jq', '', d)}"
 
+RDEPENDS:${PN}-dbg:append = " mender-auth-dbg mender-update-dbg"
```
kacf, line 20: So there *is* a `mender-dbg` package? I'm a bit confused, why does it exist if all the files are in the separate dbg packages?

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
```diff
@@ -76,22 +78,17 @@ FILES:mender-update += "\
...
     ${datadir}/mender/identity \
     ${datadir}/mender/identity/mender-device-identity \
-    ${datadir}/dbus-1/interface \
-    ${datadir}/dbus-1/interface/io.mender.Authentication1.xml \
     ${datadir}/dbus-1/system.d/io.mender.AuthenticationManager.conf \
 "
 
 FILES:mender-auth-dev += " \
-    ${datadir}/dbus-1/interface \
+    ${datadir}/dbus-1/interface/io.mender.Authentication1.xml \
```
kacf, line 91: 👍

oleorhagen:
Later, when someone is feeling brave we can do the housecleaning

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
lluiscampos: (replying both of your comments here)

The default `PACKAGES` ([documentation here](https://docs.yoctoproject.org/ref-manual/variables.html#term-PACKAGES)) will include `mender-dbg` which before this PR was owning the debug symbols for both binaries.

Related, without setting `NOAUTOPACKAGEDEBUG` all the debug symbols will end up in the _first_ package with `-dbg` suffix (mender-auth in our case) so I needed to set that one to manually package the symbols as intended.

I discussed it f2f with Ole and decided to keep the symbols in the default `mender-dbg`, which makes sense. Think also that the sources for both packages go into `mender-src`, so it is more consistent to keep both debug and source into the same package prefix.

lluiscampos:
> Ref f2f with @lluiscampos maybe we should just let it default to `mender-{dbg,src,dev}` etc for now.

Well, I left the `-dev` with our custom package. I think there it makes sense (only this specific package has DBus services) and does not conflict with the fact that all sources go into the same `mender-src` package.

oleorhagen (approved):

PR #2002, opened 2023-11-13
feat: Add a pkcs11-provider recipe

oleorhagen:
Ticket: MEN-6839
Changelog: None

### DIFF
--- a/meta-mender-demo/recipes-connectivity/pkcs11-provider/pkcs11-provider.bb
+++ b/meta-mender-demo/recipes-connectivity/pkcs11-provider/pkcs11-provider.bb
@@ -1,0 +1,17 @@
+
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://COPYING;md5=b53b787444a60266932bd270d1cf2d45 \
+                    file://LICENSES/Apache-2.0.txt;md5=3b83ef96387f14655fc854ddc3c6bd57"
+
+SRC_URI = "git://github.com/latchset/pkcs11-provider;protocol=https;branch=main"
+
+PV = "0.2+git${SRCPV}"
+SRCREV = "d994cc26790d31242fcee5df7d07d59f1698f073"
+
+S = "${WORKDIR}/git"
+
+DEPENDS:append = " openssl p11-kit autoconf-archive"
+
+inherit pkgconfig autotools
+
+FILES:${PN} =+ "${libdir}/ossl-modules/pkcs11.so"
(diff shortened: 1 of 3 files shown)


oleorhagen:
Note to self: fix merged upstream

meta-mender-demo/recipes-mender/pkcs11-provider/pkcs11-provider.bb
```diff
@@ -0,0 +1,17 @@
+
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://COPYING;md5=b53b787444a60266932bd270d1cf2d45 \
+                    file://LICENSES/Apache-2.0.txt;md5=3b83ef96387f14655fc854ddc3c6bd57"
+
+SRC_URI = "git://github.com/oleorhagen/pkcs11-provider;protocol=https;branch=main"
```
kacf, line 6: Shouldn't this be forked into `mendersoftware`?

meta-mender-demo/recipes-mender/pkcs11-provider/pkcs11-provider.bb
kacf, line 1: I'd put this file under `recipes-connectivity`.

meta-mender-demo/recipes-mender/pkcs11-provider/pkcs11-provider.bb
```diff
@@ -0,0 +1,17 @@
...
+
+PV = "0.2+git${SRCPV}"
+SRCREV = "${AUTOREV}"
```
kacf, line 9: `AUTOREV` should not be used in permanent recipes, because it causes all users to probe the repositories on every build, so it's a performance problem. We should set the SHA instead.

oleorhagen: :+1:

meta-mender-demo/recipes-mender/pkcs11-provider/pkcs11-provider.bb
kacf: If we can, yes definitely.

kacf (approved):

lluiscampos (approved):
Looks good.

Consider amending your commit message: "There was no really good place of putting this recipe - so I chose to go with demo in this instance."

oleorhagen:
> Looks good.
> 
> Consider amending your commit message: "There was no really good place of putting this recipe - so I chose to go with demo in this instance."

fixed the `to my local repo` commit message

oleorhagen:
Closing, as this is blocked by QA-587 https://northerntech.atlassian.net/browse/QA-587

PR #2005, opened 2023-11-22
Various acceptance test fixes

kacf:


kacf:
@lluiscampos: I'm thinking that we stick to removals here in the branch which eventually goes to master. But once we port to kirkstone, we skip the ones that are removed instead, in order to still keep coverage for the Golang client around for a while.

tests/acceptance/test_snapshot.py
```diff
@@ -35,6 +35,9 @@ def get_ssh_args_mender_artifact(conn):
     return "-S " + common_args
 
 
+@pytest.mark.skipif(
+    not (os.environ.get("NIGHTLY_BUILD", "false") == "true"), reason="MEN-6861",
+)
```
lluiscampos, line 40: Nice :+1:

tests/acceptance/test_bootstrap_artifact.py
```diff
@@ -62,7 +62,7 @@ def img_builder():
                 'MENDER_ARTIFACT_VERIFY_KEY = "%s"'
                 % os.path.join(os.getcwd(), signing_key("RSA").public),
                 'SYSTEMD_AUTO_ENABLE:pn-mender-client = "enable"',
-                'SYSTEMD_AUTO_ENABLE:pn-mender-update = "enable"',
+                'SYSTEMD_AUTO_ENABLE:pn-mender = "enable"',
```
oleorhagen, line 65: Q: Are you sure this will work with the new client? I think this is just a dud in this instance.

tests/acceptance/test_build.py
```diff
@@ -1307,43 +1312,34 @@ def test_mender_dbus_interface_file(
...
             "io.mender.Authentication1.xml",
-            "io.mender.Update1.xml",
         ]
+        mender_recipe = "mender"
+        mender_pkg = "mender-auth"
+
+        # Can be removed after mender-client < v4.0 goes EOL. Update Control is not available
+        # anymore in v4.0 and later.
+        if mender_update_binary == "mender":
+            EXPECTED_FILES += [
```
oleorhagen, line 1338: Nit: Why not append? :thinking:

oleorhagen (approved):

kacf:
> You can also remove these lines from `mender-client-cpp.inc`:
> 
> ```
>     # Add the update control configurations if set
>     conf_maybe_add("UpdateControlMapExpirationTimeSeconds", "MENDER_UPDATE_CONTROL_MAP_EXPIRATION_TIME_SECONDS", getvar=True, integer=True)
>     conf_maybe_add("UpdateControlMapBootExpirationTimeSeconds", "MENDER_UPDATE_CONTROL_MAP_BOOT_EXPIRATION_TIME_SECONDS", getvar=True, integer=True)
> ```

I'll follow up on this in the next PR.

PR #2009, opened 2023-11-28
Add the mender-snapshot recipe

oleorhagen:
Adds the `mender-snapshot` git recipe

Ticket: MEN-6861

### DIFF
--- a/meta-mender-core/recipes-mender/mender-snapshot/mender-snapshot.inc
+++ b/meta-mender-core/recipes-mender/mender-snapshot/mender-snapshot.inc
@@ -1,0 +1,29 @@
+DESCRIPTION = "Mender tool for creating golden-images from a given installation."
+HOMEPAGE = "https://mender.io"
+
+DEPENDS:append = " glib-2.0"
+RDEPENDS:${PN} = "glib-2.0 mender-update (>= 4.0)"
+
+B = "${WORKDIR}/build"
+
+inherit go-mod
+inherit go-ptest
+
+GO_IMPORT = "github.com/mendersoftware/mender-snapshot"
+
+do_compile() {
+    oe_runmake \
+        -C ${B}/src/${GO_IMPORT} \
+        V=1
+}
+
+do_install() {
+    oe_runmake \
+        -C ${B}/src/${GO_IMPORT} \
+        V=1 \
+        prefix=${D} \
+        bindir=${bindir} \
+        datadir=${datadir} \
+        sysconfdir=${sysconfdir} \
+        install-bin
+}
--- a/meta-mender-core/recipes-mender/mender-snapshot/mender-snapshot_git.bb
+++ b/meta-mender-core/recipes-mender/mender-snapshot/mender-snapshot_git.bb
@@ -1,0 +1,67 @@
+require mender-snapshot.inc
+
+# The revision listed below is not really important, it's just a way to avoid
+# network probing during parsing if we are not gonna build the git version
+# anyway. If git version is enabled, the AUTOREV will be chosen instead of the
+# SHA.
+def mender_snapshot_autorev_if_git_version(d):
+    version = d.getVar("PREFERRED_VERSION")
+    if not version:
+        version = d.getVar("PREFERRED_VERSION_%s" % d.getVar('PN'))
+    if not d.getVar("EXTERNALSRC") and version is not None and "git" in version:
+        return d.getVar("AUTOREV")
+    else:
+        return "c0df5df2be6a6acba4b7a55c3e64f74e0695a6f6"
+SRCREV ?= '${@mender_snapshot_autorev_if_git_version(d)}'
+
+def mender_snapshot_branch_from_preferred_version(d):
+    import re
+    version = d.getVar("PREFERRED_VERSION")
+    if version is None or version == "":
+        version = d.getVar("PREFERRED_VERSION_%s" % d.getVar('PN'))
+    if version is None:
+        version = ""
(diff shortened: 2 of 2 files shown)


meta-mender-core/recipes-mender/mender-snapshot/mender-snapshot_git.bb
```diff
@@ -0,0 +1,67 @@
...
+# anyway. If git version is enabled, the AUTOREV will be chosen instead of the
+# SHA.
+def mender_snapshot_autorev_if_git_version(d):
+    version = d.getVar("PREFERRED_VERSION")
+    if not version:
+        version = d.getVar("PREFERRED_VERSION_%s" % d.getVar('PN'))
+    if not d.getVar("EXTERNALSRC") and version is not None and "git" in version:
+        return d.getVar("AUTOREV")
+    else:
+        return "c0c410aede66e31cd75c65e1fcb6df0b3ad91367"
```
oleorhagen, line 14: TODO - Needs to be updated after https://github.com/mendersoftware/mender-snapshot/pull/10 is merged

kacf (approved):

meta-mender-core/recipes-mender/mender-snapshot/mender-snapshot.inc
```diff
@@ -0,0 +1,29 @@
+DESCRIPTION = "Mender tool for creating golden-images from a given installation."
+HOMEPAGE = "https://mender.io"
+
+DEPENDS:append = " glib-2.0"
+RDEPENDS:${PN} = "glib-2.0 mender-client (>= 4.0)"
```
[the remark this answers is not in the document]
oleorhagen: Hmm, no. I don't think so :thinking: 

We set the `virtual-provider` to `mender-client` no?

meta-mender-core/recipes-mender/mender-snapshot/mender-snapshot_git.bb
oleorhagen: But isn't it nice if it's something which builds? :thinking:

meta-mender-core/recipes-mender/mender-snapshot/mender-snapshot.inc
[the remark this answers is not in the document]
kacf: No, this was removed in 911ea43605a7159aa903dbbf6fbddf6e41e886d2. It didn't work correctly because it was a build-time only provide, and it seemed better to fix it properly by removing it instead of trying to add an additional, ambiguous runtime provide. In Yocto backwards package compatibility is not very important since you won't upgrade one package at a time, like for Debian.

Possibly we might choose a different approach on kirkstone, but let's cross that bridge when we get there, we should stick to the new names in master.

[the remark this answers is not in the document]
oleorhagen: Hmm, kk. I see. Why did we decide on doing this breaking change? :thinking:

meta-mender-core/recipes-mender/mender-snapshot/mender-snapshot_git.bb
kacf: The only cases where this SHA will be selected are:
* If you have no `PREFERRED_VERSION` set, in which case `DEFAULT_PREFERENCE` will prevent this recipe from being built.
* If `EXTERNALSRC` is set, in which case that source code will be used instead of whatever is in the SHA.

So it is never actually *used*. However, the Git fetcher still runs in the background, and it's both slower and "more rude" because it then connects on every build. If there is a valid SHA, however, it never hits the network.

So actually, it would be better if it didn't build, because then there is a mistake in the logic somewhere! 😄

meta-mender-core/recipes-mender/mender-snapshot/mender-snapshot.inc
```diff
@@ -0,0 +1,29 @@
...
+
+DEPENDS:append = " glib-2.0"
+RDEPENDS:${PN} = "glib-2.0 mender-update (>= 4.0)"
+
+B = "${WORKDIR}/build"
+
+inherit go-mod
+inherit go-ptest
+
+GO_IMPORT = "github.com/mendersoftware/mender-snapshot"
```
lluiscampos, line 12: Once we are using `go-mod`, can't we get the rid of this `GO_IMPORT` path? Am I right that the only reason to keep this is to make it compatible with today's mender-qa script (that checks out everything under `src/github.com/mendersoftware`)?

meta-mender-core/recipes-mender/mender-snapshot/mender-snapshot_git.bb
lluiscampos: Well, not setting `PREFERRED_VERSION` while still not having a final tagged recipe will use this SHA, for which case is good that the recipe actually builds.

This situation for example was an issue for me when I started building `mender-flash` with yocto.

lluiscampos (approved):

meta-mender-core/recipes-mender/mender-snapshot/mender-snapshot.inc
oleorhagen: From what I can tell, we still can't actually. It uses it for cloning.

It was a bit surprising.

meta-mender-core/recipes-mender/mender-snapshot/mender-snapshot_git.bb
oleorhagen: Updated to the first building version

PR #2016, opened 2023-12-06
Mender C++ client support for kirkstone

kacf:


meta-mender-commercial/conditional/mender-monitor/mender-monitor.inc
```diff
@@ -2,7 +2,7 @@ inherit mender-licensing
 
 inherit systemd
 
-RDEPENDS:${PN} = "bash mender-client (>= 2.5) lmdb"
+RDEPENDS:${PN} = "bash mender-auth lmdb"
```
lluiscampos, line 5: Modifying this here in the `.inc` file, and similarly for `mender-gateway`, `mender-connect`, `mender-configure` (possibly others), it will misbehave for the tagged recipes. The best would be to override `RDEPENDS:${PN}` in the recpies for the already released software.

Edit: looking at your last commit, I can see that this is intentional from your end. I think it is incorrect:  an already released addons should not depend on the latest client (specially when the default is still the old one!), rather it should depend on any client. It should be `mender-auth | mender-client (>= 2.5)`, somehow.

lluiscampos (changes requested):
Looks good. The main commit was hard to review, I skimmed many parts. See a comment below for something that we should fix.

Also, from the changelogs:
* "...recipe creates two packages. mender-auth and mender-update to provide
the C++ implementations..." -> " ...creates two packages, mender-auth and mender-update, to provide..."
* The `refac:` prefix should have been removed by the changelog generator tool. In the commit message, use `Changelog(refac)` instead of `Changelog: refac`

meta-mender-commercial/conditional/mender-monitor/mender-monitor.inc
kacf: Yes, but this should be fine due to the [`RPROVIDES` of the mender-client recipe](https://github.com/mendersoftware/meta-mender/blob/4676d0c25750067a77de7adf14f25bf87da64d1e/meta-mender-core/recipes-mender/mender-client/mender-client-go.inc#L19).

lluiscampos (approved):

kacf:
Changelog fixed.

[This](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1100029675) is the last run, I'm not re-running since the last push only fixed changelogs.

PR #2025, opened 2023-12-12
refac: mender-client: Move DBus xml files to dev package.

kacf:
They do not influence production, and thus they belong in the dev package.

Changelog: Commit

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client-go.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client-go.inc
@@ -120,15 +120,11 @@ FILES:${PN} += "\
     ${_MENDER_PERSISTENT_DIR}/mender.conf \
 "
 
-FILES:${PN} += "\
-    /usr/share/dbus-1/interface \
+
+FILES:${PN}-dev:append = " \
+    ${datadir}/dbus-1/interface \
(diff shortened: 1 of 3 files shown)


kacf:
The file has already moved in the C++ recipe, and this fixes 3.7.x, together with the Golang recipe, since we are now testing unconditionally for the file in the dev package.

meta-mender-core/recipes-mender/mender-client/mender-client-go.inc
```diff
@@ -120,15 +120,11 @@ FILES:${PN} += "\
...
 
-FILES:${PN} += "\
-    /usr/share/dbus-1/interface \
-    /usr/share/dbus-1/interface/io.mender.Update1.xml \
-    /usr/share/dbus-1/interface/io.mender.Authentication1.xml \
-"
 
-
-FILES:${PN}-dev += " \
+FILES:${PN}-dev:append = " \
```
kacf, line 124: I don't know why `append` expands correctly, and `+=` doesn't. It's some deep logic in poky's `renameVar`. It appears to be specific to the `FILES:${PN}-dev` variable.

lluiscampos (approved):

oleorhagen (approved):

PR #2027, opened 2023-12-14
C++ client support for kirkstone

kacf:


kacf:
* [3.7.x pipeline](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1106685318)
* [master pipeline](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1106685498) (I expect this one to fail because it will expect C++ and it's still Golang)

joaoLifepowr:
I've recently noticed that some of the classes are still not up-to-date with the effort of changing `mender-client-install` to `mender-update-install`.
One case is the `mender-full` class, which is good in master, but not in kirkstone:
https://github.com/mendersoftware/meta-mender/blob/kirkstone/meta-mender-core/classes/mender-full.bbclass

We ended up by manually forcing the `mender-update-install` feature on our confs, but it may silently affect others.
This feature is particularly important as it enables the dynamic bootloader target partition, instead of fixing it to /dev/mmcblk0p2.

kacf:
Hey @joaoLifepowr, the difference you are describing is intentional. It is to avoid breaking builds that are already specifying `mender-client-install` manually. The `mender-update-install` feature should already be automatically handled [here](https://github.com/mendersoftware/meta-mender/blob/kirkstone/meta-mender-core/classes/mender-maybe-setup.bbclass#L23-L27). Are you saying this is not working correctly?

joaoLifepowr:
Hey @kacf,
It was not working in our setup, our local.conf only inherits the mender-full.bbclass.
It only worked when I manually added the `mender-update-install` to the MENDER_FEATURES_ENABLE var.

I was looking at the mender-full.bbclass, it inherits mender-setup, which inherits mender-helpers and the mender-setup-*** files, none of these seem to be inheriting mender-maybe-setup. That could be the reason why it is not being handled in our case.

kacf:
@joaoLifepowr

mender-maybe-setup is being unconditionally inherited [here](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/conf/layer.conf#L18), so it should be hard to miss. I wonder if it could be an ordering issue, although bitbake usually warns about that. Unfortunately I cannot reproduce your issue, so that's all the info I have.

joaoLifepowr:
Thank you for checking @kacf,
It probably is some issue on our side, since nobody else reported it

PR #2029, opened 2023-12-14
fix: Amend D-Bus interface files path for the C++ client

lluiscampos:
It should be `interfaces` according to:
* https://dbus.freedesktop.org/doc/dbus-api-design.html


kacf (approved):
Looks good, but can you cherry-pick this commit as well: https://github.com/kacf/meta-mender/commit/ca5d20be3445e3165891fe9f80af43e120eb1fac

PR #2051, opened 2024-01-04
fix mender-client 3.5.2 recipe

PreetiSachan:
Replaced mender-client.inc with mender-client-go.inc

Signed-off-by: Preeti Sachan <preeti.sachan@intel.com>


kacf:
Thanks @PreetiSachan, looks like this slipped through the cracks. This is handled in #2053, so closing this one.

PR #2054, opened 2024-01-05
fix(mender): Fix incorrect licenses and license checksums.

kacf:
The licenses of the C++ client was wrong, and the checksum for the Golang client when building with Git was also wrong. Fix both, and consolidate licenses in one file.

Changelog: None
Ticket: None

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client_git.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client_git.inc
@@ -70,15 +70,25 @@ def mender_version_from_preferred_version(d):
 def mender_license(branch):
     import re
     if branch.startswith("1."):
+        # Golang client
         return {
-                   "license": "Apache-2.0 & BSD-2-Clause & BSD-3-Clause & ISC & MIT & OLDAP-2.8",
+            "license": "Apache-2.0 & BSD-2-Clause & BSD-3-Clause & ISC & MIT & OLDAP-2.8",
+            "license_file": "file://src/github.com/mendersoftware/mender/LICENSE",
+        }
+    elif branch.startswith("2.") or branch.startswith("3."):
+        # Golang client
+        return {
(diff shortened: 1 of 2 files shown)


meta-mender-core/recipes-mender/mender-client/mender-client_git.inc
```diff
@@ -70,15 +70,25 @@ def mender_version_from_preferred_version(d):
...
     else:
+        # C++ client
         return {
-                   "license": "Apache-2.0 & BSD-2-Clause & BSD-3-Clause & ISC & MIT & OLDAP-2.8",
+            "license": "Apache-2.0 & MIT & Unlicense",
+            "license_file": "file://LICENSE",
         }
 LIC_FILES_CHKSUM = " \
-    file://src/github.com/mendersoftware/mender/LICENSE;md5=b4b4cfdaea6d61aa5793b92efd42e081 \
+    ${@mender_license(d.getVar('MENDER_BRANCH'))['license_file']};md5=30b4554c64108561c0cb1c57e8a044f0 \
```
oleorhagen, line 91: Q: Have we updated the license for the older clients?

oleorhagen (approved):

meta-mender-core/recipes-mender/mender-client/mender-client_git.inc
kacf: I'm doing it now:

* https://github.com/mendersoftware/mender-artifact/pull/580
* https://github.com/mendersoftware/mender/pull/1550
* https://github.com/mendersoftware/mender-connect/pull/127

kacf:
The 3.7.x pipeline is not completely done, but I'll merge now, because I'm going home now, and these changes are needed to fix kirkstone, which is broken for the public. Master has already passed, both here and in kirkstone. I have high confidence that the rest of 3.7.x will pass, but if it's still broken I will fix it first thing on Monday morning.

PR #2056, opened 2024-01-08
master: Mender 3.7.1 release (with C++ client)

lluiscampos:


meta-mender-core/recipes-mender/mender-client/mender_4.0.0.bb
```diff
@@ -0,0 +1,37 @@
...
+
+SRC_URI = "gitsm://github.com/mendersoftware/mender;protocol=https;branch=4.0.x"
+
+# Tag: 4.0.0
+SRCREV = "146951aa650064af0c28b1e42eb729bcd7aa9155"
+
+# Enable this in Betas, and in branches that cannot carry this major version as
+# default.
+# Downprioritize this recipe in version selections.
+#DEFAULT_PREFERENCE = "-1"
```
kacf, line 19: Maybe you know this already, but just for future reference: We don't need to change this in kirkstone, because we are using [this](https://github.com/mendersoftware/meta-mender/blob/kirkstone/meta-mender-core/classes/mender-setup.bbclass#L232-L234) to select the correct version. I would still like to make a final manual check though. Ping me if you don't have a build handy (I do).

kacf (approved):

meta-mender-core/recipes-mender/mender-client/mender_4.0.0.bb
lluiscampos: I verified this now with https://github.com/mendersoftware/meta-mender/pull/2063 :+1:

PR #2059, opened 2024-01-09
feat: Add `MENDER_WRITING_TOOL` variable to mender-binary-delta.

kacf:
This has no effect for mender-binary-delta versions older than 1.5.

Changelog: Commit
Ticket: MEN-6774


meta-mender-commercial/recipes-mender/mender-binary-delta/mender-binary-delta_1.4.0.bb
```diff
@@ -19,3 +19,6 @@ LIC_FILES_CHKSUM = " \
 # default.
 # Downprioritize this recipe in version selections.
 #DEFAULT_PREFERENCE = "-1"
+
+# Not used in this version.
+_MENDER_WRITING_TOOL_DEFAULT = ""
```
kacf, line 24: https://northerntech.atlassian.net/browse/MEN-6938 exists to be sure we remove this from future versions.

oleorhagen (approved):
Looks good. `Writing tool` sounds a little odd to me, but then again. Can't thihnk of many better options.

I think I _might_ prefer `Flash tool` but YMMV

kacf:
That's a good point. I just discovered a different issue as well, so I'll fix it.

lluiscampos (approved):

PR #2060, opened 2024-01-11
Add meta-mender support for the watchdog functionality in the C++ mender-client

oleorhagen:
I added some simple functionality for handling watchdog timeout support in meta-mender (configuring it).

Through the `MENDER_WATCHDOG_SEC` variable.

Also added a simple acceptance test for the watchdog functionality.


meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
```diff
@@ -344,6 +347,14 @@ do_install() {
     DESTDIR='${D}' cmake_runcmake_build --target install-dbus
 }
 
+do_install:append() {
+    sed -i 's/^[#[:space:]]*PasswordAuthentication.*/PasswordAuthentication no/' ${D}${sysconfdir}/ssh/sshd_config
```
kacf, line 351: This is your own debugging code, isn't it? 🙂

tests/acceptance/test_watchdog.py
```diff
@@ -0,0 +1,158 @@
...
+        config["InventoryPollIntervalSeconds"] = 5
+        config["UpdatePollIntervalSeconds"] = 5
+        config["RetryPollIntervalSeconds"] = 2
+
+        with tempfile.NamedTemporaryFile() as mender_conf:
+            mender_conf.write(mender_conf_contents)
+            put_no_sftp(
+                mender_conf.name,
+                connection,
+                remote="/etc/mender/mender.conf",
```
kacf, line 132: Isn't the image R/O by default? I think you can modify `/data/etc/mender/mender.conf` though (you should double check that path), which the path above points to.

kacf, line 132: Client acceptance tests share QEMU instances, so the original file needs to be restored afterwards.

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
```diff
@@ -14,6 +14,9 @@ MENDER_UPDATE_POLL_INTERVAL_SECONDS ?= "1800"
 MENDER_INVENTORY_POLL_INTERVAL_SECONDS ?= "28800"
 MENDER_RETRY_POLL_INTERVAL_SECONDS ?= "300"
 
+_MENDER_WATCHDOG_DEFAULT_TIMEOUT_SEC := 86400 # 24 hours
+MENDER_WATCHDOG_SEC ?= "${@max(d.getVar('MENDER_UPDATE_POLL_INTERVAL_SECONDS'), d.getVar('_MENDER_WATCHDOG_DEFAULT_TIMEOUT_SEC')}"
```
kacf, line 18: This probably needs to be added to `meta-mender-core/conf/mender-vars.json`.

kacf (changes requested):
Nice test

tests/acceptance/test_watchdog.py
oleorhagen: Hmm, I don't think they are? :thinking: 

This works at least

oleorhagen: Nope, I just had a look at a few of the other tests. The `put_no_sftp` is pretty much copy paste from `test_update`

oleorhagen: Hmm.. Seems ur right with the build conf. Then what is going on with the other tests?

kacf (approved):
Nice solution!

PR #2062, opened 2024-01-15
feat: Add `MENDER_FLASH_TOOL` variable to mender-binary-delta.

kacf:
This has no effect for mender-binary-delta versions older than 1.5.

Changelog: Commit
Ticket: MEN-6774

(cherry picked from commit 5e31f7eb2789291c6a26e6670018e5062476c8aa)


kacf:
@TheYoctoJester: The mender-configure fix is appropriate for Kirkstone too, right? (I think it is, but just checking)

PR #2065, opened 2024-01-17
fix: Fix naming mismatch for mender-grow-data.service.

kacf:
No changelog, since we never tagged the broken version.

Changelog: None
Ticket: None


oleorhagen (approved):

kacf:
Known failure, which is also in master. Merging.

PR #2078, opened 2024-02-21
fix: Don't use `MENDER_FEATURES_ENABLE` variable directly.

kacf:
We should use `MENDER_FEATURES` because this is the final result of calculating the features, otherwise we can miss feature toggles that have special handling.

This fixes an issue where `mender-update-install` was not added correctly, leading to boot issues on RPi.

No changelog for this, because it was never in a tagged release.

Changelog: None
Ticket: MEN-7039


TheYoctoJester (approved):
Gave it a functional test and it works for me. Code looks right at a first glance, but I'm not exactly competent to fully judge it.

PR #2080, opened 2024-02-26
test: updated test to be compatible with backoff in mender-connect

danielskinstad:
Changed error check in test_mender-connect to check the new error message in connectiomanager.
https://github.com/mendersoftware/mender-connect/pull/118/commits/9b7def234b9bfeff046166dbc8ca11c20186fdf3

Ticket: MEN-5782

### DIFF
--- a/tests/acceptance/test_mender-connect.py
+++ b/tests/acceptance/test_mender-connect.py
@@ -23,6 +23,7 @@ from utils.common import (
     cleanup_mender_state,
     put_no_sftp,
     qemu_system_time,
+    version_is_minimum,
 )
 
 
@@ -271,7 +272,7 @@ class TestMenderConnect:
 
     @pytest.mark.min_mender_version("2.5.0")
     def test_mender_connect_reconnect(
-        self, request, connection, with_mock_files, with_mock_servers,
+        self, request, connection, with_mock_files, with_mock_servers, bitbake_variables
     ):
         """Test that mender-connect can re-establish the connection on remote errors"""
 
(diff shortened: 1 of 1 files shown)


tests/acceptance/test_mender-connect.py
```diff
@@ -286,10 +286,7 @@ def test_mender_connect_reconnect(
 
             # wait for error
             _ = wait_for_string_in_log(
-                connection,
-                startup_time,
-                300,
-                "eventLoop: error reconnecting: failed to connect after max number of retries",
+                connection, startup_time, 300, "connection manager failed to connect",
```
lluiscampos, line 289: This repository is versioned on a different schedule than client components, meaning that  a give version of `meta-mender` needs to be able to build and test several versions of `mender-connect` for example. So we need modify the test to be able to verify either version.

It is the first time for you encountering this problem, but it could be something like this:
```
            # wait for error
            if version_is_minimum(bitbake_variables, "mender-connect", "2.3.0"):
                error_message = "connection manager failed to connect"
            else:
                error_message = "eventLoop: error reconnecting: failed to connect after max number of retries"
            _ = wait_for_string_in_log(connection, startup_time, 300, error_message)
```

Plus adding `bitbake_variables` to the fixtures to be used in this test and importing `version_is_minimum`.

lluiscampos (changes requested):
Sorry, we need to make this test _also_ work for already released versions. See my comment below.

Once fixed as suggested, run the test locally once for each versions (latest release and your local code) and we don't need to re-run the full pipeline again :+1:

lluiscampos (approved):

lluiscampos:
> Once fixed as suggested, run the test locally once for each versions (latest release and your local code)

Tested locally by @danielskinstad

PR #2090, opened 2024-04-08
fix: Remove metadata_csum fix which is not needed anymore.

kacf:
It also breaks for some systems apparently.

Changelog: Fix rare mounting errors like these:
```
kernel: EXT4-fs error (device mmcblk0p6): ext4_mb_generate_buddy:1141: group 35, block bitmap and bg descriptor inconsistent: 32766 vs 32768 free clusters
```

Ticket: MEN-6027


lluiscampos (approved):
Based on the Git history of this fix, it is about time to remove it :+1:

kacf:
@lluiscampos: Can you recheck? This required an additional fix I did not anticipate.

kacf:
@lluiscampos: Ping

lluiscampos (approved):
Sorry that I missed the first ping. It looks good to me.

Additionally, based on your latest findings, should we remove [this warning](https://github.com/mendersoftware/mender-convert/blob/989d7df4b3f58aa20381a4ed07da914552393903/mender-convert-modify#L203)?

kacf:
Good call, added here: https://github.com/mendersoftware/mender-convert/pull/659

PR #2093, opened 2024-04-17
fix: Set SYSTEMD_UNIT_DIR correctly in mender-client

PreetiSachan:
Setting hardcoded path "/lib/systemd/system" to SYSTEMD_UNIT_DIR variable, is causing following failure: 
| do_package: Didn't find service unit 'mender-updated.service', specified in SYSTEMD_SERVICE:mender-update

Set SYSTEMD_UNIT_DIR to bitbake env variable ${systemd_system_unitdir}. 
When usrmerge DISTRO_FEATURE is enabled, ${root_prefix} points to ${exec_prefix} otherwise to ${base_prefix}.

Ref:
https://git.openembedded.org/openembedded-core/commit/?id=700848c6ebd03bf3105d09a41d758883ab875618

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/files/0001-CMakeLists.txt-fix-SYSTEMD_UNIT_DIR.patch
+++ b/meta-mender-core/recipes-mender/mender-client/files/0001-CMakeLists.txt-fix-SYSTEMD_UNIT_DIR.patch
@@ -1,0 +1,38 @@
+From f97b9a680230a7539acf371ac0bce22c6473246f Mon Sep 17 00:00:00 2001
+From: Preeti Sachan <preeti.sachan@intel.com>
+Date: Wed, 17 Apr 2024 12:44:37 +0530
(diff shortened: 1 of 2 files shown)


PreetiSachan:
PR is verified with kikrstone and scarthgap build.

ColinFinck:
I did something similar in https://github.com/enlyze/meta-mender/commit/eec8ac0a2cb1bc7608ff3738197d82a79459db83, however my fix is slightly more complex.

Your way of setting `SYSTEMD_UNIT_DIR` to `$ENV{systemd_system_unitdir}` sounds future-proof. I approve it (but I have no push rights in this repo)

kacf:
I've created https://github.com/mendersoftware/mender/pull/1617. Once that is merged, we can replace this with a variable setting instead of a patch, which is much cleaner. This also avoids adding "dirty" to the version string of the client because we have added a patch.

kacf:
Superseded by #2113.

That fix won't work for existing client versions, but will work for all versions released from now on.

PR #2117, opened 2024-05-22
fix: add coreutils as a runtime dependency for Mender monitor

TheMeaningfulEngineer:
If the image is built with a busybox version of `df` Mender monitor will start giving errors for the disk usage case:

Setting a runtime dependency to coreutils provides a real version of df instead of a busybox one.

Ticket: None
Changelog: Title


kacf (approved):
Will start a mender-qa build for this.

PR #2121, opened 2024-06-01
`dunfell` update v2024.06 - part1: Mender recipes

lluiscampos:

PR #2125, opened 2024-06-04
`dunfell` update v2024.06

lluiscampos:
The full update. It includes #2121.


meta-mender-commercial/recipes-mender/mender-binary-delta/mender-binary-delta.inc
```diff
@@ -36,14 +49,22 @@ do_version_check() {
 addtask do_version_check after do_patch before do_install
 
 do_configure() {
-    # We need the sed operation because bitbake misinterprets the JSON closing
-    # brace as end of function, unless we put spaces in front of it.
-    sed -e 's/^ //' <<EOF | cat > ${WORKDIR}/mender-binary-delta.conf
- {
+    echo "{" > ${WORKDIR}/mender-binary-delta.conf
+
+    if [ -n "${MENDER_FLASH_TOOL}" ]; then
```
kacf, line 54: Oh darn! I just realized this is wrong in both branches (master too)!

It needs to check the value, and set either `mender-flash` or `ubiupdatevol`, not just `mender-flash` unconditionally!

Could you fix both?

meta-mender-core/classes/mender-setup.bbclass
```diff
@@ -244,10 +244,10 @@ python() {
 addhandler mender_sanity_handler
 mender_sanity_handler[eventmask] = "bb.event.ParseCompleted"
 python mender_sanity_handler() {
-    if bb.utils.contains('MENDER_FEATURES_ENABLE', 'mender-partuuid', True, False, d) and d.getVar('MENDER_STORAGE_DEVICE', True) != "":
+    if bb.utils.contains('MENDER_FEATURES', 'mender-partuuid', True, False, d) and d.getVar('MENDER_STORAGE_DEVICE', True) != "":
         bb.warn("MENDER_STORAGE_DEVICE is ignored when mender-partuuid is enabled. Clear MENDER_STORAGE_DEVICE to remove this warning.")
 
-    if bb.utils.contains('MENDER_FEATURES_ENABLE', 'mender-partuuid', True, False, d) and bb.utils.contains('MENDER_FEATURES_ENABLE', 'mender-uboot', True, False, d):
+    if bb.utils.contains('MENDER_FEATURES', 'mender-partuuid', True, False, d) and bb.utils.contains('MENDER_FEATURES', 'mender-uboot', True, False, d):
```
kacf, line 250: Careful! `MENDER_FEATURES_ENABLE` end up in `DISTRO_FEATURES` in dunfell and older. From grepping I can see that there are already a couple of mistakes in dunfell regarding this, notably when handling `mender-partuuid`.

kacf (approved):
Pre-approving, I don't think it needs re-reviewing. But there are a couple things to fix below.

meta-mender-commercial/recipes-mender/mender-binary-delta/mender-binary-delta.inc
lluiscampos: Cherry-picked now :+1:

meta-mender-core/classes/mender-setup.bbclass
lluiscampos: Thanks.

Fixed this on the cherry-pick and also added a [separate commit for the existing broken feature](https://github.com/mendersoftware/meta-mender/pull/2125/commits/95f38e932a3e44b44b9d357f705ac087d617a765). Can you please give feedback on that commit? Is the changelog too much and we could do `Title` instead?

kacf: I think I'd go for "title", the full message is a bit much IMHO. Looks good otherwise!

lluiscampos:
The last force-push only changed a commit message. The previous pipeline (no code changes) can be found [here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1320969046). Merging :rocket:

PR #2126, opened 2024-06-05
fix: Fix `mender-binary-delta` configuration

lluiscampos:
Amends commit: 5e31f7eb2789291c6a26e6670018e5062476c8aa

Ticket: None
Changelog: Fix autogeneration of `mender-binary-delta.conf` so that the user value for `MENDER_FLASH` is respected, instead of the previously wrongly hard-coded value of `mender-flash`. Without this fix `mender-binary-delta` won't still work on UBI file systems.


kacf (approved):

lluiscampos:
* kirkstone

PR #2132, opened 2024-06-13
feat: Allow passing parameters to bootstrap artifact only

LorenzoRuffati:
As of right now the `mender-artifactimg.bbclass` and `mender-bootstrap-artifact.bbclass` use the same variables to build the parameters passed to mender-artifact to build the respective artifacts. This does not work properly in cases where the yocto build is composed of multiple artifacts (for instance when some part of the data partition is meant to be updated separately with an update module), since you have to choose between the bootstrap artifact not having information on this second component, or have the rootfs artifact also have the provides for the other module (despite not necessarily overwriting the data on the device)

The MENDER_ARTIFACT_PROVIDES(_GROUP), MENDER_ARTIFACT_DEPENDS(_GROUPS), and MENDER_ARTIFACT_NAME(_DEPENDS) variables have been supplemented with new "..._BOOTSTRAP" variables whose values will be added to the original variables to build the arguments passed to mender-artifact

Changelog: Title
Ticket: None

- [x] Make sure that all commits follow the conventional commit [specification](https://www.github.com/mendersoftware/mendertesting/commitlint/grammar.md) for the Mender project.

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

I have a yocto build in which a portion of the data partition is meant to be updated via an update module, I'd like the bootstrap artifact to show the relevant version information, however without this change the mender rootfs artifact generated by yocto would also carry the information (but would obviously not be updating the data partition)

Since these artifacts are generated via classes (applied to the same recipe) and not recipes I do not think it's possible to provide different values without this modification


meta-mender-core/classes/mender-bootstrap-artifact.bbclass
```diff
@@ -11,7 +26,7 @@ IMAGE_CMD:bootstrap-artifact() {
         exit 0
     fi
 
-    if [ -z "${MENDER_ARTIFACT_NAME}" ]; then
+    if [ -z "${MENDER_ARTIFACT_NAME_BOOTSTRAP}" ]; then
             bberror "Need to define MENDER_ARTIFACT_NAME variable."
```
lluiscampos, line 30:
```suggestion
            bberror "Need to define MENDER_ARTIFACT_NAME of MENDER_ARTIFACT_NAME_BOOTSTRAP variable."
```

meta-mender-core/classes/mender-bootstrap-artifact.bbclass
```diff
@@ -30,33 +45,33 @@ IMAGE_CMD:bootstrap-artifact() {
...
-        if [ -n "${MENDER_ARTIFACT_NAME_DEPENDS}" ]; then
+        if [ -n "${MENDER_ARTIFACT_NAME_DEPENDS_BOOTSTRAP}" ]; then
             cmd=""
-            apply_arguments "--artifact-name-depends" "${MENDER_ARTIFACT_NAME_DEPENDS}"
+            apply_arguments "--artifact-name-depends" "${MENDER_ARTIFACT_NAME_DEPENDS_BOOTSTRAP}"
             extra_args="$extra_args $cmd"
         fi
 
-        if [ -n "${MENDER_ARTIFACT_PROVIDES}" ]; then
+        if [ -n "${MENDER_ARTIFACT_PROVIDES}" -o -n "${MENDER_ARTIFACT_PROVIDES_BOOTSTRAP}" ]; then
```
lluiscampos, line 54: Here, and everywhere below, I don't understand why you decided to not use the `_FINAL` variables directly, like:
```suggestion
        if [ -n "${MENDER_ARTIFACT_PROVIDES_FINAL}" ]; then
```
Both would work, but I think using FINAL would be more robust against future changes in the definitions above or other variable expansion bugs.

lluiscampos (changes requested):
Hi @LorenzoRuffati and thank you for the feature. It looks good and I see that it covers a good use case.

Great commit message, by the way. I think that the changelog deserves a bit more explanation. You can use this part of your commit to be all marked for changelog:

```
Changelog: MENDER_ARTIFACT_PROVIDES(_GROUP), MENDER_ARTIFACT_DEPENDS(_GROUPS),
and MENDER_ARTIFACT_NAME(_DEPENDS) variables have been supplemented with
new "..._BOOTSTRAP" variables whose values will be added to the original
variables to build the arguments passed to mender-artifact for the bootstrap artifact
```

There is a suggestion below and a question wrt variable value checking. After you clarify that I'll run the tests and get this one merged rather soon :+1:

meta-mender-core/classes/mender-bootstrap-artifact.bbclass
LorenzoRuffati: "-n" checks for the length of the argument, due to how final is defined I think "-n" would always return true since the space between PROVIDES and PROVIDES_BOOTSTRAP would always be present even if both variables are undefined

image

There might be a way to do this otherwise in yocto that I'm not aware of

LorenzoRuffati:
Hi @lluiscampos, I have answered the question and I will make the change you suggested in your other comment.

Should I amend my existing commit or make a new one with the fixes and the correct changelog message?

meta-mender-core/classes/mender-bootstrap-artifact.bbclass
lluiscampos: Good point!

I was trying to prevent a hacky user defining `MENDER_ARTIFACT_PROVIDES_FINAL=""` in his/her local.conf and that would defeat the check. But these being internal variables for the class we don't really need to be that paranoid. Thanks for proving me wrong :+1:

LorenzoRuffati:
Ok, I added a more descriptive error regarding MENDER_ARTIFACT_NAME and amended the changelog, let me know if you'd like any other change to be made

lluiscampos (approved):
Looks good!

lluiscampos:

PR #2130, opened 2024-06-12
Kirkstone 3.6.5

kacf:

PR #2133, opened 2024-06-13
Dunfell 3.6.5

lluiscampos:
Porting the recipes from #2130 with a fix for the previous update.

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client_3.5.1.bb
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client_3.5.1.bb
@@ -23,7 +23,7 @@ SRCREV = "e48c2fc825ad0c273d19f3ce00eec867091a0806"
 # Enable this in Betas, and in branches that cannot carry this major version as
 # default.
 # Downprioritize this recipe in version selections.
-# DEFAULT_PREFERENCE = "-1"
+DEFAULT_PREFERENCE = "-1"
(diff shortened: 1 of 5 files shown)


meta-mender-core/recipes-mender/mender-client/mender-client_3.5.1.bb
```diff
@@ -23,7 +23,7 @@ SRCREV = "e48c2fc825ad0c273d19f3ce00eec867091a0806"
 # Enable this in Betas, and in branches that cannot carry this major version as
 # default.
 # Downprioritize this recipe in version selections.
-# DEFAULT_PREFERENCE = "-1"
+DEFAULT_PREFERENCE = "-1"
```
kacf, line 26: Is this in the previous tag?

If so, I think I would leave it actually. They have already been out for a long time, and downgrading because the 3.x series disappears might actually be worse in this case.

Same for the rest.

kacf (approved):

meta-mender-core/recipes-mender/mender-client/mender-client_3.5.1.bb
lluiscampos: > Is this in the previous tag?

No, it is not tagged. I introduced them in my previous PR #2125 (not yet tagged) and that is why I considered correct to amend.

PR #2134, opened 2024-06-14
chore: Fix incorrect changelog entry.

kacf:
Remove last sentence, the fix does not affect UBI directly, but makes it impossible to omit mender-flash on UBI systems.

Changelog(fix): Fix autogeneration of `mender-binary-delta.conf` so that the user value for `MENDER_FLASH` is respected, instead of the previously wrongly hard-coded value of `mender-flash`. Without this fix it is impossible to omit `mender-flash` from UBI systems, even when `ubiupdatevol` would have been sufficient.
Cancel-changelog: 196753680be9e69790cf0f17d57b661e9d0ed59b


kacf:
@lluiscampos: FYI

kacf:
Merging without review/pipeline. This is only a small doc update.

kacf:
Note that the fix is not required on dunfell, because it doesn't have a changelog entry there (no tag contains the bug).

PR #2099, opened 2024-04-22
MEN-7112: Add support for Yocto 5.0 "scarthgap"

TheYoctoJester:

PR #2140, opened 2024-06-28
(verify) MEN-7299 and MEN-7300

lluiscampos:
Launching this PR to verify the fix independently. To be merged as part of #2099.


lluiscampos:
@TheYoctoJester please review the last commit (only)

TheYoctoJester (commented):
Very good catch! Now unfortunately that rings a bell for me :-( See 4d7e55712395bf417d7c94cf471132742eb82c6d, e.g. we used to have this patched into the `mkfs.ext` defaults, and found it to be problematic. Your approach is way more elegant, but still raises the question: will it cause the same mounting problems again?

PR #2147, opened 2024-07-04
feat: Add a pkcs11-provider recipe

danielskinstad:
Commits are cherry-picked from https://github.com/mendersoftware/meta-mender/pull/2002

### DIFF
--- a/meta-mender-demo/recipes-connectivity/pkcs11-provider/pkcs11-provider.bb
+++ b/meta-mender-demo/recipes-connectivity/pkcs11-provider/pkcs11-provider.bb
@@ -1,0 +1,16 @@
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://COPYING;md5=b53b787444a60266932bd270d1cf2d45 \
+                    file://LICENSES/Apache-2.0.txt;md5=3b83ef96387f14655fc854ddc3c6bd57"
+
+SRC_URI = "git://github.com/latchset/pkcs11-provider;protocol=https;branch=main"
+
+PV = "0.2+git${SRCPV}"
+SRCREV = "d994cc26790d31242fcee5df7d07d59f1698f073"
+
+S = "${WORKDIR}/git"
+
+DEPENDS:append = " openssl p11-kit autoconf-archive"
+
+inherit pkgconfig autotools
+
+FILES:${PN} =+ "${libdir}/ossl-modules/pkcs11.so"
--- a/meta-mender-qemu/recipes-mender/mender-client/mender-client-test-files.inc
+++ b/meta-mender-qemu/recipes-mender/mender-client/mender-client-test-files.inc
@@ -7,8 +7,9 @@
 #           for the OpenSSL to load the lib and access the key from the HSM
 # * opensc: pkcs11-tool for initialization of token
 # * gnutls-bin: provides p11tool for getting the contents of the token
+# * pkcs11-provider is a third-party implementation of a PKCS#11 provider for OpenSSLv3
 DEPENDS:append:mender-image:qemuall = " softhsm opensc gnutls p11-kit libp11"
-RDEPENDS:mender-auth:append:mender-image:qemuall = " softhsm opensc gnutls p11-kit libp11 gnutls-bin"
+RDEPENDS:mender-auth:append:mender-image:qemuall = " softhsm opensc gnutls p11-kit libp11 gnutls-bin pkcs11-provider"
 
 # Needed for the TestFaultTolerance tests in the integration repository.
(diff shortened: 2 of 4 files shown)


danielskinstad:

TheYoctoJester (commented):
Is there a specific reason why this is tied to/hidden in the `mender-image` feature? That feels kinda off to me. From the users perspective, if I want to use this functionality, what do I need to add? How can I find out by looking at the `mender-auth` providing recipe or documentation?

In a perfect world, this would be a build time `PACKAGECONFIG` flag to `mender`, and adjusting `RDEPENDS` accordingly. I'm fine with non-perfect solutions too, but this one feels extremely sneaky.

meta-mender-qemu/recipes-mender/mender-client/mender-client-test-files.inc
```diff
@@ -7,8 +7,9 @@
 #           for the OpenSSL to load the lib and access the key from the HSM
 # * opensc: pkcs11-tool for initialization of token
 # * gnutls-bin: provides p11tool for getting the contents of the token
+# * pkcs11-provider is a third-party implementation of a PKCS#11 provider for OpenSSLv3
 DEPENDS:append:mender-image:qemuall = " softhsm opensc gnutls p11-kit libp11"
-RDEPENDS:mender-auth:append:mender-image:qemuall = " softhsm opensc gnutls p11-kit libp11 gnutls-bin"
+RDEPENDS:${PN}:append:mender-image:qemuall = " softhsm opensc gnutls p11-kit libp11 gnutls-bin pkcs11-provider"
```
lluiscampos, line 12: Careful: this should be `mender-auth`. The original patch has `${PN}` because by then all the softhsm was linked to `mender-client` and not `mender-auth`. But note that the rest of this file is now using `mender-auth`


```suggestion
RDEPENDS:mender-auth:append:mender-image:qemuall = " softhsm opensc gnutls p11-kit libp11 gnutls-bin pkcs11-provider"
```

meta-mender-demo/recipes-connectivity/pkcs11-provider/pkcs11-provider.bb
```diff
@@ -0,0 +1,17 @@
+
```
lluiscampos, line 1: Superfluous empty line, consider removing

meta-mender-qemu/recipes-mender/mender-client/mender_%.bbappend
```diff
@@ -1,4 +1,18 @@
...
+# * libp11: OpenSSL ENGINE plugin PKCS#11 implementation, depends on p11-kit
+#           for the OpenSSL to load the lib and access the key from the HSM
+# * opensc: pkcs11-tool for initialization of token
+# * gnutls-bin: provides p11tool for getting the contents of the token
+# * pkcs11-provider is a third-party implementation of a PKCS#11 provider for OpenSSLv3
+DEPENDS:append:mender-image:qemuall = " softhsm opensc gnutls p11-kit libp11"
+RDEPENDS:mender-auth:append:mender-image:qemuall = " softhsm opensc gnutls p11-kit libp11 gnutls-bin pkcs11-provider"
 
-# Remove mender-snapshot to save storage space. We are not running snapshot tests for this device
-RRECOMMENDS:mender-update:remove:vexpress-qemu-flash = "mender-snapshot"
```
lluiscampos, line 4: Removed by mistake when resolving the conflict? Please restore this part.

meta-mender-qemu/recipes-mender/mender-client/mender_%.bbappend
```diff
@@ -1,4 +1,18 @@
-require mender-client-test-files.inc
```
lluiscampos, line 1: I think this is wrong (I am unsure if this _was_ correct in the original PR).

Either we keep the `require` and the changes in `mender-client-test-files.inc`, or we remove completely the inc file and keep the changes here in the `mender_%.bbappend`. --> IOW check that the inc file is only included from this bbapend file.

lluiscampos (changes requested):
Few things that we can clean up. See below.

> Is there a specific reason why this is tied to/hidden in the `mender-image` feature? That feels kinda off to me.

@TheYoctoJester There is no specific reason, we could completely remove the feature scope. But note that this is already the case for the existing `softhsm` and related dependencies so in any case this PR is not making it _more_ messy.

> From the users perspective, if I want to use this functionality, what do I need to add? 

We don't expect users to use QEMU images, do we? It is only for developers and CI to test features.

lluiscampos (approved):

danielskinstad:

lluiscampos:
The failing test is unrelated. See https://github.com/mendersoftware/integration/pull/2631

PR #2151, opened 2024-08-01
chore: aligned expected Mender Artifact output with updated Mender Artifact

mzedel:


mzedel:
@mender-test-bot start pipeline --pr mender-artifact/624

tests/acceptance/test_build.py
```diff
@@ -513,7 +513,7 @@ def test_multiple_device_types_compatible(
         )
 
         output = run_verbose("mender-artifact read %s" % image, capture=True)
-        assert b"Compatible devices: '[machine1 machine2]'" in output
+        assert b"Compatible devices: [machine1, machine2]" in output
```
lluiscampos, line 516: We cannot modify the expected format here unconditionally because `meta-mender` can build/test any supported version of `mender-artifact`.

The best solution would be to use `if version_is_minimum(bitbake_variables, "mender-artifact", "3.12.0")` to guard the check (see examples in this same file).

Alternatively you could use a regex, but that will make the check weaker.

lluiscampos (changes requested):

lluiscampos (approved):

mzedel:
> @mender-test-bot start pipeline --pr mender-artifact/624

lluiscampos:
* scarthgap
* kirkstone

PR #2161, opened 2024-08-12
chore: remove warnings for mender-connect and mender-snapshot recipes

danielskinstad:


danielskinstad:

meta-mender-core/recipes-mender/mender-connect/mender-connect.inc
```diff
@@ -20,7 +20,13 @@ inherit systemd
 
 GO_IMPORT = "github.com/mendersoftware/mender-connect"
 
+# Skip text relocation warnings
+INSANE_SKIP:${PN} = "textrel"
```
lluiscampos, line 24: Have you looked a bit into the root of this issue? By looking at the [go.bbclass recipe](https://git.yoctoproject.org/poky/tree/meta/classes-recipe/go.bbclass?h=scarthgap) you can see that they pass the `-buildmode=pie` to exactly address this.

My understanding is that this is working fine on the pure go recipes, like `mender-snapshot` below, but not on go code that links with C code, like `mender-connect` and (legacy) `mender-client`.

Although you could say that we are already skipping this check in other recipes, I think it is worth spending a bit more time trying to fix the issue(s) rather than skipping the check.

danielskinstad: Yes, pie fixes the text relocations, but I didn't get why that wasn't done for the go client, so I just copied the skip.

lluiscampos: Yes, we already have the skip in other recipes so this is a suggestion, not a requirement :)

meta-mender-core/recipes-mender/mender-connect/mender-connect.inc
```diff
@@ -20,7 +20,13 @@ inherit systemd
...
+
 do_compile() {
+    # Remove file system paths from binary
+    export GOFLAGS="-trimpath"
```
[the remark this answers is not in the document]
danielskinstad: Right, so I'll ignore the warnings here, and create two PRs and add trimpath to the makefiles

meta-mender-core/recipes-mender/mender-connect/mender-connect.inc
```diff
@@ -21,6 +21,9 @@ inherit systemd
 GO_IMPORT = "github.com/mendersoftware/mender-connect"
 
 do_compile() {
+    # Remove file system paths from binary and use buildmode PIE to avoid text relocations
+    export GOFLAGS="-trimpath -buildmode=pie"
```
lluiscampos, line 25: `-trimpath` we should not apply it here. Let's rely on the Makefile (and it will be fixed only for the next releases)

lluiscampos: And with regards to `-buildmode=pie` I believe it is not working. This will only apply to the Go code and not the C code. For C code you need the equivalent gcc flag in `CGO_LDFLAGS`. Additionally, we should not do this hack-ish export but just amend to the yocto variable (out of the `do_compile`).

lluiscampos (changes requested):
It sill needs some work.

meta-mender-core/recipes-mender/mender-connect/mender-connect.inc
```diff
@@ -53,21 +48,14 @@ do_prepare_mender_connect_conf[vardeps] = " \
...
-    # install configuration
-    mkdir -p  ${D}/${sysconfdir}/mender
+do_install:append() {
+    # Install configuration
+    mkdir -p ${D}/${sysconfdir}/mender
     install -m 0600 ${B}/mender-connect.conf ${D}/${sysconfdir}/mender/mender-connect.conf
+
+    # Install systemd service
+    install -m 755 -d ${D}${systemd_unitdir}/system
+    install -m 0644 ${B}/src/${GO_IMPORT}/support/mender-connect.service ${D}${systemd_unitdir}/system/
```
lluiscampos, line 58: Pre-existing,

I noticed that the install of the systemd service for this recipe is not guarded with the `mender-systemd` feature. This was wrong and needs fixing (separate commit with changelog). It can look like:

```
do_install:append() {
    # Install configuration
    mkdir -p ${D}/${sysconfdir}/mender
    install -m 0600 ${B}/mender-connect.conf ${D}/${sysconfdir}/mender/mender-connect.conf
}

do_install:append:mender-systemd() {
    # Install systemd service
    install -m 755 -d ${D}${systemd_unitdir}/system
    install -m 0644 ${B}/src/${GO_IMPORT}/support/mender-connect.service ${D}${systemd_unitdir}/system/
}
```

lluiscampos (commented):
This is a great alternative. Something that I personally favor (to use upstream classes) instead of custom make targets wrapping go commands.

There is one major problem with this: the injection of the version. If you read the current Makefile for `mender-connect` you will find the logic to inject the version with `-ldflags`. We would then need to move that to the yocto recipe, but it is so much of an internal detail that feels wrong to expose.

I appreciate the push in the last minute for the sprint, but we need to discuss this a bit more :weary:

lluiscampos:
> There is one major problem with this: the injection of the version. If you read the current Makefile for `mender-connect` you will find the logic to inject the version with `-ldflags`. We would then need to move that to the yocto recipe, but it is so much of an internal detail that feels wrong to expose.

According to the `go` class, we can use `GO_EXTRA_LDFLAGS` for this. Maybe with a good comment for it is not such an ugly thing to do after all :angel:

danielskinstad:
Forgot to add version for snapshot, but something like this?

danielskinstad:
> > Forgot to add version for snapshot, but something like this?
> 
> Does it work? If it works this is exactly what I thought about.

```
root@qemux86-64:~# mender-connect --version
mender-connect version 2.2.1    runtime: go1.22.4
root@qemux86-64:~# mender-snapshot --version
1.0.0   runtime: go1.22.4
```

Yes, it should work

meta-mender-core/recipes-mender/mender-connect/mender-connect.inc
```diff
@@ -53,23 +47,22 @@ do_prepare_mender_connect_conf[vardeps] = " \
...
-        sysconfdir=${sysconfdir} \
-        systemd_unitdir=${systemd_unitdir} \
-        install-bin \
-        install-systemd
-
-    # install configuration
-    mkdir -p  ${D}/${sysconfdir}/mender
+# Set the version of mender-connect by passing it as an extra ldflag to go.bbclass
+# This is equivalent to how it's set in the Makefile
+GO_EXTRA_LDFLAGS:append = "-X github.com/mendersoftware/mender-connect/config.Version=$(git describe --tags --dirty --exact-match 2>/dev/null || git rev-parse --short HEAD)"
```
lluiscampos, line 52: (thinking out loud)

To be yocto-correct, we should use [PV](https://docs.yoctoproject.org/ref-manual/variables.html#term-PV) variable:

```suggestion
# This is equivalent to how it's set in the Makefile
GO_EXTRA_LDFLAGS:append = "-X github.com/mendersoftware/mender-connect/config.Version=${PV}"
```

However, this in effect will modify how the versions are seen for the `git` recipes, as there we would get something like version `master-git` instead of the Git sha. Then, to workaround this, we could modify *only for the git recipe* the GO_EXTRA_LDFLAGS to use the git command.

At the end of the day how it is right now represents better how it was behaving under the hood, so I think it is just fine :)

lluiscampos (approved):
Fantastic! See my comment below but I'd say let's merge as-is.

Thank you for your patience and tenacity searching for the best way to handle this. Sometimes a 1SP can hide many surprises :sweat_smile:

danielskinstad:

danielskinstad:

PR #2190, opened 2024-10-28
QA-784: Rework sstate cache failures workaround

lluiscampos:
The previous workaround resulted too aggressive - not only making the CI slower but also breaking other parallel tests by removing their checkout dirs, for example. See commit c66217fe.

Try this other workaround instead: disable the sstate cache before the rebuilds of this test.


kacf (approved):

lluiscampos:
@mender-test-bot start pipeline again

lluiscampos:

lluiscampos:
This workaround is not working neither. I have another idea but I'll submit separately so that I can refer to this PR from there :+1:

PR #2198, opened 2024-11-01
Fix building with Mender disabled

eigendude:

- [x] Make sure that all commits follow the conventional commit [specification](https://www.github.com/mendersoftware/mendertesting/commitlint/grammar.md) for the Mender project.

### Description

When the meta-mender stack is included but the distro is building without Mender (such as for testing), an error is raised. This is due to a variable being used, but never defined because some of the Mender conf isn't visited.

By setting the base default to "" it lets later code use the variable without all conf being processed.

## Motivation and context.

While testing, I had the Mender layer included, but I was building without Mender to hunt down a U-boot issue with the Raspberry Pi 5. This change at least let me get stuck in U-boot.

## How has this been tested?

In my local.conf, I have:

```
################################################################################
# Distribution features
################################################################################

# Set to "0" to disable Mender
ENABLE_MENDER = "1"

# Set to "0" to disable graphical features (OpenGL, windowing, Gnome, etc)
ENABLE_GRAPHICS = "1"
```

Before, when I set `ENABLE_MENDER` to 0, I got the error:

```
  ERROR: ExpansionError during parsing grub-efi-mender-precompiled_2.04.bb
  Traceback (most recent call last):
    File "<expansion>", line 1, in <module>
  bb.data_smart.ExpansionError: Failure expanding expression ${@'uboot_version_logic.inc' if d.getVar('MENDER_EFI_LOADER').startswith('u-boot') else ''} which triggered exception AttributeError: 'NoneType' object has no attribute 'startswith'
```

After, the build succeeds with `ENABLE_MENDER="0"`, and I can boot into Linux on the RPi 4 with Mender disabled.

### DIFF
--- a/meta-mender-core/recipes-bsp/grub/grub-efi-mender-precompiled_2.04.bb
+++ b/meta-mender-core/recipes-bsp/grub/grub-efi-mender-precompiled_2.04.bb
@@ -28,6 +28,8 @@ SRC_URI = " \
 
 S = "${WORKDIR}/git"
 
+MENDER_EFI_LOADER ?= ""
+
 require ${@'uboot_version_logic.inc' if d.getVar('MENDER_EFI_LOADER').startswith('u-boot') else ''}
 
 DEPENDS:append = " ${MENDER_EFI_LOADER}"


lluiscampos:
@eigendude I cannot find where `ENABLE_MENDER` is defined nor used in the layer. Can you elaborate a bit more on your setup?

The change itself looks good (thanks!) but I'd like to understand in which scenarios this is a problem so that we can test it on our side.

eigendude:
Sorry, I should have offered more context. In our distro's meta-layer we have:

```
# Conditionally include Mender support based on ENABLE_MENDER
require ${@'conf/distro/include/curie-mender.inc' if d.getVar('ENABLE_MENDER') == '1' else ''}
```

ENABLE_MENDER is a flag I added to local.conf to control Mender inclusion in the meta layer.

Then in curie-mender.inc we have:

```
# Inherit the Mender class
INHERIT += "mender-full"
```

So this PR fixes building with the Mender layer present but without the `INHERIT += "mender-full"` line being processed.

lluiscampos (changes requested):
I finally managed to reproduce locally and verify your fix :+1: Sorry that it took some time.

The code looks good!

Two things:
* The commit body needs a `Ticket: None` entry (being a fix a `Ticket` tag is needd)
* I suggest we rephrase a bit the changelog entry. This is a parse error and I would phrase it like "Fix parsing error for `grub-efi` recipe when building with Mender disabled"

lluiscampos (approved):

lluiscampos:
@eigendude Yes, very good! Thanks

lluiscampos:
The failing test is due to this PR being out of date from `master` - namely missing #2203 

Otherwise is :green_circle: so merging as-is.

PR #2199, opened 2024-11-02
QA-784: Workaround SPDX/sstate issue by explicit running the task

lluiscampos:

PR #2181, opened 2024-10-11
chore: Add recipe for `mender-flash` 1.0.1

lluiscampos:

PR #2200, opened 2024-11-04
QA-784: Swap order of "pn-" prefix to workaround SPDX/sstate issue

lluiscampos:
Setting/unsetting PREFERRED_VERSION and rebuilding the recipes invalidates SPDX data for them.

The recipes that use externalsrc will correctly recreate the data the next time a rootfs is being built, but for some inexplicable reason this does not happen for recipes that using sources from upstream, an instead the build will rely on the stale `do_create_runtime_spdx_setscene` from before the tests.

At the time of writing this, this is the case for `mender-flash` so that consecutive tests would fail with:
```
ERROR: No SPDX file found for package mender-flash, False sstate:mender-flash:core2...
```

But the really interesting bit is that this happens only if we build first with `PREFERRED_VERSION:recipe` and then
`PREFERRED_VERSION:pn-recipe`, but not if done in the reverse order (??) nor if build only for one of the two variants (??).

Note that we have already tried cleaning everything (too aggressive), not using the cache (it didn't work) and manually recreating the SPDX data (it didn't work):
* https://github.com/mendersoftware/meta-mender/pull/2181/commits/c66217fe4
* https://github.com/mendersoftware/meta-mender/pull/2190
* https://github.com/mendersoftware/meta-mender/pull/2199

### DIFF
--- a/tests/acceptance/test_build.py
+++ b/tests/acceptance/test_build.py
@@ -461,7 +461,6 @@ b524b8b3f13902ef8014c0af7aa408bc  ./usr/local/share/ca-certificates/mender/serve
         + [("mender-flash", version) for version in versions_of_recipe("mender-flash")]
         + [("mender-flash", None)],
     )
-    @pytest.mark.skip(reason="QA-784")
     def test_preferred_versions(self, prepared_test_build, recipe, version):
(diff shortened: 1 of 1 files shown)


lluiscampos:
@mender-test-bot start pipeline - patience is the new black

PR #2202, opened 2024-11-06
QA-784: Use a dedicated workspace for `test_preferred_versions`

lluiscampos:
Besides moving the test to a separate class, this commit modifies the
following:
* Use a class scoped test build dir, that unlike the default one is not
shared across all tests in the same runner.
* Disable write to shared state cache to not modify anything (note that
the test can still read the cache via the `SSTATE_MIRRORS` setup).
* Replace the aggressive `cleanall` with `clean` - we still want to
clean the recipe for further rebuilds withing the same test build dir
* And enable the test \o/

The reordering of the code affects the ordering of the tests execution
in CI, but before this commit I made sure that even with the previous
ordering the fix is "good enough" to get a green pipeline.

Finally, if you still have energy to keep reading, see all the previous
attempts that we have done to workaround this annoying issue:
* https://github.com/mendersoftware/meta-mender/pull/2181/commits/c66217fe4
* https://github.com/mendersoftware/meta-mender/pull/2190
* https://github.com/mendersoftware/meta-mender/pull/2199
* https://github.com/mendersoftware/meta-mender/pull/2200

### DIFF
--- a/tests/acceptance/test_build.py
+++ b/tests/acceptance/test_build.py
@@ -60,23 +60,6 @@ def extract_partition(img, number, dstdir):
     )
 
 
-def versions_of_recipe(recipe_name, recipe_dir=None):
-    """Returns a list of all the versions we have of the given recipe, excluding
-    git recipes."""
-
-    if recipe_dir == None:
-        recipe_dir = recipe_name
-
-    versions = []
-    for entry in os.listdir("../../meta-mender-core/recipes-mender/%s/" % recipe_dir):
-        match = re.match(
-            r"^%s_([1-9][0-9]*\.[0-9]+\.[0-9]+[^.]*)\.bb" % recipe_name, entry
-        )
-        if match is not None:
-            versions.append(match.group(1))
-    return versions
-
-
 class TestBuild:
     @pytest.mark.cross_platform
     @pytest.mark.min_mender_version("1.0.0")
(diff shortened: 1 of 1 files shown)


lluiscampos:
@mender-test-bot start pipeline :ice_cream:

lluiscampos:
@mender-test-bot start pipeline :fish_cake:

lluiscampos:
Ping @danielskinstad too - sorry I missed you in this super important PR :grin:

tests/acceptance/test_build.py
```diff
@@ -496,6 +501,12 @@ def test_preferred_versions(self, prepared_test_build, recipe, version):
                         is not None
                     ):
                         continue
+
+                    elif line.startswith("SSTATE_DIR"):
+                        new_fd.write(
+                            "# QA-784: Cache write disabled for this test\n# %s" % line
+                        )
```
kacf, line 508: Won't this, combined with the changes from mender-image-tests, build everything from scratch? If the sstate cache is not available, and it's a fresh build directory as well, it won't have anywhere to get the previously built files from.

tests/acceptance/test_build.py
```diff
@@ -510,11 +521,11 @@ def test_preferred_versions(self, prepared_test_build, recipe, version):
...
 
-        # QA-784: setting/unsetting PREFERRED_VERSION and rebuilding the recipes breaks some times
-        # the state cache so that consecutive tests will fail in a very inexplicable way.
-        # Cleaning the cache here for each version built seems to workaround the issue...
-        run_verbose("%s && bitbake --cmd cleanall %s" % (init_env_cmd, recipe))
+        # Clean leftovers to have a fresh start for the next recipe
+        run_verbose("%s && bitbake --cmd clean %s" % (init_env_cmd, recipe))
+
 
+class TestBuildMoreTODOMove:
```
kacf, line 528: Forgotten?

tests/acceptance/test_build.py
lluiscampos: No, because the test adds [the mirrors here](https://github.com/mendersoftware/mender-image-tests/blob/698cb01171f7ef1faaabbe8311f21a06cf1a0432/tests/utils/fixtures/fixtures.py#L470). I wrote "cache write" because it actually reads from the cache.

(Also, this alone is not enough, it still runs intro trouble building for example `mender` recipes that depens on other things like `mender-snapshot` (see [CI log here](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/8286531361)), and that is why I also kept the soft clean below)

tests/acceptance/test_build.py
lluiscampos: I'll rework after your feedback but I wanted to make absolutely sure that the tests were being ordered the same way (hence I split this class into two) to validate the approach. Ref my general comment above.

tests/acceptance/test_build.py
kacf: Ok, thanks for explaining! Then it looks good, I think that deserves a comment! 🙂

kacf (approved):
Nice fix! I like this much better than the full sstate clean.

lluiscampos:

lluiscampos:

PR #2213, opened 2024-11-28
Cleanups for mender recipe around `PACKAGECONFIG`

lluiscampos:

PR #2215, opened 2024-12-02
master: Add recipes from Mender 3.7 LTS release

lluiscampos:
Changelog: Add recipe for `mender` 4.0.5
Changelog: Add recipe for `mender-artifact` 3.11.3
Changelog: Add recipe for `mender-gateway` 1.2.1

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender_4.0.5.bb
+++ b/meta-mender-core/recipes-mender/mender-client/mender_4.0.5.bb
@@ -1,0 +1,38 @@
+require mender-client-cpp.inc
+require mender_4.x.inc
+
+################################################################################
+#-------------------------------------------------------------------------------
+# THINGS TO CONSIDER FOR EACH RELEASE:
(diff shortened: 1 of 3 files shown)


meta-mender-core/recipes-mender/mender-client/mender_4.0.5.bb
```diff
@@ -0,0 +1,37 @@
+require mender-client-cpp.inc
```
danielskinstad, line 1: The other recipes also require `mender_4.x.inc`, is that not needed here?

danielskinstad (approved):
Looks good to me, just a question in the mender-client recipe

meta-mender-core/recipes-mender/mender-client/mender_4.0.5.bb
lluiscampos: Good catch! Yes, that is indeed needed here. I made these recipes before #2213 and later I didn't update.

Fixed now, thanks!

PR #2218, opened 2024-12-04
scarthgap: updates for release v2024.12

lluiscampos:
Bringing from `master` everything that has not been yet ported since last release.

PR #2178, opened 2024-10-10
feat: add mender-prepopulate-inactive-partition feature

TheYoctoJester:
By default, the inactive partition of a generated image is empty to allow for more efficient compression.

As squashfs (a supported ARTIFACTIMG_FSTYPE) does not support empty filesystems, it already triggers implicit logic in the mender-part-images.bbclass to populate the inactive partition with the full filesystem.

This commit adds an explicit trigger through the MENDER_FEATURES mechanism using the "mender-prepopulate-inactive-partition" key.

The primary use case is to check the bootloader integrations (A/B switching, A/B update, A/B rollback) directly on the generated images without additional modifications or processing.

# External Contributor Checklist

🚨 Please review the [guidelines for contributing](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md) to this repository.

- [ ] Make sure that all commits follow the conventional commit [specification](https://www.github.com/mendersoftware/mendertesting/commitlint/grammar.md) for the Mender project.

The majority of our contributions are fixes, which means your commit should have
the form below:

```
fix: <SHORT DESCRIPTION OF FIX>

<OPTIONAL LONGER DESCRIPTION>

Changelog: <USER-FRIENDLY-CHANGE-DESCRIPTION> or <None>
Ticket: <TICKET NUMBER> or <None>
```

- [ ] Make sure that all commits are signed with [`git --signoff`](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work). Also note that the signoff author must match the author of the commit.

### Description

Please describe your pull request.

Thank you!

PR #2220, opened 2024-12-06
kirkstone: updates for release v2024.12

lluiscampos:
Bringing from `scarthgap` #2218 and few other things that were pending.


lluiscampos:
FYI @TheYoctoJester, I didn't bring #2178 for considering it of low value for `kirkstone`. Do you think we should port it?

meta-mender-core/recipes-mender/mender-client/mender_4.x.inc
```diff
@@ -0,0 +1,5 @@
+
+# The docker module depends on bash, and of course on docker. However, docker is
+# a very large requirement, which we will not mandate. Bash however, we require,
+# because otherwise the Yocto QA checks will complain.
+RDEPENDS:mender-update:append = "${@bb.utils.contains('PACKAGECONFIG', 'modules', ' bash', '', d)}"
```
lluiscampos, line 5: Note to self,

On `kirkstone`, this needs to be added for 3.x recipes too.

meta-mender-core/recipes-mender/mender-connect/mender-connect.inc
```diff
@@ -13,19 +13,13 @@ SYSTEMD_SERVICE:${PN} = "mender-connect.service"
 
 B = "${WORKDIR}/build"
 
-inherit go
+inherit go-mod
```
lluiscampos, line 16: Note to self,

This change is the culprit of the CI failures. Changing the compile steps has added two dependencies on `mender-connect`:
* go-runtime >= 1.17.13
* libstd.so()(64bit)

Which increased the size of the final image making it not fit for one of the configurations.

To investigate:
* Why?
* Does it affect the same way `scarthgap` builds?
* Should we revert or just increase the image size?

lluiscampos: See:
* https://github.com/mendersoftware/meta-mender/pull/2221

meta-mender-core/recipes-mender/mender-client/mender_4.x.inc
lluiscampos: False alarm, 3.x recipes use `mender-client-go.inc` which was not modified (iow has this depenendy).

danielskinstad (approved):

lluiscampos:
For reference, the build with Mender 3.7.8 and updating `kirkstone-next` was @ [Pipeline 1578583510](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/1578583510)

PR #2221, opened 2024-12-07
MEN-7275 (follow-up): Prefer static linkage for `mender-connect` and `mender-snapshot`

lluiscampos:
This is a follow-up from 0b85500.

By switching to `go.bbclass` definitions for compile and install tasks, we have also switched from static linking (golang's default) to dynamic linking (by passing `-linkshared` to the go commands - see `go.bbclass` source code).

While using shared linkage on golang binaries have benefits (and presumably is preferred in Yocto), we didn't intend to change the linkage of Mender binaries in the commit mentioned above.

(Note that `mender-connect` is actually dynamically linked for other dependencies, but at least the runtime and stdlib shall be statically linked).

### DIFF
--- a/meta-mender-core/recipes-mender/mender-connect/mender-connect.inc
+++ b/meta-mender-core/recipes-mender/mender-connect/mender-connect.inc
@@ -20,6 +20,9 @@ inherit systemd
 
 GO_IMPORT = "github.com/mendersoftware/mender-connect"
 
+# Prefer static linkage (golang's default)
+GO_LINKSHARED = ""
+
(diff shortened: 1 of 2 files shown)


lluiscampos:

danielskinstad (approved):
Nice find! I don't have a cached Yocto build atm, so I can't test it. But I read the go.bbclass, and saw someone else recommend the same approach, and it looks good.

lluiscampos:
The CI failure is unrelated. Merging this PR.

lluiscampos:
* scarthgap

PR #2226, opened 2024-12-18
Add recipes from Mender Client and Mender Gateway releases

lluiscampos:
Changelog: Add recipe for `mender` 5.0.0
Changelog: Add recipe for `mender-gateway` 2.0.0
Changelog: Add recipe for `mender-connect` 2.3.0
Changelog: Add recipe for `mender-artifact` 4.0.0
Changelog: Add recipe for `mender-monitor` 1.4.0

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender_5.0.0.bb
+++ b/meta-mender-core/recipes-mender/mender-client/mender_5.0.0.bb
@@ -1,0 +1,40 @@
+require mender-client-cpp.inc
+
+################################################################################
+#-------------------------------------------------------------------------------
+# THINGS TO CONSIDER FOR EACH RELEASE:
+# - SRC_URI (particularly "branch")
+# - SRCREV
+# - DEFAULT_PREFERENCE
+#-------------------------------------------------------------------------------
+
+SRC_URI = "gitsm://github.com/mendersoftware/mender;protocol=https;branch=5.0.x"
+
+# Tag: 5.0.0
+SRCREV = "47313c6d30db1da2d24ee099b72e0917d2b82254"
+
+# Enable this in Betas, and in branches that cannot carry this major version as
+# default.
+# Downprioritize this recipe in version selections.
+#DEFAULT_PREFERENCE = "-1"
+
+################################################################################
+
+# DO NOT change the checksum here without make sure that ALL licenses (including
+# dependencies) are included in the LICENSE variable below. Note that for
+# releases, we must check the LIC_FILES_CHKSUM.sha256 file, not the LICENSE
+# file.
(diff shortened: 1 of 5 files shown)


meta-mender-core/recipes-mender/mender-client/mender_5.0.0.bb
```diff
@@ -0,0 +1,41 @@
+require mender-client-cpp.inc
+require mender_4.x.inc
```
danielskinstad, line 2: Should we have a `mender_5.x.inc` and include that instead? Might be a bit overkill since it would be similar to the current one, so feel free to ignore this, I don't have a strong opinion

danielskinstad (approved):
Looks good to me, just a comment/question in the mender-client recipe

meta-mender-core/recipes-mender/mender-client/mender_5.0.0.bb
lluiscampos: Good observation, thanks.

But no, this `require` is only needed for 4.x recipes because it handles dependencies for the community update modules that we have removed from 5.x.

lluiscampos: Oh, wait! You actually found a problem! -> the `require` needs to be removed :see_no_evil:

lluiscampos: Fixed now. Big thanks

TheMeaningfulEngineer (approved):
The hashes correspond to the latest commits for mender, connect and artifact.

Feels like some versioning duplication.
First the filename version (i.e. 5.0.0), than the branch (5.0.x), then the hash (5.0.0 git tag).
But yeah, don't have a snippet for a suggestion solution and it's easy to theorize solutions without skin in the game. :)

LGTM

lluiscampos:
> Feels like some versioning duplication. First the filename version (i.e. 5.0.0), than the branch (5.0.x), then the hash (5.0.0 git tag). 

Here you are just fighting the system,
* filename -> that is how you define PV inyocto
* the branch -> yocto is strict about this, you must specify the branch for the git fetcher
* the tag is just a comment for a human to read. Tags in Git are mutable, so that is why yocto requires git sha instead.

lluiscampos:
Not running the integration pipeline after the [last push](https://github.com/mendersoftware/meta-mender/compare/c2d6124b01e476f596d8e5e1f98caa927347a6d0..f6cf044e925cad5bbc603eb92e40cc931dcb62e8). The change is considered low risk and we are about to test all these recipes again on the stable branches `scarthgap` and `kirkstone` :+1:

PR #2246, opened 2025-01-22
chore: Fix syntax for test code figuring out the mender versions

lluiscampos:
These went unnoticed when we updated the syntax from kirkstone to scarthgap. All of them however relate to internal testing, so no real damage has been done elsewhere.


lluiscampos:

PR #2268, opened 2025-03-11
:test_tube: MEN-8159: Releases Mender Client LTS 5.0 and 4.0

lluiscampos:
Work in progress


merlin-northern (approved):
okey all the client side looks good to me -- I checked all the md5sums from the client repo in both branches..

PR #2269, opened 2025-03-13
feat: Make sure /data partition is last

katomaso:
Make sure /data partition is last when generating .WKS file dynamically.

Now having mender-growfs-data makes more sense because there will be space for expanding the partition

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

Thank you!


lluiscampos:
Hi @katomaso and thank you for your contribution.

I need some time to look at your work - this week I am traveling and working with reduced capacity. From the top of my head, I remember that we have somewhere other use case (and an explicit test) to have the extra partitions _after_ the data one, so I need to refresh my mind and then propose a way to make both use cases possible.

Thanks for your patience :sweat_smile:

TheYoctoJester:
Hi @katomaso, while I agree that your PR provides value for the use case of extra partitions AND growing the data partition, unfortunately it conflicts with the documented behavior at https://docs.mender.io/operating-system-updates-yocto-project/variables#mender_extra_parts. Hence, it can't be easily changed as there probably are users depending on it.

Possible ways forward might be:
- printing a warning or even error if both the partition growing feature and extra partitions are configured, as the behavior will be conflicting. I think this is the most obvious thing to do.
- changing the growing script to optionally pick the last partition
- adding your proposed strategy as a secondary route
... and there possibly are more, or combinations thereof.

Thoughts? Also @lluiscampos

katomaso:
Hi @TheYoctoJester 

My thoughts on possible ways forward inlined:

- _printing a warning or even error if both the partition growing feature and extra partitions are configured, as the behavior will be conflicting. I think this is the most obvious thing to do._
  This is current behaviour of many projects including mender-kernel that we use. Simply warning the user does not solve our issue.

- _changing the growing script to optionally pick the last partition_
  I would not go this way since there is no guarantee on the order of extra partitions. I tried to add ordering to the code but using the order from `MENDER_EXTRA_PARTS` (not its flags) but the recipes appends can be arbitrary so even this was not reliable. Only an extra configuration option like `MENDER_EXTRA_PARTS_LAST_PART` might be viable. But still, this does not solve the issue of having extra parts and resizable data partition.
 
- _adding your proposed strategy as a secondary route_
   I would go this way though. Maybe in combination with previous option but that would be out of my scope of interest. Adding a boolean config like `MENDER_EXTRA_PARTS_BEFORE_DATA` (or smt like that - naming is the hard part of computer science) that would control this behaviour. This could be automatically true if the user activates the `growfs-data` feature.

How do you feel about this?

TheYoctoJester:
> * _printing a warning or even error if both the partition growing feature and extra partitions are configured, as the behavior will be conflicting. I think this is the most obvious thing to do._
>   This is current behaviour of many projects including mender-kernel that we use. Simply warning the user does not solve our issue.

Yeah, users tend to ignore warnings. However, it is not exactly an error state, as the resulting image will work - so it really does not mandate an error to fail out. I think we can place some responsibility on users here.

> * _adding your proposed strategy as a secondary route_
>   I would go this way though. Maybe in combination with previous option but that would be out of my scope of interest. Adding a boolean config like `MENDER_EXTRA_PARTS_BEFORE_DATA` (or smt like that - naming is the hard part of computer science) that would control this behaviour. This could be automatically true if the user activates the `growfs-data` feature.

Yeah, naming is hard. But let's put that aside for a second, then I think this is a good route. Auto-enabling? No, because it would require at least a warning again, and nothing is won.

lluiscampos:
@katomaso @optical-o This PR has deviated now quite a bit from the original scope (reviewed by @TheYoctoJester ). Could we split the work in separate PRs? At least the first part (to make sure `/data` partition is last) we can wrap up and merge.

PR #2279, opened 2025-05-12
fix: Add full version of wget for mender inventory mender-inventory-geo

tactmaster:
### Description

mender-inventory-geo need --tires for wget.
wget in busybox doesn't have this option therefore need a full verion of wget to work

### DIFF
--- a/meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
+++ b/meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
@@ -63,6 +63,7 @@ RRECOMMENDS:mender-update:append = "${@bb.utils.contains('PACKAGECONFIG', 'modul
 RRECOMMENDS:mender-update:append = "${@bb.utils.contains('PACKAGECONFIG', 'modules', ' mender-flash', '', d)}"
 
 PACKAGECONFIG[inventory-network-scripts] = ",,,"
+RDEPENDS:mender-update:append = "${@bb.utils.contains('PACKAGECONFIG', 'inventory-network-scripts', ' wget', '', d)}"
 
 # NOTE: Splits the mender.conf file by default into a transient and a persistent config. Needs to be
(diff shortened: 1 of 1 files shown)


meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
```diff
@@ -37,7 +37,7 @@ inherit cmake
 
 DEPENDS += "lmdb googletest boost libarchive dbus"
 RDEPENDS:mender-auth:append = " boost-filesystem dbus"
-RDEPENDS:mender-update:append = " boost-filesystem libarchive lmdb dbus"
+RDEPENDS:mender-update:append = " boost-filesystem libarchive lmdb dbus wget"
```
lluiscampos, line 40: The specific script that needs `wget` is optional and controlled with `inventory-network-scripts`. So instead of adding here the dependency we should add a new line like:

```
RDEPENDS:mender-update:append = "${@bb.utils.contains('PACKAGECONFIG', 'inventory-network-scripts', ' wget', '', d)}"
```

Also, for better context and consistency with current code, add it just below `PACKAGECONFIG[inventory-network-scripts] = ",,,"` :pray:

lluiscampos (changes requested):
Good catch! See below

meta-mender-core/recipes-mender/mender-client/mender-client-cpp.inc
tactmaster: You know was I was looking for a place like that to put it. I just couldn't quite work it out.
Altered now.
Also thank my gitfu is improving.

lluiscampos: > Also thank my gitfu is improving

:man_dancing: :clap:

lluiscampos (approved):

lluiscampos:
@mender-test-bot start client pipeline :racing_car:

PR #2287, opened 2025-07-01
fix: remove ssh server limits on connections for testing

dhaustein:
https://northerntech.atlassian.net/browse/QA-472

- increases limits on SSH connections and auth attempts in the CI image

### DIFF
--- a/tests/meta-mender-ci/recipes-connectivity/openssh/openssh_%.bbappend
+++ b/tests/meta-mender-ci/recipes-connectivity/openssh/openssh_%.bbappend
@@ -1,0 +1,8 @@
+do_install:append() {
+    echo "" >> ${D}${sysconfdir}/ssh/sshd_config
+    echo "# Removing limits on max connections so tests do not get limited" >> ${D}${sysconfdir}/ssh/sshd_config
+    echo "LoginGraceTime 0" >> ${D}${sysconfdir}/ssh/sshd_config
+    echo "MaxAuthTries 1000000" >> ${D}${sysconfdir}/ssh/sshd_config
+    echo "MaxSessions 1000000" >> ${D}${sysconfdir}/ssh/sshd_config
+    echo "MaxStartups 1000000" >> ${D}${sysconfdir}/ssh/sshd_config
+}


dhaustein:
had to decrease the numbers a bit from MAX_INT to like a 1 000 000. OpenSSH has these values as INT so MAX_INT should've worked but it turns out the config with MAX_INT was silently ignored. I spent a lot of time trying to find out why the original config wasn't taking effect. 

Here's some testing results now that it works

Default SSH server configuration
```shell
$ ./test_ssh_percentage.sh root@127.0.0.1 200
Starting SSH connection test...
--------------------------------
Target Host: 127.0.0.1
Username: root
Attempting: 200 concurrent connections
Timeout per connection: 10 seconds
--------------------------------
All 200 connection attempts have been launched. Waiting for them to complete...

------ Test Complete ------
Total Connections Attempted: 200
Successful Connections:      114
Failed Connections:          86
Failure Percentage:          43.00%
```

Changed SSH server config to allow more connection
```shell
$ ./test_ssh_percentage.sh root@127.0.0.1 200
Starting SSH connection test...
--------------------------------
Target Host: 127.0.0.1
Username: root
Attempting: 200 concurrent connections
Timeout per connection: 10 seconds
--------------------------------
All 200 connection attempts have been launched. Waiting for them to complete...

------ Test Complete ------
Total Connections Attempted: 200
Successful Connections:      199
Failed Connections:          1
Failure Percentage:          0.50%
```

dhaustein:
About the failing [test:acceptance:vexpress_qemu:flash/master](https://gitlab.com/Northern.tech/Mender/mender-qa/-/jobs/10568791410) - this is happening also on master, so it's very likely **not** related to my changes here, since all the other checks are passing.

elkoniu (approved):

tests/meta-mender-ci/recipes-connectivity/openssh/openssh_%.bbappend
```diff
@@ -0,0 +1,8 @@
+do_install:append() {
+    echo "" >> ${D}${sysconfdir}/ssh/sshd_config
+    echo "# Removing limits on max connections so tests do not get limited" >> ${D}${sysconfdir}/ssh/sshd_config
+    echo "LoginGraceTime 0" >> ${D}${sysconfdir}/ssh/sshd_config
```
vpodzime, line 4: Are you sure this one is necessary? The default is 120 seconds. I don't see how the clients in our CI could miss that.

vpodzime (approved):
Looks good to me in general. We don't have to worry about this being too open because it's just for (virtual) devices running in the CI, right?

tests/meta-mender-ci/recipes-connectivity/openssh/openssh_%.bbappend
dhaustein: I saw this mentioned in the context of making the other values higher too so I just included it as well. I don't think this is going to be a problem, if anything, it's not going to have a practical effect if we could fit into the 120 secs before.

oldgiova:
Merging with failing checks because unrelated, as asked by @dhaustein

PR #2291, opened 2025-07-18
feat: retrieve journal log from tested vm

dhaustein:
Ticket: QA-1099

- adds scp call to the VM under test to export journal log and copy it over to CI workspace for archiving


lluiscampos:
Not the exact same patch, but I assume this is updated already after https://github.com/mendersoftware/meta-mender/pull/2303 ? Please close this one :broom:

PR #2308, opened 2025-09-08
feat: add mender-orchestrator-support recipes

danielskinstad:
Add recipes for mender-orchestrator-support package that provides:
- Interfaces for mender-orchestrator communication
- Update modules for orchestrator manifest handling

Ticket: MEN-7201

### DIFF
--- a/tests/acceptance/test_mender-orchestrator-support.py
+++ b/tests/acceptance/test_mender-orchestrator-support.py
@@ -1,0 +1,49 @@
+#!/usr/bin/python
+# Copyright 2025 Northern.tech AS
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+import subprocess
+
+import pytest
+
+from utils.common import (
+    build_image,
+    latest_build_artifact,
+)
+
+
+@pytest.mark.cross_platform
+class TestMenderOrchestratorSupport:
+    @pytest.mark.only_with_image("ext4")
+    def test_build_mender_orchestrator_support(
+        self, request, bitbake_variables, prepared_test_build, bitbake_image
+    ):
+        build_image(
+            prepared_test_build["build_dir"],
+            prepared_test_build["bitbake_corebase"],
+            bitbake_image,
+            ['IMAGE_INSTALL:append = " mender-orchestrator-support"'],
+        )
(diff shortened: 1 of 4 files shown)


tests/acceptance/test_mender-orchestrator-support.py
```diff
@@ -0,0 +1,50 @@
...
+import pytest
+
+from utils.common import (
+    build_image,
+    latest_build_artifact,
+)
+
+
+@pytest.mark.cross_platform
+@pytest.mark.min_mender_version("4.0.0")
```
michalkopczan, line 27: orchestrator does not work with 4.0.0 AFAIR, sometime later "resume" action was added which is needed by the orchestrator. I know this is not critical here, and I don't know the specific version where it started working, but we could mark 5.0.2 here - as this is the version where "empty response from server" error was fixed. We would make sure that we actually have a working system because of using 5.0.2

danielskinstad: Good point, thanks

meta-mender-core/recipes-mender/mender-orchestrator-support/mender-orchestrator-support_0.1.bb
```diff
@@ -0,0 +1,28 @@
...
+# THINGS TO CONSIDER FOR EACH RELEASE:
+# - SRC_URI (particularly "branch")
+# - SRCREV
+# - DEFAULT_PREFERENCE
+#-------------------------------------------------------------------------------
+
+SRC_URI = "git://github.com/mendersoftware/mender-orchestrator-support;protocol=https;branch=main"
+
+# We don't actually have a tag 0.1, so point it to the latest revision of main
+SRCREV = "a074100cf481ece4e3e201bb75a6b32344e98385"
```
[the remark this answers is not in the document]
danielskinstad: We could do a git tag. I just did what the ticket suggested and added an arbitrary version.

tests/acceptance/test_mender-orchestrator-support.py
danielskinstad: As Lluis said, installing the modules/interfaces doesn't really require a specific mender version at all, so I think it's fine to just remove it.

meta-mender-core/recipes-mender/mender-orchestrator-support/mender-orchestrator-support.inc
```diff
@@ -0,0 +1,24 @@
...
+        -C ${S} \
+        DESTDIR=${D} \
+        prefix=${D} \
+        install
+
+    # Manually install the orchestrator-manifest module
+    install -d -m 755 ${D}${datadir}/mender/modules/v3/
+    install -m 755 ${S}/modules/mender-orchestrator-manifest/module/v3/mender-orchestrator-manifest \
+        ${D}${datadir}/mender/modules/v3/
+}
```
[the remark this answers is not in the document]
danielskinstad: Ah, yeah I think it should, good catch.

danielskinstad:
@lluiscampos did you mean something like `meta-mender-core/mender-commercial/mender-orchestrator-support `?

lluiscampos (approved):

PR #2314, opened 2025-09-17
u-boot-mender: fix a circular dependency issue

liuming50:
A flaw was introduced by commit:
```
commit 772642ce5f2ed0bb730f833dfadb35225d349822
Author: Jonas Norling <jonas.norling@greeneggs.se>
Date:   Tue May 7 09:08:34 2019 +0200

    Fix build failing in do_image_ubimg task

    do_image_ubimg was allowed to run before do_deploy_append_mender-uboot
    was executed, leading to it not finding uboot.env in the deployment
    area. This fixes the problem by forcing the deploy to happen before
    do_package, which the image targets will naturally depend on.
```

by adding a task dependency do_package -> do_deploy, it can lead to a circular dependency issue between kernel and uboot in some cases like when secureboot (kernel sign/uboot sign) is enabled.

The issue the above commit was trying to fix is to let do_image_ubimg depend on u-boot:do_deploy, so let's just do so. (Same dependency has already been there for partitioning images)

the form below:

```

<OPTIONAL LONGER DESCRIPTION>

```

### Description

Thank you!


TheYoctoJester:
Hi @liuming50,

Thanks for the submission! Two things to fix/consider.
1. our commit linting requires the conventional commit specification to be followed, so the patch name should probably start with `fix: `. See also the spec at https://www.github.com/mendersoftware/mendertesting/commitlint/grammar.md.
2. In 772642ce5f2ed0bb730f833dfadb35225d349822, only a part of the task dependency is added, the `before do_package` part. Your PR removes the whole line, and I'm not sure if that might break other situations which "just" package `u-boot`. Would reverting to the state before the offending commit be an option for your use case, or is the circular dependency still present then?

Greetz,
Josef

lluiscampos:

lluiscampos (approved):
Thank you for your contributions. They look good to me.

PR #2445, opened 2026-05-07
fix: copy grow-data template to B before sed substitution

lluiscampos:
Otherwise subsequent runs will not find the `@...@` markers in the already modified template in workdir and nothing gets updated.

Fixes an issue were `mender-growfs-data` script is not updated after changing relevant `MENDER_*` variables unless we do a full sstate clean and rebuild.


danielskinstad:
Hola :wave:  Not every day I see the name "Lluis" in my notifications anymore; nice to see you fixing stuff as always!
The fix looks legit to me -- let's run a pipeline

danielskinstad (approved):
LGTM.
The failing pipeline is due to a pre-existing, unrelated, failure in integration.

lluiscampos:
Thank you for your speedy review @danielskinstad 🍻 I will be around for a while 😇

PR #2449, opened 2026-05-26
test: Added acceptance test for DEFAULT_PREFERENCE

rewanrashid-boop:
Added an acceptance test that checks whether DEFAULT_PREFERENCE='-1' is set when adding a recipe with a major version bump to maintenance branches.

Changelog: None
Ticket: QA-1586

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -14,6 +14,19 @@ include:
   - project: 'Northern.tech/Mender/mendertesting'
     file: '.gitlab-ci-github-status-updates.yml'
 
+test:check-version-bump:
+  stage: test
+  image: registry.gitlab.com/northern.tech/mender/mender-test-containers/base-ubuntu:master
+  variables:
+    GIT_SUBMODULE_STRATEGY: normal
+  before_script:
+    - export DEBIAN_FRONTEND=noninteractive
+    - python3 -m venv .venv
+    - source .venv/bin/activate
+    - pip3 install pytest
+  script:
+    - pytest -v -rs -s tests/version/test_version_bump.py
+
 default:
   tags: !reference [.qa-common-default-runner, tags]
   retry: !reference [.qa-common-default-retry, retry]
--- a/tests/version/test_version_bump.py
+++ b/tests/version/test_version_bump.py
@@ -1,0 +1,204 @@
+#!/usr/bin/python
+# Copyright 2026 Northern.tech AS
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+import os
+import json
+import re
+import subprocess
+import urllib.request
+from pathlib import Path
+import pytest
+
+GITHUB_AUTH = os.getenv("GITHUB_AUTH")
+CI_COMMIT_SHA = os.getenv("CI_COMMIT_SHA")
+RECIPE_PATTERN = r"^mender(-[a-z0-9-]+)?_([0-9]+)\.[0-9]+\.[0-9]+\.bb$"
+REPO_DIR = Path(__file__).resolve().parent.parent.parent
+YOCTO_BRANCHES = ["scarthgap", "wrynose"]
+
+if not GITHUB_AUTH or not CI_COMMIT_SHA:
+    pytest.exit("Missing required environment variables: GITHUB_AUTH or CI_COMMIT_SHA")
+
+
+def github_api_get(url):
+    """Helper to make authenticated GitHub API requests."""
+    req = urllib.request.Request(url)
+    req.add_header("Accept", "application/vnd.github+json")
+    req.add_header("User-Agent", "GitLab-CI-Python")
+    req.add_header(*GITHUB_AUTH.split(":", 1))
+
+    try:
+        with urllib.request.urlopen(req) as response:
+            return json.loads(response.read().decode("utf-8"))
+    except urllib.error.HTTPError as e:
+        if e.code in [401, 403]:
+            pytest.fail(
+                f"GitHub API Error ({e.code}) for URL: {url}. "
+                f"Your GITHUB_AUTH token header was rejected by GitHub's gateway."
+            )
+        else:
+            print(f"GitHub API Request Failed for {url}: {e}")
+            return None
+    except Exception as e:
+        print(f"GitHub API Request Failed for {url}: {e}")
+        return None
+
+@pytest.fixture(scope="session")
+def pr_num_from_sha():
+    """Returns the pull request number from environment or GitHub Search API."""
+    # Pull from GitLab's branch ref name (e.g., 'pr_2449')
+    ref_name = os.getenv("CI_COMMIT_REF_NAME", "")
+    ref_match = re.match(r"^pr_(\d+)$", ref_name)
+    if ref_match:
+        return ref_match.group(1)
+
+    # Fallback to Search API
+    url = f"https://api.github.com/search/issues?q=repo:mendersoftware/meta-mender+type:pr+sha:{CI_COMMIT_SHA}"
+    data = github_api_get(url)
+
+    if data and data.get("items"):
+        return str(data["items"][0].get("number", ""))
+
+    pytest.skip(
+        f"Commit {CI_COMMIT_SHA} is not yet indexed or associated with a GitHub PR."
+    )
+
+
+@pytest.fixture(scope="session")
+def target_branch(pr_num_from_sha):
+    """Returns the target branch of the pull request."""
+    url = f"https://api.github.com/repos/mendersoftware/meta-mender/pulls/{pr_num_from_sha}"
+    data = github_api_get(url)
+
+    if data and "base" in data:
+        return data["base"].get("ref")
+
+    pytest.skip(f"Could not retrieve target branch details for PR #{pr_num_from_sha}.")
+
+
+@pytest.fixture(scope="session")
+def target_recipes(target_branch):
+    """Gets the target branch recipe data."""
+    if target_branch not in YOCTO_BRANCHES:
+        pytest.skip(
+            "Skipping check: Target branch is not a maintenance branch (scarthgap / wrynose)"
+        )
+
+    cmd = ["git", "ls-tree", "-r", target_branch]
+
+    result = subprocess.run(
+        cmd, cwd=REPO_DIR, check=True, capture_output=True, text=True
+    )
+
+    # Find and process recipes
+    latest_recipes = {}
+    for filepath in result.stdout.splitlines():
+        if not filepath.endswith(".bb"):
+            continue
+
+        filename = os.path.basename(filepath)
+        match = re.match(RECIPE_PATTERN, filename)
+        if match:
+            major_version = int(match.group(2))
+            suffix = match.group(1) if match.group(1) else ""
+            base_name = f"mender{suffix}"
+
+            # Keep track of the highest major version found for this component type
+            if (
+                base_name not in latest_recipes
+                or major_version > latest_recipes[base_name]["major_version"]
+            ):
+                latest_recipes[base_name] = {
+                    "recipe": filename,
+                    "major_version": major_version,
+                }
+
+    # Return a dictionary mapped by filename
+    return {
+        base_name: item["major_version"] for base_name, item in latest_recipes.items()
+    }
+
+
+@pytest.fixture(scope="session")
+def commit_recipes(pr_num_from_sha):
+    """Gets the recipes modified or added in the pull request."""
+    url = f"https://api.github.com/repos/mendersoftware/meta-mender/pulls/{pr_num_from_sha}/files"
+    files_json = github_api_get(url)
+
+    recipes = []
+    if not files_json or not isinstance(files_json, list):
+        pytest.skip(
+            f"Could not retrieve the modified files list for PR #{pr_num_from_sha}."
+        )
+
+    commit_files = {
+        file_entry["filename"] for file_entry in files_json if "filename" in file_entry
+    }
+
+    for filepath in commit_files:
+        filename = os.path.basename(filepath)
+        match = re.match(RECIPE_PATTERN, filename)
+        if match:
+            suffix = match.group(1) if match.group(1) else ""
+            base_name = f"mender{suffix}"
+            recipes.append(
+                {
+                    "recipe": filename,
+                    "base_name": base_name,
+                    "major_version": int(match.group(2)),
+                }
+            )
+
+    return recipes
+
+
+def test_default_preference(target_branch, target_recipes, commit_recipes):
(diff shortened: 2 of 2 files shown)


rewanrashid-boop:

tests/acceptance/test_backport_preference.py
```diff
@@ -0,0 +1,118 @@
...
+    try:
+        return int(parts[0])
+    except (ValueError, IndexError):
+        return 0
+
+
+def get_modified_recipes_in_patch():
+    try:
+        # Get base commit the MR split from
+        if os.environ.get("CI_MERGE_REQUEST_DIFF_BASE_SHA"):
```
danielskinstad: This variable won't be set in CI as we're not actually running the pipeline on an "MR" in gitlab. The fallback below is also IMO too naive; I think we should be able to run the test locally and in CI if we have this as an acceptance test.

danielskinstad (changes requested):
Thanks for taking this task :rocket:
I haven't actually reviewed all of the code yet, but I commented on something that we need to change. I think the next steps are for you to try to find an alternative to the logic based on `CI_MERGE_REQUEST_DIFF_BASE_SHA` and ensuring it works in CI.

To test it in CI, I would try to create some dummy recipe with a major bump to try to trigger a failure. Let me know if want any help with testing it.

If an acceptance tests turns out to be too hard or causes too many issues, maybe an acceptance test isn't the way to go. We could also consider having this a separate job in CI, but I would only consider that if we can't find a nice way to have it as an acceptance test.

rewanrashid-boop:
I changed the test quite a bit here: I added ta new directory "tests/version" in meta-mender where the test now `test_version_bump.py` is. I also modified the `.gitlab-ci.yml` with a rule to start a new job with this test on maintenance branches including master. The test essentially checks the commit the pipeline started for to see if there are any changes made to versioned recipe files. If yes, it collects these and finds the latest version of the same files in meta-mender master, and compares to see if we have a major version bump. If yes then check if `DEFAULT_PREFERENCE = '-1'`:
1. not set → fail
2. set → pass

tests/acceptance/test_backport_preference.py
rewanrashid-boop: New test below

.gitlab-ci.yml
```diff
@@ -15,6 +15,20 @@ include:
...
+  image: python:3.12-slim
+  variables:
+    GIT_SUBMODULE_STRATEGY: normal
+  before_script:
+    - apt-get update && apt-get install -y git gcc liblzo2-dev
+    - pip install -r tests/acceptance/requirements_py3.txt
+  script:
+    - git config --global --add safe.directory $CI_PROJECT_DIR
+    - python3 -m pytest -v -c /dev/null tests/version/test_version_bump.py
+  rules:
```
danielskinstad: This won't work. We should run it on PRs targeting scarthgap or wrynose, not on the branches themselves.
Maybe we can do something like:
```
    changes:
      paths:
        - '**/mender_*.bb'
        - '**/mender-*.bb'
```
https://docs.gitlab.com/ci/yaml/#ruleschanges

This won't fix it properly, as it would run on regular pipelines PRs/pipelines, so we may need to think a bit more about it.

.gitlab-ci.yml
```diff
@@ -15,6 +15,20 @@ include:
   - project: 'Northern.tech/Mender/mendertesting'
     file: '.gitlab-ci-github-status-updates.yml'
 
+test:check-version-bump:
+  stage: test
+  image: python:3.12-slim
+  variables:
+    GIT_SUBMODULE_STRATEGY: normal
+  before_script:
+    - apt-get update && apt-get install -y git gcc liblzo2-dev
```
danielskinstad: Why do we need gcc and liblzo2-dev?

.gitlab-ci.yml
```diff
@@ -15,6 +15,20 @@ include:
   - project: 'Northern.tech/Mender/mendertesting'
     file: '.gitlab-ci-github-status-updates.yml'
 
+test:check-version-bump:
+  stage: test
+  image: python:3.12-slim
```
danielskinstad: I'd use one of the base container images from mender-test-containers so we don't need to install git every time we run the job. E.g.
`registry.gitlab.com/northern.tech/mender/mender-test-containers/base-debian:master`

danielskinstad (commented):
AFAICT, we shouldn't need to clone and compare it the master branch -- you should already have all the information available to determine the default preference. However, I don't know the best approach here, so maybe we can think about it and discuss it sometime next week.

.gitlab-ci.yml
rewanrashid-boop: this was probably left over from when i had the test in /tests/acceptance and can be removed. Pytest was being too helpful and trying to run the bitbake fixtures there so i included it as a temporary workaround to test but ended up not using it anyway, can just remove it :)

.gitlab-ci.yml
rewanrashid-boop: right, okay so instead of checking the ocmmit branch we should somehow check the target of the pr and if that is master, scarthgap or wrynose then the test should run, something like that ?

rewanrashid-boop: oh yeah right, and I remember we talked about the problem of how the external pull request target branch variable will only be set in some cases and won't cover other case e.g. when we start a pipeline manually gitlab

tests/version/test_version_bump.py
```diff
@@ -0,0 +1,235 @@
...
+from pathlib import Path
+import pytest
+
+GITHUB_AUTH = os.getenv("GITHUB_AUTH")
+CI_COMMIT_SHA = os.getenv("CI_COMMIT_SHA")
+RECIPE_PATTERN = r"^mender(-[a-z0-9-]+)?_([0-9]+)\.[0-9]+\.[0-9]+\.bb$"
+REPO_DIR = Path(__file__).resolve().parent.parent.parent
+
+@pytest.fixture(scope="session")
+def pr_num_from_sha():  # DONE!
```
p-targowicz: I think its leftover comment 
```suggestion
def pr_num_from_sha():
```

tests/version/test_version_bump.py
```diff
@@ -0,0 +1,235 @@
...
+    process_curl.stdout.close()
+    process_curl.wait()
+
+    pr_num = process_jq.stdout.strip()
+
+    return pr_num
+
+
+@pytest.fixture(scope="session")
+def target_branch(pr_num_from_sha):  # DONE!
```
p-targowicz: same here 
```suggestion
def target_branch(pr_num_from_sha):
```

tests/version/test_version_bump.py
```diff
@@ -0,0 +1,235 @@
...
+
+    cmd = [
+        "curl",
+        "-L",
+        "-H",
+        "Accept: application/vnd.github+json",
+        "-H",
+        f"Authorization: Bearer {GITHUB_AUTH}",
+        "-H",
+        "X-GitHub-Api-Version: 2026-03-10",
```
p-targowicz: this version seems to be made up, we can either drop this arg or use latest one `2022-11-28`

tests/version/test_version_bump.py
```diff
@@ -0,0 +1,235 @@
...
+                base_name = f"mender{suffix}"
+                recipes.append(
+                    {
+                        "recipe": filename,
+                        "base_name": base_name,
+                        "major_version": int(match.group(2)),
+                    }
+                )
+    except json.JSONDecodeError:
+        print(f"Error: Could not parse API reponse with PR number: {pr_num_from_sha}")
```
p-targowicz: NIT: minor typo - `reponse` instead of `response`

tests/version/test_version_bump.py
```diff
@@ -0,0 +1,235 @@
...
+
+    return recipes
+
+
+def test_default_preference(target_branch, master_recipes, commit_recipes):
+    """Validates that backported major bumps include DEFAULT_PREFERENCE = '-1'."""
+
+    if target_branch not in ["scarthgap", "wrynose"]:
+        pytest.skip(
+            f"Skipping check: Target branch is not a maintenance branch (scarthgap / wrynose)"
```
p-targowicz: f-string not used can be dropped

tests/version/test_version_bump.py
```diff
@@ -0,0 +1,235 @@
...
+        filename = commit_item["recipe"]
+        base_name = commit_item["base_name"]
+        commit_version = commit_item["major_version"]
+
+        # Check if this exact file existed in master
+        if base_name in master_recipes:
+            master_version = master_recipes[base_name]
+
+            # If the current commit features a higher major version than master
+            if commit_version > master_version:
```
p-targowicz: this compares against MASTER, but master always holds the newest version -> for a real backport commit_version > master_version is essentially never true so the assert at the bottom cant be true

The correct reference is the TARGET maintenance branch's current state, not master
siimpler: detect when >1 major of a component is present in the tree and assert the highest-major file has `DEFAULT_PREFERENCE = '-1'`

.gitlab-ci.yml
```diff
@@ -15,6 +15,21 @@ include:
...
     file: '.gitlab-ci-github-status-updates.yml'
 
+test:check-version-bump:
+  stage: test
+  image: registry.gitlab.com/northern.tech/mender/mender-test-containers/base-ubuntu:master
+  variables:
+    GIT_SUBMODULE_STRATEGY: normal
+  before_script:
+    - export DEBIAN_FRONTEND=noninteractive
+    - apt-get update -y && apt-get install -y --no-install-recommends python3 python3-pip python3-venv
```
p-targowicz: are we sure `registry.gitlab.com/northern.tech/mender/mender-test-containers/base-ubuntu:master` has  `jq` installed ? we use it in script but we dont install it here so we assume its there

.gitlab-ci.yml
```diff
@@ -15,6 +15,21 @@ include:
...
+    - python3 -m venv .venv
+    - source .venv/bin/activate
+    - pip3 install pytest
+  script:
+    - git fetch origin master:master --depth=50
```
p-targowicz: this is fine on a detached pull/NNNN/head ref (normal PR pipeline exit 0), **BUT** on a master pipeline (HEAD on master) git refusing to fetch into branch 'refs/heads/master' and job dies before pytest, currently this would run on master too

lluiscampos:
Greetings from an external contributor that is so happy to see this task coming through 🚀

rewanrashid-boop:
@lluiscampos Lluis!!!!

.gitlab-ci.yml
rewanrashid-boop: I think it is yeah https://github.com/mendersoftware/mender-test-containers/blob/c262c665422c7e814a44da107c8e718639e415a3/base-ubuntu/Dockerfile#L15

tests/version/test_version_bump.py
rewanrashid-boop: I suppose i can change the master_recipes function into target_recipes which and modify slightly to get the latest recipes from the target branch i.e. scarthgap or wrynose and then down in the actual test we compare the commit recipes with the target latest recipe

tests/version/test_version_bump.py
rewanrashid-boop: fixed just  havent  commit yet

tests/version/test_version_bump.py
rewanrashid-boop: fixed just  havent  commit yet

tests/version/test_version_bump.py
rewanrashid-boop: removed the line entirely for simplicity

.gitlab-ci.yml
rewanrashid-boop: I suppose then this line can just be removed? Since we are on master on meta mender i can, in the new target_recipes function, just checkout into the target branch and grep there for the latest recipes and then checking for default preference can proceed as normal now correctly comparing the commit recipes and the target branch current recipes.

.gitlab-ci.yml
```diff
@@ -14,6 +14,20 @@ include:
...
     file: '.gitlab-ci-github-status-updates.yml'
 
+test:check-version-bump:
+  stage: test
+  image: registry.gitlab.com/northern.tech/mender/mender-test-containers/base-ubuntu:master
+  variables:
+    GIT_SUBMODULE_STRATEGY: normal
+  before_script:
+    - export DEBIAN_FRONTEND=noninteractive
+    - apt-get update -y && apt-get install -y --no-install-recommends python3 python3-pip python3-venv
```
danielskinstad: This seems like something that should be added to the base image's dependencies itself in mender-test-containers.

tests/version/test_version_bump.py
```diff
@@ -0,0 +1,224 @@
...
+        print(f"Error: Could not parse API response with PR number: {pr_num_from_sha}")
+        print(result.stderr)
+
+    return recipes
+
+
+def test_default_preference(target_branch, target_recipes, commit_recipes):
+    """Validates that backported major bumps include DEFAULT_PREFERENCE = '-1'."""
+
+    if target_branch not in ["scarthgap", "wrynose"]:
```
danielskinstad: I am wondering if we shouldn't make this more visible? Maybe as a list defined at the top of the file
```
YOCTO_BRANCHES = [
    "scarthgap",
    "wrynose",
]

**Comment by rewanrashid-boop** [comment]
Removed some of the weird jq subprocesses and used urllib.requests for API calls instead. Also changed to look in the target branch repository for .bb files rather than master as discussed.

File: tests/version/test_version_bump.py [hunk 15]
```diff
@@ -0,0 +1,206 @@
...
+    if not pr_num_from_sha:
+        return "unknown"
+
+    url = f"https://api.github.com/repos/mendersoftware/meta-mender/pulls/{pr_num_from_sha}"
+    data = github_api_get(url)
+
+    if data and "base" in data:
+        return data["base"].get("ref", "unknown")
+
+    return "unknown"
```
danielskinstad: Why do we want to return "unknown"? If the target branch is unknown, the test won't work and could just be instantly skipped, right?

tests/version/test_version_bump.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+@pytest.fixture(scope="session")
+def target_recipes(target_branch):
+    """Gets the master branch recipe data."""
+    # Safeguard: If branch resolution failed, skip git command execution entirely
+    if not target_branch or target_branch in ["unknown", "null"]:
```
danielskinstad: If we have a pytest.skip instead of returning "unknown", then this "safeguard" would be redundant, right?

tests/version/test_version_bump.py
p-targowicz: sounds fair

.gitlab-ci.yml
p-targowicz: I guess so 👍

.gitlab-ci.yml
rewanrashid-boop: should i then remove it and udpate mender-test-containers or is there another image with these dependencies i should use?

tests/version/test_version_bump.py
rewanrashid-boop: Agree

tests/version/test_version_bump.py
rewanrashid-boop: Alright, these things should all have been replaced with pytest.skip now

rewanrashid-boop:
`target_branch` skips skips for some reason now that I'm looking at the CI log. I'll look into why it is not finding it. Here it should skip at `"Skipping check: Target branch is not a maintenance branch (scarthgap / wrynose)"`

danielskinstad (approved):

rewanrashid-boop:
* wrynose

PR #2462, opened 2026-06-16
fix: extend grep for version check in mender-gateay.inc

danielskinstad:
Ticket: None
Changelog: Fix an issue where the version check for mender-gateway was too strict for binaries compiled with newer golang versions.


danielskinstad:
Commit with amended changelog from https://github.com/mendersoftware/meta-mender/pull/2461 (so we don't need to wait with the merge)

elkoniu (approved):
LGTM!

PR #2466, opened 2026-06-22
fix(libubootenv): Fail the build when detecting miss-alignment

lluiscampos:


lluiscampos:
Dear reviewers,

I am not 100% sure about this being a fix. According do [the docs](https://docs.yoctoproject.org/next/dev-manual/debugging.html#logging-with-bash), `bberror` is used for recoverable errors, but our recipe does not handle it in any way so I think that the safest is to fail hard.

danielskinstad:
@mender-test-bot start client pipeline (sorry for the long wait)

lluiscampos:
@danielskinstad 🏓

danielskinstad (approved):
LGTM :)

PR #2491, opened 2026-07-24
WIP

danielskinstad:
Testing


danielskinstad:
Looks like this would also require a patch of the kernel, but it would make things extremely slow.

PR #2511, opened 2026-08-19
fix(orchestrator): fix addint mqtt demo Interface in bbappend

michalkopczan:
Ticket: MEN-9982


michalkopczan:

michalkopczan:
@mender-test-bot start client pipeline --pr mender-qa/903 --pr meta-mender/2511

michalkopczan:
I've run a [pipeline](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/2775226560) for both PRs that should fix the issue:
https://github.com/mendersoftware/meta-mender/pull/2511
https://github.com/mendersoftware/mender-qa/pull/903

However, due to mender-qa being special in a way that its version (branch used) is defined differently than all other component's revisions, it causes issues with versioning information. Particularly, it assumes `local filename=/builds/Northern.tech/Mender/stage-artifacts/mender-delta-container-modules/delta-docker-compose-pr_903.tar.xz`, so delta-docker-compose version is pr_903, while in reality it is the mender-qa's version used.

Due to this and the fact that these two PRs should fix the issue, we decided with @vpodzime that it's better to merge both now and eventually fix it further if needed, instead of trying to make the combined pipeline work - as it's an extremely rare situation.

PR #2512, opened 2026-08-21
fix: derive closed-source S dir from the tarball name

danielskinstad:
S used to be derived from PV (which follows PREFERRED_VERSION), so with "main-git%" the recipe would look for e.g. delta-docker-compose-main even when the tarball (and the directory inside it) was named delta-docker-compose-master, failing the build. The tarball always unpacks to a directory with the same name as itself, so instead strip the "<repo_name>-" prefix and the .tar.* extension from the filename and use whatever is left ("main", "master", a sha or a tag) for S. PV is unchanged and still follows PREFERRED_VERSION.

Ticket: None
Changelog: Derive closed-source S dir from the tarball name instead of `PREFERRED_VERSION`. This fixes builds where e.g. `PREFERRED_VERSION` is `main-git%` but the tarball is named `-master`, resulting in the S dir not being found.

### DIFF
--- a/meta-mender-commercial/classes/mender-closed-source-utils.bbclass
+++ b/meta-mender-commercial/classes/mender-closed-source-utils.bbclass
@@ -57,11 +57,25 @@ def mender_closed_source_pv_from_preferred_version(d, srcrev):
         branch = pref_version.rsplit("-git", 1)[0]
     return "%s%s" % (branch, srcrev)
 
-def mender_closed_source_tarball_dir_from_pv(d, pv):
-    # PV is "<branch>-git+<sha>" for git builds, "<branch>" for branch
-    # tarballs, or a tagged version like "1.2.3" / "1.2.3-build1". The
-    # directory inside the tarball is the trailing "<sha>" for git builds,
-    # otherwise the literal PV.
-    if "-git+" in pv:
-        return pv.split("-git+", 1)[1]
-    return pv
+def mender_closed_source_tarball_dir_version_from_src_uri(d, src_uri, repo_name):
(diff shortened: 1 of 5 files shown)


danielskinstad:

michalkopczan (approved):

meta-mender-commercial/classes/mender-closed-source-utils.bbclass
```diff
@@ -57,11 +57,25 @@ def mender_closed_source_pv_from_preferred_version(d, srcrev):
...
 
-def mender_closed_source_tarball_dir_from_pv(d, pv):
-    # PV is "<branch>-git+<sha>" for git builds, "<branch>" for branch
-    # tarballs, or a tagged version like "1.2.3" / "1.2.3-build1". The
-    # directory inside the tarball is the trailing "<sha>" for git builds,
-    # otherwise the literal PV.
-    if "-git+" in pv:
-        return pv.split("-git+", 1)[1]
-    return pv
+def mender_closed_source_tarball_dir_from_src_uri(d, src_uri, repo_name):
```
vpodzime: `...tarball_dir_version...`? Otherwise the name is misleading because it doesn't return neither dirname nor a full path.

vpodzime (approved):
Looks good to me otherwise.

meta-mender-commercial/classes/mender-closed-source-utils.bbclass
danielskinstad: Alright :+1:

danielskinstad:
*wrynose
*scarthgap

PR #2525, opened 2026-09-16
feat: add qemu image with mender-orchestrator preinstalled

mzedel:
This deliberately does not use the mender-orchestrator-install feature. That feature also sets MENDER_DEVICE_TIER, which is consumed by the mender-client recipe rather than by the image, and the client package is
shared by every image in a build directory.
The existing orchestrator tests all pass a topology.yaml through SRC_URI, which is exactly what suppresses INSTALL_DEMO_TOPOLOGY.


mzedel:
@mender-test-bot start client pipeline --pr mender-qa/914

meta-mender-demo/mender-commercial/recipes-extended/images/mender-orchestrator-image-full-cmdline.bb
```diff
@@ -0,0 +1,32 @@
...
+# Orchestrator deployments need more space, doubled as per MEN-7405 recommendation
+MENDER_STORAGE_TOTAL_SIZE_MB ?= "2048"
+MENDER_DATA_PART_SIZE_MB ?= "512"
+
+# Deliberately not using MENDER_FEATURES_ENABLE = "mender-orchestrator-install" here
+# as that also sets MENDER_DEVICE_TIER, which is consumed by the general mender-client
+# recipe (do_prepare_mender_conf), vs. only this image. Consumers can set DeviceTier
+# at runtime instead, via the container's DEVICE_TIER env.
+IMAGE_INSTALL:append = " mender-orchestrator mender-orchestrator-support"
+
```
danielskinstad, line 8: You also need to set `MENDER_DEVICE_TIER = "system"`
https://docs.mender.io/orchestrate-updates/installation/yocto-project

mzedel:
@mender-test-bot start client pipeline --pr mender-qa/914

mzedel:
@mender-test-bot start client pipeline --pr mender-qa/914

mzedel:
@mender-test-bot start client pipeline --pr mender-qa/914 --pr mender/2028

meta-mender-demo/mender-commercial/recipes-extended/images/mender-orchestrator-image-full-cmdline.bb
mzedel: Thanks for taking a look - I went with the full feature instead so the system tier was set right away 🤗

meta-mender-demo/mender-commercial/recipes-extended/images/mender-orchestrator-image-full-cmdline.bb
```diff
@@ -0,0 +1,44 @@
...
+        bb.fatal("%s is missing, cannot set DeviceTier for the orchestrator" % conf)
+
+    with open(conf) as fd:
+        config = json.load(fd)
+    config["DeviceTier"] = d.getVar("MENDER_DEVICE_TIER")
+    with open(conf, "w") as fd:
+        json.dump(config, fd, indent=4, sort_keys=True)
+}
+set_orchestrator_device_tier[vardeps] += "MENDER_DEVICE_TIER"
+ROOTFS_POSTPROCESS_COMMAND += "set_orchestrator_device_tier;"
```
danielskinstad, line 29: This should already be taken care of by the `mender-orchestrator-install` feature https://github.com/mendersoftware/meta-mender/blob/df064eb4dcc5fce2ae01c4a78cd7d8f3666bcef7/meta-mender-core/classes/mender-setup.bbclass#L242

tests/acceptance/commercial/test_mender-orchestrator.py
```diff
@@ -149,3 +149,33 @@ def test_build_mender_orchestrator_mender_feature(
         ).decode()
         data = yaml.safe_load(conf)
         assert data["DeviceTier"] == "system", "Expected 'DeviceTier' to be 'system'"
+
+    def test_build_mender_orchestrator_demo_topology(
```
danielskinstad: This test is quite similar to https://github.com/mendersoftware/meta-mender/blob/df064eb4dcc5fce2ae01c4a78cd7d8f3666bcef7/tests/acceptance/commercial/test_mender-orchestrator-support.py#L33

I would rather modify the pre-existing test to check the demo topology (iow remove the one we're inserting in the test) instead of creating a new test

meta-mender-demo/mender-commercial/recipes-extended/images/mender-orchestrator-image-full-cmdline.bb
mzedel: I tried running the image without it now and it apparently isn't taken over from there, as now the device again says `standard` tier:
image

Since I can't put it into better words I'll quote an ancient roman philosopher:

That override resolves against `DISTROOVERRIDES`, which `mender-maybe-setup.bbclass:23` builds from that recipe's `MENDER_FEATURES_ENABLE`. `MENDER_DEVICE_TIER` is only ever read by `do_prepare_mender_conf` in the mender-client recipe (`mender-client-cpp.inc:218`), which generates `mender.conf`. Our image recipe appends the feature to its own datastore only, so:

- `IMAGE_INSTALL:append:mender-orchestrator-install` → applies ✓ (same recipe)
- `MENDER_DEVICE_TIER:mender-orchestrator-install` → does not reach mender-client ✗

The assumption is correct when the feature is set build-wide, which is exactly how `test_build_mender_orchestrator_mender_feature` exercises it — that test passes the feature as a `local.conf` line, and then the `DeviceTier == "system"` assertion passes. Per-image it's only half the feature.

tests/acceptance/commercial/test_mender-orchestrator.py
mzedel: moved it in there 👌

meta-mender-demo/mender-commercial/recipes-extended/images/mender-orchestrator-image-full-cmdline.bb
danielskinstad: Right, it only seems to work when enabled in local.conf. Alright then, sorry for the noise

danielskinstad (approved):
LGTM

PR #2526, opened 2026-09-16
fix(qemu): raise systemd device timeout for /data and /boot/efi

danielskinstad:
The qemu CI devices run under TCG without KVM on oversubscribed runners. On the wrynose images udev regularly needs more than systemd's default 90 s to settle /dev/sda*, so the /data and /boot/efi mounts fail and the device boots into emergency mode without sshd or mender. The integration tests then see "Device never rebooted". Give the qemu machines a generous device timeout; real boards are untouched.

Ticket: None
Changelog: None


danielskinstad:

p-targowicz (approved):

michalkopczan (approved):

